PR #1, opened 2018-01-15
error handling, unit test framework, error test cases, initial projec…

JonathanHenson:
…t setup.

error handling, unit test framework, error test cases, initial project setup.


CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+cmake_minimum_required (VERSION 3.0)
+project (aws-c-common)
+
+set(PROJECT_NAME aws-c-common)
+
+ file(GLOB AWS_COMMON_HEADERS
```
bdonlan: inconsistent indentation

include/aws/common/common.h
```diff
@@ -0,0 +1,100 @@
...
+                         _a < _b ? _a : _b; })
+
+#define AWS_MAX(a, b) ({ AWS_TYPE_OF (a) _a = (a); \
+                         AWS_TYPE_OF (b) _b = (b); \
+                         _a > _b ? _a : _b; })
+
+#define AWS_CACHE_ALIGN AWS_ALIGN(AWS_CACHE_LINE)
+
+typedef enum aws_common_error {
+    AWS_ERROR_SUCCESS,
```
bdonlan: Would recommend an explicit = 0 here, mostly just for documentation.

include/aws/common/error.h
```diff
@@ -0,0 +1,104 @@
...
+
+struct aws_error_info {
+    int error_code;
+    const char *literal_name;
+    const char *error_str;
+    const char *lib_name;
+    const char *formatted_name;
+};
+
+struct aws_error_info_list {
```
bdonlan: Would suggest having sizeof(aws_error_info) and sizeof(aws_error_info_list) fields in this struct for future expansion.

include/aws/common/error.h
```diff
@@ -0,0 +1,104 @@
...
+
+/*
+ * Returns libname concatenated with error string.
+ */
+AWS_COMMON_API const char *aws_error_debug_str(int err);
+
+/*
+ * Raises `err` to the installed callbacks, and sets the thread's error.
+ */
+AWS_COMMON_API void aws_raise_error(int err);
```
bdonlan: Have this return `-1` (or whatever the standard return code is for errored functions). This lets it be used in code like:

```
if (condition) return aws_raise_error(AWS_ERR_FOO);
```

include/aws/common/error.h
```diff
@@ -0,0 +1,104 @@
...
+}
+
+typedef void(*aws_error_handler)(int err, void *ctx);
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * Returns the latest error code on the current thread, or 0 if none have occurred.
```
bdonlan: Does this function clear the error code, or does it leave it there?

include/aws/common/error.h
```diff
@@ -0,0 +1,104 @@
...
+ */
+AWS_COMMON_API void aws_restore_error(int err);
+
+/*
+ * Sets an application wide error handler function. This will be overridden by the thread local handler
+ */
+AWS_COMMON_API void aws_set_global_error_handler_fn(aws_error_handler handler, void *ctx);
+
+/*
+ * Sets a thread-local error handler function. This will override the global handler.
```
bdonlan: How do we unset it? Also this needs to return the old handler so it can be restored.

source/error.c
```diff
@@ -0,0 +1,135 @@
...
+#include <aws/common/error.h>
+#include <aws/common/common.h>
+#include <assert.h>
+
+AWS_THREAD_LOCAL int last_error = 0;
+
+static aws_error_handler global_handler = NULL;
+static void *global_error_context = NULL;
+
+AWS_THREAD_LOCAL aws_error_handler thread_handler = NULL;
```
bdonlan: static as well

source/error.c
```diff
@@ -0,0 +1,135 @@
...
+int aws_last_error(void) {
+    return last_error;
+}
+
+static const struct aws_error_info *get_error_by_code(int err) {
+    if(err >= AWS_MAX_ERROR_SLOTS * AWS_ERROR_SLOT_SIZE) {
+        return NULL;
+    }
+
+    int slot_index = err / AWS_ERROR_SLOT_SIZE;
```
bdonlan: We should make AWS_ERROR_SLOT_SIZE a power of two to avoid division.

source/error.c
```diff
@@ -0,0 +1,135 @@
...
+    int error_index = err % AWS_ERROR_SLOT_SIZE;
+
+    volatile const struct aws_error_info_list *error_slot = error_slots[slot_index];
```
bdonlan: This doesn't need to be volatile (the array at L34 does however).

tests/aws_test_harness.h
```diff
@@ -0,0 +1,177 @@
...
+    size_t size;
+    void *blob;
+};
+
+static void *mem_acquire_malloc(struct aws_allocator *config, size_t size) {
+    struct memory_test_config *test_config = (struct memory_test_config *)config;
+    test_config->allocated += size;
+    struct memory_test_tracker *memory = (struct memory_test_tracker *) malloc(size + sizeof(struct memory_test_tracker));
+    memory->size = size;
+    memory->blob = memory + sizeof(struct memory_test_tracker);
```
bdonlan: Technically this may not be properly aligned for all datastructures after being returned (e.g. this is now not aligned for 256-bit aligned vector operations).

bdonlan: Also, this is incorrect pointer arithmetic - `memory` is `struct memory_test_tracker *` and is advanced in multiples of `sizeof(struct memory_test_tracker)`. As such here you're moving forward `sizeof(struct memory_test_tracker) * sizeof(struct memory_test_tracker)` bytes.

tests/aws_test_harness.h
```diff
@@ -0,0 +1,177 @@
...
+    return memory->blob;
+}
+
+static void mem_release_free(struct aws_allocator *config, void *ptr) {
+    struct memory_test_config *test_config = (struct memory_test_config *)config;
+
+    struct memory_test_tracker *memory = (struct memory_test_tracker *) (ptr - sizeof(struct memory_test_tracker));
```
bdonlan: Pointer arithmetic on `void *` is not allowed by C99. GCC accepts it as an extension; turn on a higher warning level if it didn't flag this.

tests/aws_test_harness.h
```diff
@@ -0,0 +1,177 @@
...
+        FAIL("Expected:%02x got:%02x - " message, expected[i], got[i], ## __VA_ARGS__);\
+    }\
+}
+
+typedef void(*aws_test_before)(struct aws_allocator *allocator, void *ctx);
+typedef int(*aws_test_run)(struct aws_allocator *allocator, void *ctx);
+typedef void(*aws_test_after)(struct aws_allocator *allocator, void *ctx);
+
+struct aws_test_harness {
+    aws_test_before on_before;
```
bdonlan: Is before/after really needed?

tests/error_test.c
```diff
@@ -0,0 +1,251 @@
...
+    aws_set_thread_local_error_handler_fn(error_test_thread_local_cb, &cb_data);
+
+    aws_raise_error(test_error_2.error_code);
+    ASSERT_INT_EQUALS(test_error_2.error_code, aws_last_error(), "Expected error code %d, but was %d",
+                      test_error_2.error_code, aws_last_error());
+
+
+    ASSERT_NOT_NULL(cb_data.last_seen, "last error should not have been null");
+    ASSERT_FALSE(cb_data.global_cb_called, "Global Callback should not have been invoked");
+    ASSERT_TRUE(cb_data.tl_cb_called, "Global Callback should have been invoked");
```
bdonlan: Thread local callback should have been invoked

tests/error_test.c
```diff
@@ -0,0 +1,251 @@
...
+
+    ASSERT_INT_EQUALS(test_error_2.error_code, cb_data.last_seen, "Expected error code %d, but was %d",
+                      test_error_2.error_code, cb_data.last_seen);
+    ASSERT_STR_EQUALS(test_error_2.error_str, aws_error_str(cb_data.last_seen), "Expected error string %s, but got %s",
+                      test_error_2.error_str, aws_error_str(cb_data.last_seen));
+    ASSERT_STR_EQUALS(test_error_2.lib_name, aws_error_lib_name(cb_data.last_seen), "Expected error libname %s, but got %s",
+                      test_error_2.lib_name, aws_error_lib_name(cb_data.last_seen));
+
+    return 0;
+}
```
bdonlan: When we add a thread abstraction we should test the thread-locality of the error APIs.

include/aws/common/error.h
JonathanHenson: does not clear it, just retrieves it. aws_reset_error() resets it

include/aws/common/error.h
JonathanHenson: good idea.

include/aws/common/error.h
JonathanHenson: set it to NULL would unset it, agreed on the return.

source/error.c
JonathanHenson: good catch

source/error.c
JonathanHenson: will do.

source/error.c
JonathanHenson: ok

tests/aws_test_harness.h
JonathanHenson: it's at the highest level, maybe it's because that code isn't being hit yet.

tests/aws_test_harness.h
JonathanHenson: they will be. They aren't yet.

tests/error_test.c
JonathanHenson: will update

tests/error_test.c
JonathanHenson: I'll ad a todo.

source/error.c
JonathanHenson: according to compiler rules it does.

JonathanHenson: initialization discards ‘volatile’ qualifier from pointer target type [-Werror=discarded-qualifiers]

tests/aws_test_harness.h
JonathanHenson: fixed.

include/aws/common/error.h
JonathanHenson: lets come back to this one. We have some structs will have to do this for soon and I'd like to nail the pattern in that use case.

source/error.c
JonathanHenson: crap forgot this one. one sec

JonathanHenson: wait lets hold off on this one. for one, this isn't hit in high-perf code, and I like the predictable ranges we get with 1000.

source/error.c
bdonlan: Hmm, in that case we put volatile at the wrong point in the declaration. That should be a 

```
static const struct aws_error_info_list *volatile error_slots[AWS_MAX_ERROR_SLOTS] = { 0 };
```

That is, we want the _pointer to_ the structure to be volatile, but the structure itself, being a compile time constant, need not be volatile.

source/error.c
bdonlan: What about predictable ranges of `0x1000`?

source/error.c
JonathanHenson: ah.... pesky qualifiers. That worked.

source/error.c
JonathanHenson: that's a good idea.

JonathanHenson: let's do 0x400 = 1024

CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+cmake_minimum_required(VERSION 3.0)
+project(aws-c-common)
+
+set(PROJECT_NAME aws-c-common)
```
marcomagdy: not used, but should be?

CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+        source_group("Source Files\\windows" FILES ${AWS_COMMON_OS_SRC})
+    endif ()
+
+    set(PLATFORM_LIBS "Kernel32")
+else ()
+    file(GLOB AWS_COMMON_OS_SRC
+            "source/posix/*.c"
+            )
+
+    set(PLATFORM_LIBS "pthread" "rt")
```
marcomagdy: librt doesn't exist on macos

CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+if (BUILD_SHARED_LIBS)
+    set(LIBTYPE SHARED)
+    if (WIN32)
+        add_definitions("-DUSE_IMPORT_EXPORT")
+    endif ()
+else ()
+    set(LIBTYPE STATIC)
+endif ()
+
+add_library(aws-c-common ${LIBTYPE} ${COMMON_HEADERS} ${COMMON_SRC})
```
marcomagdy: ${LIBTYPE} is not needed. CMake builds the correct thing based on the value of BUILD_SHARED_LIBS

CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+set_target_properties(aws-c-common PROPERTIES LINKER_LANGUAGE C C_STANDARD 99)
+set(CMAKE_C_FLAGS_DEBUGOPT "")
+
+#set warnings
+if (MSVC)
+    target_compile_options(aws-c-common PRIVATE /W4)
+else ()
+    target_compile_options(aws-c-common PRIVATE -Wall -Wno-long-long -pedantic -Werror)
```
marcomagdy: +1 for PRIVATE

CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+    if (WIN32)
+        add_definitions("-DUSE_IMPORT_EXPORT")
+    endif ()
+else ()
+    set(LIBTYPE STATIC)
+endif ()
+
+add_library(aws-c-common ${LIBTYPE} ${COMMON_HEADERS} ${COMMON_SRC})
+set_target_properties(aws-c-common PROPERTIES LINKER_LANGUAGE C C_STANDARD 99)
+set(CMAKE_C_FLAGS_DEBUGOPT "")
```
marcomagdy: What is this for?

CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+        $<INSTALL_INTERFACE:include>)
+
+target_link_libraries(aws-c-common ${PLATFORM_LIBS})
+
+file(GLOB TEST_SRC "tests/*.c")
+file(GLOB TEST_HDRS "tests/*.h")
+file(GLOB TESTS ${TEST_HDRS} ${TEST_SRC})
+
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/tests)
```
marcomagdy: better to use target_include_directories

CMakeLists.txt
bdonlan: I used this cmake snippet to probe for these libraries when playing around with macos, if it helps.

```
CHECK_LIBRARY_EXISTS("rt" "clock_gettime" "" HAVE_LIBRT)
if(HAVE_LIBRT)
    set(PLATFORM_LIBS ${PLATFORM_LIBS} "rt")
endif()

CHECK_LIBRARY_EXISTS("pthread" "pthread_mutex_lock" "" HAVE_LIBPTHREAD)
if(HAVE_LIBRT)
    set(PLATFORM_LIBS ${PLATFORM_LIBS} "pthread")
endif()
```

include/aws/common/error.h
bdonlan: We should document this explicitly.

JonathanHenson: done

source/error.c
```diff
@@ -0,0 +1,152 @@
...
+
+static AWS_THREAD_LOCAL aws_error_handler thread_handler = NULL;
+AWS_THREAD_LOCAL void *thread_handler_context = NULL;
+
+#ifndef AWS_MAX_ERROR_SLOTS
+#define AWS_MAX_ERROR_SLOTS 16
+#endif
+
+/* to multiply by 16, you shift left by 5 */
+#define SLOT_MULTIPLIER_SHIFTS 0x05
```
marcomagdy: to multiply by 16, you shift left by 4

JonathanHenson: ugh..... what I get for counting bits

JonathanHenson: that was dumb anyways, since I moved it to a static const int, using normal multiplication is fine.

PR #2, opened 2018-01-18
Cross platform primitives

JonathanHenson:
Posix and Windows:

Threads, mutex, and clock.

For windows mutex I opted for CriticalSections instead of a fullblown mutex.


source/posix/clock.c
```diff
@@ -0,0 +1,42 @@
...
+}
+
+int aws_sys_clock_get_ticks(uint64_t *timestamp) {
+    struct timespec ts;
+    int ret_val = clock_gettime(CLOCK_REALTIME, &ts);
+    if(ret_val) {
+        return aws_raise_error(AWS_ERROR_CLOCK_FAILURE);
+    }
+
+    *timestamp = (uint64_t)((ts.tv_sec * NS_PER_SEC) + ts.tv_nsec);
```
marcomagdy: I think this might be UB if tv_sec is not uint64_t.
Better to cast one of the operands to uint64_t first before the multiplication.

source/windows/clock.c
```diff
@@ -0,0 +1,49 @@
...
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/clock.h>
+#include <Windows.h>
+
+static const LONGLONG MUS_PER_SEC = 1000000;
+static const LONGLONG NS_PER_MUS = 1000;
+static const LONGLONG FILE_TIME_TO_NS = 100;
+static const LONGLONG EC_TO_UNIX_EPOCH = 11644473600LL;
```
marcomagdy: What's EC? add a comment?
Why use LONGLONG instead of the more explicit int32_t or int64_t?

source/posix/clock.c
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+#include <aws/common/clock.h>
+#include <time.h>
+
+static const long NS_PER_SEC = 1000000000;
+
+int aws_high_res_clock_get_ticks(uint64_t *timestamp) {
+    struct timespec ts;
+    int ret_val = clock_gettime(CLOCK_MONOTONIC, &ts);
```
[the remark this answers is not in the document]
bdonlan: clock_gettime exists since OSX 10.12.

include/aws/common/mutex.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include <aws/common/common.h>
+#ifdef _WIN32
+#include <Windows.h>
+#else
+#include <pthread.h>
+#endif
+
+struct aws_mutex {
+    struct aws_allocator *allocator;
```
bdonlan: Is this needed?

JonathanHenson: no, a long time ago we discussed keeping it there for long term api compatibility.

JonathanHenson: but at the moment, thread and mutex don't need allocators

source/windows/clock.c
JonathanHenson: Windows has a different epoch than linux epoch. That's the conversion number.

LONGLONG is what's in the windows api reference and "technically" LONGLONG is defined with a precise width (only a minimum)

source/posix/clock.c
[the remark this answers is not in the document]
JonathanHenson: fortunately, supporting ancient apple operating systems is something anyone rarely needs to do.

source/posix/thread.c
```diff
@@ -0,0 +1,153 @@
...
+
+    if(attr_return == EAGAIN)
+        return aws_raise_error(AWS_ERROR_THREAD_INSUFFICIENT_RESOURCE);
+
+    if(attr_return == EPERM) {
+        return aws_raise_error(AWS_ERROR_THREAD_NO_PERMISSIONS);
+    }
+
+    if(attr_return == ENOMEM) {
+        return AWS_ERROR_OOM;
```
JonathanHenson: woops! that should be return aws_raise_error(AWS_ERROR_OOM)

source/posix/clock.c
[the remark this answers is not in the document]
marcomagdy: It doesn't build on my machine. I'm running El Capitan (10.11.6) not ancient. I'm assuming this library will be used in client SDKs too, so we need to support those platforms.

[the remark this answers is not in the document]
JonathanHenson: well, crap... ok, I'll have to add it.

tests/mutex_test.c
```diff
@@ -0,0 +1,82 @@
...
+
+    struct aws_thread thread;
+    aws_thread_init(&thread, allocator);
+    ASSERT_SUCCESS(aws_thread_create(&thread, mutex_thread_fn, &mutex_data, 0), "thread creation failed with error %d", aws_last_error());
+
+    int thread_2_wrote_first = mutex_data.thread_2_written;
+    mutex_data.thread_1_written = 1;
+
+    while (!mutex_data.thread_2_detected_write_from_thread_1) continue;
+    thread_2_wrote_first |= mutex_data.thread_2_written;
```
bdonlan: what is this supposed to do? there's no synchronization with the thread 2 write so this could be either true or false (thus losing the original observation of whether thread 2 wrote first)

include/aws/common/thread.h
```diff
@@ -0,0 +1,100 @@
...
+#endif
+
+typedef enum aws_thread_detach_state {
+    AWS_THREAD_JOINABLE = 1,
+    AWS_THREAD_DETACHED = 2,
+} aws_thread_detach_state;
+
+struct aws_thread_options {
+    aws_thread_detach_state detach_state;
+    size_t stack_size;
```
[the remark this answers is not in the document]
JonathanHenson: if you pass NULL to the aws_thread_create, then defaults are used. Should we just add that to the documentation, or would you want something more?

include/aws/common/thread.h
```diff
@@ -0,0 +1,100 @@
...
+ */
+AWS_COMMON_API int aws_thread_detach(struct aws_thread *thread);
+
+/**
+ * Joins the calling thread to a thread instance. Returns when thread is finished.
+ */
+AWS_COMMON_API int aws_thread_join(struct aws_thread *thread);
+
+/**
+ * Cleans up the os-level thread, and deallocates itself.
```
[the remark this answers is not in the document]
JonathanHenson: {
    std::thread t(&myfunc, &my_data);
    t.detach();
}

What's the difference?

source/posix/mutex.c
```diff
@@ -0,0 +1,65 @@
...
+int aws_mutex_init(struct aws_mutex *mutex, struct aws_allocator *allocator) {
+
+    mutex->allocator = allocator;
+    pthread_mutexattr_t attr;
+    pthread_mutexattr_init(&attr);
+    pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_NORMAL);
+    pthread_mutex_init(&mutex->mutex_handle, &attr);
+    pthread_mutexattr_destroy(&attr);
+
+    return AWS_ERROR_SUCCESS;
```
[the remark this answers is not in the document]
JonathanHenson: sure, it was more convenient since success is 0, but maybe we should just have a separate constant called AWS_OP_SUCCESS and AWS_OP_FAILURE?

source/posix/thread.c
```diff
@@ -0,0 +1,153 @@
...
+
+static void *thread_wrapper_fn (void *arg_pass) {
+    struct aws_thread_wrapper *wrapper = (struct aws_thread_wrapper *)arg_pass;
+    wrapper->func(wrapper->arg);
+    return 0;
+}
+
+void aws_thread_clean_up (struct aws_thread *thread) {
+    /* this likely won't do anything if the thread has already terminated, but it shouldn't hurt to send a cancel signal just in case.
+       should we join here? */
```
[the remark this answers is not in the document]
JonathanHenson: well good. So just delete the comment?

source/posix/thread.c
```diff
@@ -0,0 +1,153 @@
...
+
+int aws_thread_detach(struct aws_thread *thread) {
+    int err_no = pthread_detach(thread->thread_id);
+
+    if (err_no) {
+        if (err_no == EINVAL) {
+            return aws_raise_error(AWS_ERROR_THREAD_NOT_JOINABLE);
+        }
+        else if (err_no == ESRCH) {
+            return aws_raise_error(AWS_ERROR_THREAD_NO_SUCH_THREAD_ID);
```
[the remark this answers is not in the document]
JonathanHenson: not joinable... probably, I can't imagine the latter would ever occur given our API

source/windows/mutex.c
```diff
@@ -0,0 +1,46 @@
...
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/mutex.h>
+
+int aws_mutex_init(struct aws_mutex *mutex, struct aws_allocator *allocator) {
+    mutex->allocator = allocator;
+    InitializeCriticalSection(&mutex->mutex_handle);
```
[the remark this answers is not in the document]
JonathanHenson: There's no error code for InitializeCriticalSection

https://msdn.microsoft.com/en-us/library/windows/desktop/ms683472(v=vs.85).aspx

I'll look into SRW locks

tests/clock_test.c
```diff
@@ -0,0 +1,46 @@
...
+*/
+
+#include <aws/common/clock.h>
+#include <aws_test_harness.h>
+
+static int test_high_res_clock_increments(struct aws_allocator *allocator, void *ctx) {
+    uint64_t ticks = 0, prev = 0;
+
+    for (unsigned i = 0; i < 100; ++i) {
+        ASSERT_SUCCESS(aws_high_res_clock_get_ticks(&ticks), "High res get ticks failed with error %d", aws_last_error());
```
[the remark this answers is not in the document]
JonathanHenson: do you have an idea for a better test? Should I just add some sleep() calls?

tests/clock_test.c
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    return 0;
+}
+
+static int test_sys_clock_increments(struct aws_allocator *allocator, void *ctx) {
+    uint64_t ticks = 0, prev = 0;
+
+    for (unsigned i = 0; i < 100; ++i) {
+        ASSERT_SUCCESS(aws_sys_clock_get_ticks(&ticks), "Sys clock res get ticks failed with error %d", aws_last_error());
```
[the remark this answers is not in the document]
JonathanHenson: good point... any ideas for a better test?

tests/mutex_test.c
JonathanHenson: It's supposed to detect when thread 2 has run and hit the next acquire call. I can't force the mutex to not be reentrant on Windows (AFAIK) and this is the best I could come up with.

source/windows/mutex.c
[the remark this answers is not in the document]
JonathanHenson: "An SRW lock is the size of a pointer. The advantage is that it is fast to update the lock state. The disadvantage is that very little state information can be stored, so SRW locks cannot be acquired recursively"

Boom! sold

include/aws/common/thread.h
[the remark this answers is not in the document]
bdonlan: It's a matter of documentation. The thread continues running (it's not forcibly terminated). The handle to the thread is cleaned up.

source/posix/thread.c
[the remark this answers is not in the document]
bdonlan: Deciding to detach a thread that's already detached seems like it's better as an idempotent operation to me.

source/windows/mutex.c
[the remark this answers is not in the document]
bdonlan: Ah - critical section OOMs result in a structured exception thrown: https://blogs.msdn.microsoft.com/oldnewthing/20140911-00/?p=44103

tests/clock_test.c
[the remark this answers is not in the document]
bdonlan: That might be best.

tests/clock_test.c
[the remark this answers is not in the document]
bdonlan: I'd just verify that it changes after a sleep.

tests/mutex_test.c
bdonlan: The while is fine, but the subsequent observation of thread_2_written is unsynchronized.

A more robust test is simply: Two threads race to lock, read a volatile counter into a local, increment by one, write back, unlock. They each report the number of times they incremented at the end. Check that the counter equals the sum of the increments.

source/windows/mutex.c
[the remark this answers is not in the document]
JonathanHenson: I'm moving to SRW regardless because a.) we don't have to support XP or server 2003 (and windows is a different scenario than supporting legacy linux because it's a commercial OS that reaches end of life), and b.) it's not reentrant which makes testing a hell of a lot easier.

tests/mutex_test.c
JonathanHenson: I think I can get all of this with non-reentrant mutexes.

source/windows/mutex.c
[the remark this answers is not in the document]
JonathanHenson: However, what baffles the hell out of me is that .NET 3.5 supports SRW and .NET 3.5 is deployed to windows xp where the api isn't available. My guess is that nt.dll exports it, and since dotnet is doing a dynamic bind, it picks it up, but it wasn't added to the headers until the windows sdk for vista.

include/aws/common/thread.h
```diff
@@ -0,0 +1,100 @@
...
+
+struct aws_thread_wrapper {
+    void (*func)(void *arg);
+    void *arg;
+};
+
+struct aws_thread {
+    struct aws_allocator *allocator;
+
+    struct aws_thread_wrapper thread_wrapper;
```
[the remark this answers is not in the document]
JonathanHenson: I had a lengthy reply on why I disagreed, and then saw the reason you are right. Good callout, will change.

source/posix/thread.c
```diff
@@ -0,0 +1,153 @@
...
+
+            if(attr_return) {
+                goto cleanup;
+            }
+        }
+    }
+
+    thread->thread_wrapper.func = func;
+    thread->thread_wrapper.arg = context;
+    attr_return = pthread_create(&thread->thread_id, attributes_ptr, thread_wrapper_fn, &thread->thread_wrapper);
```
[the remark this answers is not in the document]
JonathanHenson: I must be missing something here:

if(options) {
        attr_return = pthread_attr_init(&attributes);

        if(attr_return) {
            goto cleanup;
        }

        attributes_ptr = &attributes;


....

attributes_ptr is initialized.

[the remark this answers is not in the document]
JonathanHenson: if it is NULL in this case, that follows the pthread spec.... meaning NULL means use the defaults.

source/posix/thread.c
```diff
@@ -0,0 +1,153 @@
...
+
+    if(options) {
+        attr_return = pthread_attr_init(&attributes);
+
+        if(attr_return) {
+            goto cleanup;
+        }
+
+        attributes_ptr = &attributes;
+        int detach_state = 0;
```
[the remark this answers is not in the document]
JonathanHenson: ? I'm not sure what your point is on this one.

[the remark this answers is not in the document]
bdonlan: In other words, what if PTHREAD_CREATE_DETACHED is defined to be zero? In this case the if at line 64 will be false, and so we won't set this (non-default) flag.

source/posix/thread.c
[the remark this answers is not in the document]
bdonlan: Ah, whoops, I missed that line somehow. This is why I usually separate declarations from statements with a blank line :)

source/posix/thread.c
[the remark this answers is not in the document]
JonathanHenson: ah! got it, good catch.

source/posix/thread.c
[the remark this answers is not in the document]
JonathanHenson: I just updated to not care about the error code, since invalid thread id shouldn't even be possible and not joinable doesn't matter since we'll just treat the detach call as idempotent.

include/aws/common/thread.h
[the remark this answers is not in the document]
bdonlan: What if I only want to configure one of the aws_thread_options parameters and leave the rest (including ones only defined in new releases created after my code was written) at reasonable defaults?

include/aws/common/common.h
```diff
@@ -78,12 +78,26 @@ AWS_COMMON_API void aws_load_error_strings(void);
...
+    AWS_ERROR_THREAD_INSUFFICIENT_RESOURCE,
+    AWS_ERROR_THREAD_NO_PERMISSIONS,
+    AWS_ERROR_THREAD_NOT_JOINABLE,
+    AWS_ERROR_THREAD_NO_SUCH_THREAD_ID,
+    AWS_ERROR_THREAD_DEADLOCK_DETECTED,
+    AWS_ERROR_MUTEX_NOT_INIT,
+    AWS_ERROR_MUTEX_TIMEOUT,
+    AWS_ERROR_MUTEX_CALLER_NOT_OWNER,
+    AWS_ERROR_MUTEX_FAILED,
+    AWS_ERROR_CLOCK_FAILURE,
```
bdonlan: I'd suggest having an AWS_ERROR_UNKNOWN instead of an AWS_[subsystem]_FAILURE generic error code.

include/aws/common/thread.h
```diff
@@ -0,0 +1,99 @@
...
+struct aws_thread_options {
+    aws_thread_detach_state detach_state;
+    size_t stack_size;
+};
+
+struct aws_thread {
+    struct aws_allocator *allocator;
+
+#ifdef _WIN32
+    HWND thread_handle;
```
bdonlan: this should be a `HANDLE` not a `HWND`.

include/aws/common/thread.h
```diff
@@ -0,0 +1,99 @@
...
+ * Initializes a new platform specific thread object struct (not the os-level thread itself).
+ */
+AWS_COMMON_API int aws_thread_init(struct aws_thread *thread, struct aws_allocator *allocator);
+
+/**
+ * Creates an OS level thread and associates it with func. context will be passed to func when it is executed.
+ * options will be applied to the thread if they are applicable for the platform.
+ */
+AWS_COMMON_API int aws_thread_create(struct aws_thread *thread, void(*func)(void *arg),
+                                                      void *context, struct aws_thread_options *options);
```
bdonlan: Name the callback argument the same in the prototype and the outer call (i.e. call it either `arg` or `context` but be consistent)

include/aws/common/thread.h
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Detaches thread from this instance of thread. You probably want to call aws_core_thread_destroy()
+ * immediately afterwards.
+ */
+AWS_COMMON_API int aws_thread_detach(struct aws_thread *thread);
+
+/**
+ * Joins the calling thread to a thread instance. Returns when thread is finished.
+ */
+AWS_COMMON_API int aws_thread_join(struct aws_thread *thread);
```
bdonlan: Timeout?

source/common.c
```diff
@@ -16,6 +16,11 @@
 #include <aws/common/common.h>
 #include <stdlib.h>
 
+#ifdef _MSC_VER 
+#pragma warning( push )
+#pragma warning( disable : 4100)
```
bdonlan: What warning are we suppressing?

source/posix/mutex.c
```diff
@@ -0,0 +1,78 @@
...
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/mutex.h>
+#include <errno.h>
+
+void aws_mutex_clean_up(struct aws_mutex *mutex) {
+    aws_mutex_release(mutex);
```
bdonlan: Why are we implicitly unlocking the mutex?

Note that the OS X manpage for `pthread_mutex_unlock` documents:

> Calling pthread_mutex_unlock() with a mutex that the calling thread does not hold will result in undefined behavior.

source/posix/thread.c
```diff
@@ -0,0 +1,169 @@
...
+
+static void *thread_fn(void *arg) {
+    struct thread_wrapper *wrapper = (struct thread_wrapper *)arg;
+    wrapper->func(wrapper->arg);
+    aws_mem_release(wrapper->allocator, wrapper);
+    return NULL;
+}
+
+void aws_thread_clean_up (struct aws_thread *thread) {
+    /* this likely won't do anything if the thread has already terminated, but it shouldn't hurt to send a cancel signal just in case.*/
```
bdonlan: Why are we implicitly cancelling the thread? Does the user normally expect that destroying a thread handle would affect the thread state? Who actually writes async-cancel-safe multithreaded code?

source/posix/thread.c
```diff
@@ -0,0 +1,169 @@
...
+struct thread_wrapper {
+    struct aws_allocator *allocator;
+    void(*func)(void *arg);
+    void *arg;
+};
+
+static void *thread_fn(void *arg) {
+    struct thread_wrapper *wrapper = (struct thread_wrapper *)arg;
+    wrapper->func(wrapper->arg);
+    aws_mem_release(wrapper->allocator, wrapper);
```
bdonlan: You could copy the `thread_wrapper` to the stack and immediately deallocate it.

source/windows/mutex.c
```diff
@@ -0,0 +1,46 @@
...
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/mutex.h>
+
+int aws_mutex_init(struct aws_mutex *mutex, struct aws_allocator *allocator) {
+    mutex->allocator = allocator;
+    mutex->mutex_handle.Ptr = 0;
```
bdonlan: Where's the `InitializeSRWLock`? https://msdn.microsoft.com/en-us/library/windows/desktop/ms683483(v=vs.85).aspx

source/windows/thread.c
```diff
@@ -0,0 +1,93 @@
...
+
+#include <aws/common/thread.h>
+
+struct thread_wrapper {
+    struct aws_allocator *allocator;
+    void(*func)(void *arg);
+    void *arg;
+};
+
+static DWORD thread_wrapper_fn(LPVOID arg) {
```
bdonlan: This needs to be `static DWORD WINAPI`. The `WINAPI` modifies the calling convention for the function.

source/windows/thread.c
```diff
@@ -0,0 +1,93 @@
...
+    thread->allocator = allocator;
+
+    return AWS_OP_SUCCESS;
+}
+
+int aws_thread_create(struct aws_thread *thread, void(*func)(void *arg), void *context, struct aws_thread_options *options) {
+
+    SIZE_T stack_size = 0;
+
+    if (options && options->stack_size > 0) {
```
bdonlan: Where do we check the detached state in the options?

Also - what are the semantics guaranteed by the detached state? Are we still required to clean up the `aws_thread`?

include/aws/common/common.h
JonathanHenson: CLOCK or MUTEX failed is more useful when you queue up the errors or send out metrics than a generic unknown though.

include/aws/common/thread.h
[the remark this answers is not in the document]
JonathanHenson: I mean, we can have a function that is called:

struct aws_thread_options *aws_thread_options_platform_defaults(void);

if you'd like.

tests/mutex_test.c
```diff
@@ -0,0 +1,108 @@
...
+    aws_mutex_init(&mutex_data.mutex, allocator);    
+
+    struct aws_thread thread;
+    aws_thread_init(&thread, allocator);
+    ASSERT_SUCCESS(aws_thread_create(&thread, mutex_thread_fn, &mutex_data, 0), "thread creation failed with error %d", aws_last_error());
+    int finished = 0;
+    int increments = 0;
+    while (!finished) {
+        aws_mutex_acquire(&mutex_data.mutex);
+        /*add some fairness for thread startup time.*/
```
bdonlan: Not sure I understand what this comment is referring to.

source/common.c
JonathanHenson: unused function argument warning for windows, since visual c doesn't technically have a "c" compiler, it carries this one over from C++.

source/posix/mutex.c
JonathanHenson: Will remove.

source/common.c
JonathanHenson: I'll add a comment.

source/posix/thread.c
JonathanHenson: good call, will update

source/posix/thread.c
JonathanHenson: fair enough, will remove.

source/windows/mutex.c
JonathanHenson: you only call InitializeSRWLock on dynamically allocated memory. If it's statically allocated, it just gets initialized to zero.

"The InitializeSRWLock function is used to initialize a SRW lock dynamically. To initialize the structure statically, assign the constant SRWLOCK_INIT"

SRWLOCK_INIT is just {0}, but that doesn't compile so I manually did the init.

source/windows/thread.c
JonathanHenson: good catch.

source/windows/thread.c
JonathanHenson: I don't think the actual thread state should be in the options, but I can add the state to aws_thread for checking.  You are always required to cleanup aws_thread. The semantics should match std::thread

source/windows/mutex.c
bdonlan: You don't actually know if this is statically allocated or not. However, since this should be ABI-stable, leave a comment explaining that this is a faster version of `InitializeSRWLock`.

tests/mutex_test.c
JonathanHenson: If thread takes too long to startup, the test doesn't acquire the lock very many times. This is added to make sure thread 1 doesn't do all the work before thread 2 gets a chance.

source/windows/thread.c
bdonlan: See thread.h:L28. What semantics does the `aws_thread_detach_state` enumeration have, if we can ignore it?

tests/mutex_test.c
bdonlan: You're just acquiring and releasing here, so if thread 1 starts up before your acquire you still have this problem. Acquire the mutex before launching any threads if you want to do this.

source/windows/mutex.c
JonathanHenson: The way I parse the meaning for "static" from the msdn docs, is it is a member of the struct as a value, not as a pointer to a different allocation.

include/aws/common/thread.h
JonathanHenson: no timeout: I'm, going for these semantics:

http://en.cppreference.com/w/cpp/thread/thread/join

include/aws/common/thread.h
JonathanHenson: woops, will update

include/aws/common/thread.h
JonathanHenson: will update

source/windows/mutex.c
bdonlan: Static allocation just means you can allocate a SRW lock as a statically allocated structure without needing to coordinate an initialization call at startup. The documentation doesn't actually say that setting a heap SRW lock to 0 performs initialization, but the ABI implies it, so it'll have to work, at least until windows comes up with a new platform with a different initializer.

source/windows/thread.c
JonathanHenson: I don't think we ignore it. That goes in the options for creation, and will be applied then. As far as the current state of the thread, I propose, we put an instance of aws_thread_detach_state there as well.

tests/mutex_test.c
JonathanHenson: thread 1 is already started (it's the main thread). This will keep going until thread 2 acquires the lock once and then they start competing for the lock.

source/windows/thread.c
bdonlan: Where exactly do we use `aws_thread_detach_state` in the win32 thread creation path? And again, what are the semantics of setting detached state on creation?

JonathanHenson: well, that's just an oversight on my part, not a statement about the API. 

I'll add a if(detach) CloseHandle() to the windows version.

bdonlan: Sure, but what's the value of specifying detachable? I still need to clean up the `aws_thread`. What's the point in having a flag that says "Make the thing that is returned less useful, but don't make it any easier for me to use it"?

bdonlan: I'm questioning the value of detach-on-create when this doesn't fully absolve the caller of any cleanup activities.

The point of the posix detached state is that you don't need to detach the thread later. On windows, this isn't an option, and instead you close the thread handle explicitly. In our API, if we are going to offer a 'detach-on-creation' option, this should - as with posix - absolve the caller from any need to perform any cleanup activities after the thread is created.

If we're not prepared to absolve the caller of any cleanup activities, then being able to pass a detach-on-creation option is useless - the _only_ function it has is to break `join`. Why would anyone pass an option that breaks a feature and does nothing else?

JonathanHenson: sure, I get your point now. 

Maybe we can just remove it from thread creation, that's more inline with 

std::thread semantics anyways.

include/aws/common/thread.h
```diff
@@ -49,6 +50,12 @@ struct aws_thread {
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+/**
+ * Returns an instance of system default thread options.
```
bdonlan: You need to specify how this is allocated, who owns cleanup, and whether it can be safely modified in-place.

source/windows/mutex.c
```diff
@@ -17,6 +17,8 @@
 
 int aws_mutex_init(struct aws_mutex *mutex, struct aws_allocator *allocator) {
     mutex->allocator = allocator;
+    /* A more efficient variant of InitializeSRWLock.
+       See https://msdn/microsoft.com/en-us/library/windows/desktop/ms683483(v=vs.85).aspx */
```
bdonlan: Link is typoed. Also the linked URL does not explain why setting `.Ptr = 0` is a legal way to initialize the lock.

Why aren't we just calling InitializeSRWLock anyway?

source/windows/thread.c
```diff
@@ -14,29 +14,40 @@
 */
 
 #include <aws/common/thread.h>
+#include <assert.h>
+
+static struct aws_thread_options default_options = {
```
bdonlan: Having a static mutable struct is just asking for someone to start modifying it from multiple threads.

include/aws/common/thread.h
```diff
@@ -57,17 +64,23 @@ AWS_COMMON_API int aws_thread_init(struct aws_thread *thread, struct aws_allocat
 /**
  * Creates an OS level thread and associates it with func. context will be passed to func when it is executed.
  * options will be applied to the thread if they are applicable for the platform.
+ * You must either call join or detach after creating the thread and before calling clean_up. 
  */
 AWS_COMMON_API int aws_thread_create(struct aws_thread *thread, void(*func)(void *arg),
```
bdonlan: Potential point of confusion: I can see people not knowing that they need to call `_init` before `_create`. After all, shouldn't the create function initialize whatever it's creating?

I'd call it something different, like `_launch` instead. Also, you can avoid a lot of the compatibility/default issues with `aws_thread_options` by having setters on the thread structure instead.

source/windows/thread.c
```diff
@@ -46,38 +57,49 @@ int aws_thread_create(struct aws_thread *thread, void(*func)(void *arg), void *c
...
+    }
 
-    WaitForSingleObject(thread->thread_handle, INFINITE);
-
-    return AWS_OP_SUCCESS;
+    return aws_raise_error(AWS_ERROR_THREAD_NOT_JOINABLE);
 }
 
 void aws_thread_clean_up(struct aws_thread *thread) {
+    /* cleaning up a joinable thread is undefined. Don't do it. Either join or detach. The cpp stdlib actually calls std::terminate in this case */
```
bdonlan: Whatever the C++ stdlib does is irrelevant to our API. Why shouldn't we allow you to clean up a joinable thread? We clearly know if it's joinable or not, and we know how to detach it. And, in fact, if we remove this assert it'll work just fine. What future-proofing are we trying to accomplish here?

source/windows/thread.c
```diff
@@ -89,5 +111,5 @@ void aws_thread_current_sleep(uint64_t nanos) {
     /* We don't really have a better option here for windows that isn't super complex AND we don't have a use case
      * yet where we should have sleeps anywhere other than for context switches and testing. When that time arises
      * put the effort in here. */
-    Sleep(nanos / 1000000);
+    Sleep((DWORD)(nanos / 1000000));
```
bdonlan: Is this cast needed?

tests/aws_test_harness.h
```diff
@@ -88,8 +88,8 @@ static int total_failures;
 #define ASSERT_FALSE(condition, format, ...) do { if(condition) { FAIL(format, ## __VA_ARGS__); } } while(0)
 
 #define ASSERT_SUCCESS(condition, format, ...) do { if(condition) { FAIL(format, ## __VA_ARGS__); } } while(0)
-#define ASSERT_FAILS(condition, format, ...) do { if(condition == -1) { FAIL(format, ## __VA_ARGS__); } } while(0)
-#define ASSERT_ERROR(error, condition, format, ...) do { if(condition == -1) { FAIL(format, ## __VA_ARGS__); } if(aws_last_error_code() != error) { FAIL(format, ## __VA_ARGS__) } } while(0)
+#define ASSERT_FAILS(condition, format, ...) do { if(condition != -1) { FAIL(format, ## __VA_ARGS__); } } while(0)
```
bdonlan: Use the macros instead of `-1`?

JonathanHenson: ok

include/aws/common/thread.h
JonathanHenson: ok, I'll add,

statically allocated, copy if you want to mutate.

include/aws/common/thread.h
JonathanHenson: fair enough.

include/aws/common/thread.h
bdonlan: Then this should be a const pointer.

source/windows/mutex.c
JonathanHenson: "To initialize the structure statically, assign the constant SRWLOCK_INIT to the structure variable." isn't an explanation?

"Why aren't we just calling InitializeSRWLock anyway?"

Because of this language:
"The InitializeSRWLock function is used to initialize a SRW lock dynamically."

What does this mean? Does it do an allocation? It's marked as an output parameter.

source/windows/thread.c
JonathanHenson: the assert is because if you aren't either calling join or detach before cleaning up the thread it's a bug and quite possibly a resource leak.

source/windows/thread.c
JonathanHenson: yeah, Sleep takes DWORD, nanos / 1000000 is uint64_t

source/windows/mutex.c
bdonlan: It cannot do an allocation because it is documented that SRW locks are not required to be cleaned up or deallocated. I think you're overthinking this - SRW locks don't care what kind of memory they're used in; you can use InitializeSRWLock on a static variable if you want.

In any case, if you're doing something that isn't in the docs - i.e. setting a random member (which is not documented) to a constant zero (also not documented as a valid value for that undocumented member), you need to have a really convincing argument for why this is 1) permitted and 2) better than the documented methods. You can't just link a MSDN page which says to do something different from what you're actually doing.

source/windows/thread.c
bdonlan: I don't buy that argument. The idea that 'join' releases resources is a very pthread-specific notion. In windows, waiting for a handle does not release it. In java, `Thread.join` does not release any resources.

I would argue that `detach` is superfluous (why not just use `clean_up`?) and `join` should be idempotent.

source/windows/mutex.c
JonathanHenson: I'm doing exactly what the page says to do:

lock = SRWLOCK_INIT;

is literally identical to what I'm doing.

The macro replacement of SRWLOCK_INIT is {0}

I'm fine using the InitializeCall, but the attempt here is to follow the documentation, not the opposite.

JonathanHenson:
I literally do not care, if you want it to be something else we'll change it. Bryan what's your vote on the matter.

source/windows/mutex.c
bdonlan: To be clear, the documentation says to use `SRWLOCK_INIT` for static initialization, which is a macro which could be redefined in the future. For example, it could be redefined to be `{1}` on a new platform, or the name or type of the member could change. While existing code on x86 or x86_64 windows must remain ABI compatible, code built on newer SDKs might break, or code on (insert fancy new mobile architecture) might break as well.

When something says `static initialization` in this context, it means initializing something at compile time (vs initializing it at runtime). The documentation doesn't say anything that would imply that `InitializeSRWLock` would be unacceptable in any context.

bdonlan:
`lock`/`unlock` seems clearer to me. It's notable that the underlying platform primitives are also `lock`/`unlock`.

JonathanHenson:
fine, lock/unlock it is

source/windows/mutex.c
JonathanHenson: like I said, I'm fine using the call.

It's not the words "static initialization" that is confusing, it's the words: "The InitializeSRWLock function is used to initialize a SRW lock dynamically" that is ambiguous. It implies that it wants to do an allocation. 

Anyways, let's close this thread, InitializeSRWLock is fine.

source/windows/thread.c
JonathanHenson: ok, deleting the "detached" state from the thread, and deleting the detach function, moving to clean up.

source/posix/thread.c
```diff
@@ -39,13 +39,14 @@ static void *thread_fn(void *arg) {
     return NULL;
 }
 
-struct aws_thread_options *aws_default_thread_options(void) {
+const struct aws_thread_options *aws_default_thread_options(void) {
```
bdonlan: I'm going to point out here that this API is fundamentally incompatible with having ABI forwards compatibility with extending structures with new members - existing code doesn't know how to copy the structure, and the technique of sticking the structure size in the first member won't even work with this.

This is a place where we need to commit to a model of how we handle ABI compat concerns before committing too much to this API, I think.

JonathanHenson: It can be a fair tradeoff IFF we get the thread options right before going GA, since that's not something that changes very much.

PR #3, opened 2018-01-20
Array list and unit tests.

JonathanHenson:
Templated array list implementation.


include/aws/common/array_list.h
```diff
@@ -0,0 +1,265 @@
...
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/common.h>
+#include <aws/common/error.h>
+#include <stdint.h>
+#include <string.h>
+
```
bdonlan: Making this a big template thing seems overly complicated. Why not just stick the structure size into the array list structure, and return `void *`s?

include/aws/common/array_list.h
```diff
@@ -0,0 +1,265 @@
...
+    }                                                                                                                  \
+                                                                                                                       \
+    list->current_size = 0;                                                                                            \
+    list->length = 0;                                                                                                  \
+    list->data = 0;                                                                                                    \
+    list->alloc = 0;                                                                                                   \
+}                                                                                                                      \
+                                                                                                                       \
+/**
+* Pushes a the dereferenced value of val onto the end of internal list->
```
bdonlan: `a the`, `->`?

include/aws/common/array_list.h
```diff
@@ -0,0 +1,265 @@
...
+*/                                                                                                                     \
+static inline int aws_##name##_array_list_push_back(struct aws_##name##_array_list *list, T *val) {                    \
```
bdonlan: `const T *`

include/aws/common/array_list.h
```diff
@@ -0,0 +1,265 @@
...
+#define aws_typed_array_list_declaration(name, T)                                                                      \
+struct aws_##name##_array_list {                                                                                       \
+    struct aws_allocator *alloc;                                                                                       \
+    size_t current_size;                                                                                               \
+    size_t length;                                                                                                     \
+                                                                                                                       \
+    T *data;                                                                                                           \
+};
+
+/* To define your typed array list, use this macro (most likely in your .c file) */
```
bdonlan: Both 'declaration' and 'definition' will cause compile errors if inserted twice in the same translation unit. Also, the functions are `static inline` so need to be included in each translation unit that use them anyway. Why not combine them?

include/aws/common/array_list.h
```diff
@@ -0,0 +1,265 @@
...
+}                                                                                                                      \
+                                                                                                                       \
+/**
+* Pushes a the dereferenced value of val onto the end of internal list->
+*/                                                                                                                     \
+static inline int aws_##name##_array_list_push_back(struct aws_##name##_array_list *list, T *val) {                    \
+                                                                                                                       \
+    if (list->length == list->current_size) {                                                                          \
+        if (list->alloc) {                                                                                             \
+            size_t new_size = list->current_size << 1;                                                                 \
```
bdonlan: Overflow check, zero length check.

include/aws/common/array_list.h
```diff
@@ -0,0 +1,265 @@
...
+static inline int aws_##name##_array_list_copy(const struct aws_##name##_array_list *from,                             \
+                        struct aws_##name##_array_list *to) {                                                          \
+    if (to->current_size >= from->length) {                                                                            \
+        memcpy(to->data, from->data, from->length * sizeof(T));                                                        \
+        to->length = from->length;                                                                                     \
+        return AWS_OP_SUCCESS;                                                                                         \
+    }                                                                                                                  \
+                                                                                                                       \
+    return aws_raise_error(AWS_ERROR_LIST_DEST_COPY_TOO_SMALL);                                                        \
```
bdonlan: Why doesn't this allocate space for a dynamic list?

include/aws/common/array_list.h
JonathanHenson: Because the whole semantic is to copy the data whatever it is. 

Casing an int to void * is cumbersome and error prone. Then we have to do memcpy everywhere and casts to uint8_t etc.... This is just safer if we are going to be doing value semantics.

include/aws/common/array_list.h
JonathanHenson: DECLARATION can go in your header file without bloating it
DEFINITION can be used in your .c file

They'll only be a compile error if you declare same name and type multiple times, and in that case I'm not sure I care.

include/aws/common/array_list.h
JonathanHenson: yep, good catch.

include/aws/common/array_list.h
JonathanHenson: this one is complicated. That doesn't work when T is something like `void *` c const rules are weird.

include/aws/common/array_list.h
JonathanHenson: good catch.

include/aws/common/array_list.h
JonathanHenson: And this is a rare occasion, where it's getting used in lots of places, and almost always with value semantics.

include/aws/common/array_list.h
JonathanHenson: Fair. Will add.

include/aws/common/array_list.h
bdonlan: We already use pointer references in all get/put sites currently, looks like. It wouldn't make use sites any more cumbersome as far as I can see.

include/aws/common/array_list.h
bdonlan: What if I want to expose an array list as something I pass as an out/in parameter to a function in a header file? Who is responsible for declaring it?

If I do declare it in my header, only my one .c file can use that type, because I can only 'define' it once in my .c file, right? And since the definitions are all static, I can't even redeclare the functions manually to use them elsewhere.

Normally with C, things can be "declared" more than once, but only "defined" once. Let's not make our own definitions of those terms, and let's make sure we handle the very basic use case of "I want this structure to be passed between two .c files and manipulated by both".

include/aws/common/array_list.h
bdonlan: That means if we're passed in a `const void *` and pass it along to push_back it'll fail with a really weird error. Now I need to cast away the constness. There's some real hidden complexity here with this pseudo-templating.

We're not writing C++. We're writing C. When you have a generic datastructure in C, you use `void *`s and `memcpy`. You don't go and reimplement C++ templates in your header files, because strange non-idiomatic stuff like this happens.

include/aws/common/array_list.h
JonathanHenson: eh, if we get rid of the template this whole thing can just go away anyways....

include/aws/common/array_list.h
JonathanHenson: I'll redo it.

include/aws/common/array_list.h
```diff
@@ -0,0 +1,126 @@
...
+    void *data;
+};
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    /**
+    * Initializes an array list with an array of void *. In this mode, the array size will grow
```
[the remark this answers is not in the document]
JonathanHenson: yep, will remove

include/aws/common/array_list.h
```diff
@@ -0,0 +1,126 @@
...
+    */
+    AWS_COMMON_API void aws_array_list_clean_up(struct aws_array_list *list);
+
+    /**
+    * Pushes the memory pointed to by val onto the end of internal list
+    */
+    AWS_COMMON_API int aws_array_list_push_back(struct aws_array_list *list, const void *val);
+
+    /**
+    * Returns the element at the front of the list if it exists. If list is empty, AWS_CORE_LIST_ERROR_EMPTY will be raised
```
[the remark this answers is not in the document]
JonathanHenson: will update

source/array_list.c
```diff
@@ -0,0 +1,247 @@
...
+void aws_array_list_clean_up(struct aws_array_list *list) {
+    if (list->alloc && list->data) {
+        aws_mem_release(list->alloc, list->data);
+    }
+
+    list->current_size = 0;
+    list->item_size = 0;
+    list->length = 0;
+    list->data = NULL;
+    list->alloc = 0;
```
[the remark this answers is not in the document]
JonathanHenson: Good catch, I'd already moved over to that, but obviously missed some in the port.

source/array_list.c
```diff
@@ -0,0 +1,247 @@
...
+                return aws_raise_error(AWS_ERROR_LIST_EXCEEDS_MAX_SIZE);
+            }
+
+            void *temp = aws_mem_acquire(list->alloc, new_size);
+
+            if (!temp) {
+                return aws_raise_error(AWS_ERROR_OOM);
+            }
+
+            memset((uint8_t *)temp + list->length, 0, new_size - list->current_size);
```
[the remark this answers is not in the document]
JonathanHenson: doh! good catch

source/array_list.c
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    return aws_raise_error(AWS_ERROR_LIST_EMPTY);
+}
+
+int aws_array_list_pop_front(struct aws_array_list *list) {
+    if (list->length > 0) {
+        size_t last_bytes = list->item_size * (list->length - 1);
+        memmove(list->data, (void *)((uint8_t *)list->data + list->item_size), last_bytes);
+        memset((void *)((uint8_t *)list->data + last_bytes), 0, list->item_size);
```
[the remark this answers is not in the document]
JonathanHenson: I like debuggable code though. Having a set of 0s in unused space makes tracking down the bugs easier, but it's also a performance penalty, maybe I'll ifdef this off with the debug flag.

source/array_list.c
```diff
@@ -0,0 +1,247 @@
...
+
+    return aws_raise_error(AWS_ERROR_LIST_EMPTY);
+}
+
+void aws_array_list_clear(struct aws_array_list *list) {
+    if (list->length > 0) {
+        memset(list->data, 0, list->current_size);
+        list->length = 0;
+    }
+}
```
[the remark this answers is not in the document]
JonathanHenson: That should be in clean_up or shrink_to_fit, not in clear.

include/aws/common/array_list.h
```diff
@@ -0,0 +1,126 @@
...
+    * Initializes an array list with an array of void *. In this mode, the array size will grow
+    * by a factor of 2 upon insertion if space is not available. initial_item_allocation is the number of elements
+    * you want space allocated for. item_size is the size of each element in bytes. Mixing items types is not supported
+    * by this API.
+    */
+    AWS_COMMON_API int aws_array_list_init_dynamic(struct aws_array_list *list,
+        struct aws_allocator *alloc, size_t initial_item_allocation, size_t item_size);
+
+    /**
+    * Initializes an array list with a preallocated array of void *. array_size is the total allocation size of raw_array (not the number of elements),
```
bdonlan: We could eliminate the error condition (and division) by using an item count instead of an array size here. This also makes the parameters analogous to the dynamic init case.

include/aws/common/array_list.h
```diff
@@ -0,0 +1,126 @@
...
+     */
+    AWS_COMMON_API void aws_array_list_clear(struct aws_array_list *list);
+
+    /**
+     * If in dynamic mode, shrinks the allocated array size to the minimum amount necessary to store its elements.
+     */
+    AWS_COMMON_API int aws_array_list_shrink_to_fit(struct aws_array_list *list);
+
+    /**
+     * Copies the elements from from to to. If to is in static mode, it must at least be the same length as from.
```
bdonlan: We should be clear in documentation (and function naming) as to whether the `to` array is _replaced_ or _appended to_. My instinct is to assume it replaces the `to` array.

include/aws/common/array_list.h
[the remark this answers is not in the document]
bdonlan: Would it be better to get a pointer to the element to allow in-place manipulation?

source/array_list.c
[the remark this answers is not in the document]
bdonlan: It can be more helpful to use a nonzero sentinel value (e.g. `0xDD`) or even a multi-byte pattern (`0xDEADBEEF`) to mark uninitialized memory - the reason being, often _initialized_ structures are all-zero.

source/array_list.c
```diff
@@ -0,0 +1,247 @@
...
+        return AWS_OP_SUCCESS;
+    }
+    return aws_raise_error(AWS_ERROR_INVALID_INDEX);
+}
+
+int aws_array_list_set_at(struct aws_array_list *list, const void *val, size_t index) {
+    if (list->current_size > index * list->item_size) {
+        memcpy((void *)((uint8_t *)list->data + (list->item_size * index)), val, list->item_size);
+
+        /* this isn't perfect but its the best I can come up with for detecting length changes*/
```
bdonlan: I'm a bit confused about this. Does `set_at` implicitly extend the list? If so we need to also be prepared to resize the backing array. And this behavior needs to be documented in the header.

tests/CMakeLists.txt
```diff
@@ -25,3 +25,16 @@ add_test(mutex_is_actually_mutex_test ${TEST_BINARY_NAME} mutex_is_actually_mute
...
+add_test(array_list_iteration_test ${TEST_BINARY_NAME} array_list_iteration_test)
+add_test(array_list_preallocated_iteration_test ${TEST_BINARY_NAME} array_list_preallocated_iteration_test)
+add_test(array_list_preallocated_push_test ${TEST_BINARY_NAME} array_list_preallocated_push_test)
+add_test(array_list_shrink_to_fit_test ${TEST_BINARY_NAME} array_list_shrink_to_fit_test)
+add_test(array_list_shrink_to_fit_static_test ${TEST_BINARY_NAME} array_list_shrink_to_fit_static_test)
+add_test(array_list_clear_test ${TEST_BINARY_NAME} array_list_clear_test)
+add_test(array_list_copy_test ${TEST_BINARY_NAME} array_list_copy_test)
+add_test(array_list_not_enough_space_test ${TEST_BINARY_NAME} array_list_not_enough_space_test)
+add_test(array_list_not_enough_space_test_failure ${TEST_BINARY_NAME} array_list_not_enough_space_test_failure)
+
```
bdonlan: Unfortunately, github refused to load the diff for array_list_test.c, so I can't leave comments there directly, but some observations:

1. A lot of the message strings seem really redundant. Can we just add a simple `ASSERT_SUCCESS(foo())` macro, which will auto-format `test_foo.c:321: foo() failed unexpectedly due to error code 1234 (Error String)`? Same for ASSERT_*_EQUALS, most of them say the same thing as their parameters, but with more words. Having a specific string can be occasionally useful - but the majority of cases, the arguments are self-descriptive.
2. Are we leaking memory if the tests fail?

include/aws/common/array_list.h
JonathanHenson: I like that, it's less cumbersome for the API anyways.

tests/CMakeLists.txt
JonathanHenson: 1. Yes I've noticed that, I'd be game for having ASSERTS that don't take a format string.
2. Sort of and I think that's fine compared to having to worry about tests with cleanup blocks on failure. If the tests fail, I don't really care if the memchecks fail. Where I want them is when the tests pass logically, I want the build to fail if there's a leak.

However, the  before/after hooks allow for a more C++ fixtures approach, if that's what you want, and there the cleanup could be done regardless of pass/fail status.

include/aws/common/array_list.h
JonathanHenson: it should be replacing to, I'll add some docs.

source/array_list.c
JonathanHenson: My thought is that set_at should function within the existing bounds of the array (like in place replacement), and not do the resize. The semantics of having the list grow makes more sense in the push_back style interface.

include/aws/common/array_list.h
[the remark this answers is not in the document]
JonathanHenson: I'm hoping to have consistent value semantics for this, but I do see the value of allowing in-place manipulation. I'm not sure how to add that to the API (without adding another function) without making the API more complicated though. 

Maybe additional functions WOULD be best?

source/array_list.c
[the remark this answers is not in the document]
JonathanHenson: Good thought, I'll change, and still have it sitting behind a DEBUG flag.

include/aws/common/array_list.h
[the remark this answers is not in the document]
bdonlan: You can just add one function:

    AWS_COMMON_API int aws_array_list_elem_ptr(struct aws_array_list *list, void **pElem);

source/array_list.c
bdonlan: This is going to be difficult to use reliably, because whether set_at works `at all` now depends on a property of the list that 1) you have no direct control over and 2) is not likely to be examined in most use cases. It can result in code that works fine in initial testing, but then in production suddenly breaks when the list sizes align differently. It's better to have a consistent rule that method success is never tied to capacity when you have a dynamic list.

include/aws/common/array_list.h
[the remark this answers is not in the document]
JonathanHenson: wouldn't we need a function like this for back, front, and get_at?

source/array_list.c
JonathanHenson: I'll play around with it and see what happens.

include/aws/common/array_list.h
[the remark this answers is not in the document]
bdonlan: front is covered by passing zero. Back you could pass in length, or we could define `-1` to be back (and in general, `-N` meaning N from the end). I wouldn't consider those essential though.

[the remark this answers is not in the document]
JonathanHenson: I'm not sure I understand, where in the function signature you showed, would I pass 0 or -1 etc... 

How is that simpler than having:

/**
     * Copies the address of the element at the front of the list if it exists to *val. If list is empty, AWS_CORE_LIST_ERROR_EMPTY will be raised
     */
    int aws_array_list_front_ptr(const struct aws_array_list *list, void **val);

/**
     * Copies the address of the element at the end of the list to *val if it exists. If list is empty, AWS_CORE_LIST_ERROR_EMPTY will be raised.
     */
    int aws_array_list_back_ptr(const struct aws_array_list *list, void **val);

/**
     * Copies the memory address of the element at index to *val. If element does not exist, AWS_ERROR_INVALID_INDEX will be raised.
     */
    int aws_array_list_get_at_ptr(const struct aws_array_list *list, void **val, size_t index);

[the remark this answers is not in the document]
bdonlan: Pretend there was a `ssize_t index` on it :)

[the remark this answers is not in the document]
JonathanHenson: ok

PR #4, opened 2018-01-23
CMake improvements (see below)

marcomagdy:
*Description of changes:*
- separate the test cmake scripts from source
- treat warnings as errors when compiling with msvc
- baseline on cmake version 3.5 instead of 3.0
- avoid declaring vars that cmake already has an equivalent counterpart for

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,8 +14,6 @@
 cmake_minimum_required(VERSION 3.0)
 project(aws-c-common)
 
-set(PROJECT_NAME aws-c-common)
-
 file(GLOB AWS_COMMON_HEADERS
         "include/aws/common/*.h"
         )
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -11,11 +11,9 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.5)
```
JonathanHenson: We should think critically before doing this and make sure it's necessary, I'd like to baseline on the lowest possible revision.

marcomagdy:
rebased and updated the PR. Ready to merge if approved.

marcomagdy:
Actually, you can go ahead and merge it right away from here.

PR #5, opened 2018-01-26
Add intrusive linked_list implementation

marcomagdy:


source/linked_list.c
```diff
@@ -0,0 +1,124 @@
...
+
+    list->length++;
+}
+
+struct aws_linked_list_node * aws_linked_list_front(struct aws_linked_list *list) {
+    return list->head;
+}
+
+struct aws_linked_list_node* aws_linked_list_pop_front(struct aws_linked_list *list) {
+    struct aws_linked_list_node *out = list->head;;
```
JonathanHenson: ;;

tests/linked_list_test.c
```diff
@@ -0,0 +1,239 @@
...
+    item = (struct int_value *)aws_linked_list_front(&list);
+    ASSERT_TRUE(item != 0, "List front() should have been successful.");
+    aws_linked_list_pop_front(&list);
+    ASSERT_INT_EQUALS(fourth.value, item->value, "Item should have been the fourth item.");
+    ASSERT_INT_EQUALS(list_size - 4, aws_linked_list_length(&list), "List size should be %d.", (int)list_size - 4);
+
+    aws_linked_list_clean_up(&list);
+    RETURN_SUCCESS("%s() pass", "test_linked_list_order_push_back_pop_front");
+
+failure:
```
JonathanHenson: This is still using the old style for unit tests, not the new AWS_TEST_CASE. in particular, you get rid of RETURN_SUCCESS and the failure goto.

include/aws/common/linked_list.h
```diff
@@ -0,0 +1,95 @@
...
+extern "C" {
+#endif
+
+    /**
+    * Initializes a doubly linked list.
+    */
+    AWS_COMMON_API void aws_linked_list_init(struct aws_linked_list *list);
+
+    /**
+    * Deallocates all nodes and resets list for either reuse or deletion.
```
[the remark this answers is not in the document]
marcomagdy: right. Which makes `aws_linked_list_clear` and `aws_linked_list_cleanup` redundant. I will remove the *_clear version.

tests/linked_list_test.c
```diff
@@ -67,11 +67,7 @@ static int test_linked_list_order_push_back_pop_front() {
     ASSERT_INT_EQUALS(list_size - 4, aws_linked_list_length(&list), "List size should be %d.", (int)list_size - 4);
 
     aws_linked_list_clean_up(&list);
-    RETURN_SUCCESS("%s() pass", "test_linked_list_order_push_back_pop_front");
-
-failure:
-    aws_linked_list_clean_up(&list);
-    return FAILURE;
+    return SUCCESS;
```
JonathanHenson: these should either just be return 0 or return AWS_OP_SUCCESS. The success macro is old, and I probably just need to delete it.

include/aws/common/linked_list.h
```diff
@@ -0,0 +1,91 @@
...
+*
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/common.h>
+#include <stdint.h>
+
```
bdonlan: An `aws_container_of` macro would be helpful.

include/aws/common/linked_list.h
```diff
@@ -0,0 +1,91 @@
...
+struct aws_linked_list_node {
+    struct aws_linked_list_node *next;
+    struct aws_linked_list_node *prev;
+};
+
+struct aws_linked_list {
```
bdonlan: Consider having the 'list' itself be a node, and having the list be circular. This is how it's done in the linux kernel and leads to a dramatic simplification in the linked list code - a lot of the branches can be removed. It also means we can be a lot more flexible in how we manage spliced-out sections of the list (something that isn't really handled with this API), as any single node can define a list on its own.

include/aws/common/linked_list.h
```diff
@@ -0,0 +1,91 @@
...
+};
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    /**
+    * Initializes a doubly linked list.
+    */
+    AWS_COMMON_API void aws_linked_list_init(struct aws_linked_list *list);
```
bdonlan: These methods are good candidates for `static inline` methods.

include/aws/common/linked_list.h
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+    * Pushes an element to the front of the list.
+    */
+    AWS_COMMON_API void aws_linked_list_push_front(struct aws_linked_list *list, struct aws_linked_list_node *node);
+
+    /**
+    * Copies a memory address of the first element into item.
+    */
+    AWS_COMMON_API struct aws_linked_list_node* aws_linked_list_front(struct aws_linked_list *list);
```
bdonlan: `head/tail` are more commonly how the head and tail of the list are referred to. Regardless, be consistent with naming of struct members vs naming of methods here.

source/linked_list.c
```diff
@@ -0,0 +1,120 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/linked_list.h>
+#include <string.h>
+
+void aws_linked_list_init(struct aws_linked_list *list) {
+    list->head = 0;
```
bdonlan: NULL

source/linked_list.c
```diff
@@ -0,0 +1,120 @@
...
+    list->length = 0;
+    list->tail = 0;
+}
+
+void aws_linked_list_clean_up(struct aws_linked_list *list) {
```
bdonlan: Because linked lists are intrusive and can't be truly deallocated, this method isn't really meaningful.

tests/linked_list_test.c
```diff
@@ -0,0 +1,223 @@
...
+
+#include <aws/common/linked_list.h>
+#include <aws_test_harness.h>
+
+struct int_value {
+    struct aws_linked_list_node node;
+    int value;
+};
+
+static int test_linked_list_order_push_back_pop_front() {
```
bdonlan: Where's the test for `_remove`?

tests/linked_list_test.c
```diff
@@ -0,0 +1,223 @@
...
+    aws_linked_list_init(&list);
+
+    struct int_value first = (struct int_value){.value = 1};
+    struct int_value second = (struct int_value){.value = 2};
+    struct int_value third = (struct int_value){.value = 3};
+    struct int_value fourth = (struct int_value){.value = 4};
+
+    ASSERT_INT_EQUALS(0, aws_linked_list_length(&list), "List size should be 0.");
+
+    aws_linked_list_push_back(&list, (struct aws_linked_list_node *)&first);
```
bdonlan: `aws_linked_list` is feeling a bit verbose to me. Can/should we abbreviate our list names? `aws_alist` (array), `aws_dlist` (doubly linked intrusive) for example.

source/linked_list.c
JonathanHenson: maybe we should just keep clear() and remove clean_up()?

bdonlan: There is no clear() currently.

JonathanHenson: There was and we removed it because of redundancy

marcomagdy:
I've converted the linked_list to a node based circular one. 
I haven't implemented aws_container_of yet, I really dislike the naming, very unintuitive to me. I'll come up with something else.
As far as renaming, I agree they're verbose. I'd go for aws_list, aws_slist (if we ever do a singly linked list), aws_array_list. dlist, and alist are too cryptic IMO

include/aws/common/linked_list.h
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+    * Resets the list for either reuse or deletion.
+    */
+    AWS_COMMON_API void aws_linked_list_clear(struct aws_linked_list_node *head);
+
+    /**
+    * Adds a node before the specified head.
+    */
+    AWS_COMMON_API void aws_linked_list_prepend(struct aws_linked_list_node *head, struct aws_linked_list_node *node);
```
JonathanHenson: head seems like a misleading name to me.

tests/linked_list_test.c
```diff
@@ -0,0 +1,100 @@
...
+
+    head = &list;
+    aws_linked_list_init(head);
+
+    struct int_value first = (struct int_value){.value = 1};
+    struct int_value second = (struct int_value){.value = 2};
+    struct int_value third = (struct int_value){.value = 3};
+    struct int_value fourth = (struct int_value){.value = 4};
+
+    aws_linked_list_prepend(head, (struct aws_linked_list_node *)&first);
```
JonathanHenson: This API confuses me, I'm willing to bet in confuses users. I suggest a rename to insert_before and insert_after, and some macros at least for push/pop semantics.

include/aws/common/linked_list.h
```diff
@@ -0,0 +1,86 @@
...
+#define aws_linked_list_push_back(head, new_node) \
+    aws_linked_list_insert_before((head), (new_node))
+
+#define aws_linked_list_push_front(head, new_node) \
+    aws_linked_list_insert_after((head), (new_node))
+
+#define aws_linked_list_pop_back(head) \
+    aws_linked_list_remove((head)->prev)
+
+#define aws_linked_list_pop_front(head) \
```
JonathanHenson: what about getting the front element and the back elements?

marcomagdy: They don't make sense in a circular list. Everything is relative to a specified node; so if you specify a node and ask for head, it should return itself, and tail returns prev.

include/aws/common/linked_list.h
```diff
@@ -0,0 +1,86 @@
...
+};
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    /**
+    * Initializes a doubly linked list.
+    */
+    AWS_COMMON_API void aws_linked_list_init(struct aws_linked_list_node *head);
```
bdonlan: I still recommend static inlining these. Function call overhead may be significant when doing heavy list manipulation.

include/aws/common/linked_list.h
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+    * Resets the list for either reuse or deletion.
```
bdonlan: Resetting an intrusive list is not required for deletion.

It's better to just have init initialize the element to represent a list of a single element. Note also that `clear` and `init` have the same behavior, but the docs aren't clear at all about this, nor are they clear about the difference between the two.

source/linked_list.c
```diff
@@ -0,0 +1,55 @@
...
+        struct aws_linked_list_node *prev,
+        struct aws_linked_list_node *next) {
+
+    next->prev = new;
+    new->next = next;
+    new->prev = prev;
+    prev->next = new;
+}
+
+void aws_linked_list_insert_before(struct aws_linked_list_node *head, struct aws_linked_list_node *new) {
```
bdonlan: You could also implement this as:

    aws_linked_list_insert_after(head->prev, new);

This eliminates the need for a helper function, which can be useful if you're going to static inline these.

source/linked_list.c
```diff
@@ -0,0 +1,55 @@
...
+}
+
+void aws_linked_list_insert_after(struct aws_linked_list_node *head, struct aws_linked_list_node *new) {
+    aws_linked_list_add(new, head, head->next);
+}
+
+struct aws_linked_list_node *aws_linked_list_remove(struct aws_linked_list_node *head) {
+    head->prev->next = head->next;
+    head->next->prev = head->prev;
+    return head->next;
```
bdonlan: I'd suggest resetting head as well.

include/aws/common/linked_list.h
marcomagdy: It seems like Clang doesn't like static inline functions in header very much.

https://patchwork.kernel.org/patch/9746913/

include/aws/common/linked_list.h
```diff
@@ -54,31 +53,70 @@ extern "C" {
     /**
      * Initializes a circular doubly-linked list.
      */
-    AWS_COMMON_API void aws_linked_list_init(struct aws_linked_list_node *head);
+    static inline AWS_COMMON_API void aws_linked_list_init(struct aws_linked_list_node *head);
 
     /**
      * Removes all the elements from the list.
```
bdonlan: How useful of a function is this? How do you find all the elements that were cleared so you can free them later?

marcomagdy: It would be useful only in the static list case, or if the contents of the list are ints/pointers.

bdonlan: The list has no contents; it is _contained within_ another list. So you still need to have something find and free list nodes. I don't think this is particularly useful given that use case.

marcomagdy: no big deal will remove it.

include/aws/common/linked_list.h
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+#include <aws/common/common.h>
+
+#define aws_linked_list_of(type, name) struct name {                                                                   \
+    struct aws_linked_list_node list;                                                                                  \
+    type value;                                                                                                        \
+}                                                                                                                      \
+
+#define aws_container_of(ptr, type, member)  ((type *)ptr)->member
```
bdonlan: Actually hold on a sec, this macro is still broken. What if the list isn't the first member of whatever it's in?

bdonlan: The normal definition is something like:

```
#define container_of(ptr, container_type, container_member) \
  ((container_type *)((uint8_t *)ptr - (ptrdiff_t)&((container_type *)NULL)->container_member)
```

PR #6, opened 2018-01-31
Buffer and integer encoding with tests.

JonathanHenson:


source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+static const uint8_t bytes_4 = 4 * 8;
+static const uint8_t bytes_3 = 3 * 8;
+static const uint8_t bytes_2 = 2 * 8;
+static const uint8_t bytes_1 = 1 * 8;
+static const uint8_t mask = 0xff;
+
+
+int aws_hex_encode(const uint8_t *to_encode, size_t to_encode_len, char *output, size_t *output_size) {
+
+    size_t encoded_len = (to_encode_len << 1) + 1;
```
marcomagdy: add overflow check?
```C
if(encoded_len <= to_encode_len) raise_error(...);
```

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+        output[str_index++] = base64_encoding_table[(block >> 18) & 0x3F];
+        output[str_index++] = base64_encoding_table[(block >> 12) & 0x3F];
+        output[str_index++] = base64_encoding_table[(block >> 6) & 0x3F];
+        output[str_index++] = base64_encoding_table[block & 0x3F];
+    }
+
+    if(remainder_count > 0)
+    {
+        output[block_count * 4 - 1] = '=';
+        if(remainder_count == 1)
```
marcomagdy: shouldn't this be `if (remainder_count == 2)`?

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+
+    if (*output_size < encoded_len) {
+        return aws_raise_error(AWS_ERROR_INVALID_BUFFER_SIZE);
+    }
+
+    /* empty string should have output of "" */
+    if(encoded_len == 1 && *to_encode == '\0') {
+        *output_size = 1;
+        *output = '\0';
+    }
```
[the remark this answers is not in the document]
JonathanHenson: It doesn't and it's more complicated to make it do so, due to the difference between an empty buffer, and a const char * string.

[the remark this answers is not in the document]
JonathanHenson: I can add a return statement here.

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+
+    *output_size = 0;
+
+    for (size_t i = 0; i < to_encode_len; ++i) {
+
+        output[(*output_size)++] = hex_chars[to_encode[i] >> 4 & 0x0f];
+        output[(*output_size)++] = hex_chars[to_encode[i] & 0x0f];
```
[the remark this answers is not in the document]
JonathanHenson: changed.

source/encoding.c
JonathanHenson: no

source/encoding.c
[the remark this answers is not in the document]
marcomagdy: Discussed separately. We shouldn't be special casing empty strings. Input should be treated as byte buffer.

include/aws/common/encoding.h
```diff
@@ -0,0 +1,96 @@
...
+AWS_COMMON_API int aws_hex_encode(const uint8_t *to_encode, size_t to_encode_len, char *output, size_t *output_size);
+
+/*
+ * Base 16 (hex) decodes the contents of to_decode and stores the result in output.
+ * If output is NULL, output_size will be set to what the output_size should be.
+ */
+AWS_COMMON_API int aws_hex_decode(const char *to_decode, size_t to_decode_len, uint8_t *output, size_t *output_size);
+
+/*
+ * Base 64 encodes the contents of to_encode and stores the result in output.
```
bdonlan: Should we support urlsafe b64?

include/aws/common/encoding.h
```diff
@@ -0,0 +1,96 @@
...
+/*
+ * Base 64 decodes the contents of to_decode and stores the result in output.
+ * If output is NULL, output_size will be set to what the output_size should be.
+ */
+AWS_COMMON_API int aws_base64_decode(const char *to_decode, size_t to_decode_len, uint8_t *output, size_t *output_size);
+
+/*Add a 64 bit unsigned integer to the buffer, ensuring network - byte order
+ * return the new position in the buffer for the next operation.
+ */
+AWS_COMMON_API uint8_t *aws_add_uint64_to_buffer(uint8_t *buffer, uint64_t value);
```
bdonlan: I would recommend a simpler API:

```
/*
 * Write a 64-bit value to *buf in network byte order. buf need not be aligned.
 */
void aws_write_u64(void *buf, uint64_t value);

/*
 * Reads a 64-bit value from *buf in network byte order. buf need not be aligned.
 */
uint64_t aws_read_u64(const void *buf);
```

There's not a lot of value add for returning the adjusted pointer unless you're also doing bounds checks - in which case you might be able to add some spectre mitigations here. I have some prototype code along these lines I can share.

This is also another place where I would strongly recommend writing these as static inline functions in a header file. An unaligned 64-bit big-endian read is two instructions on x86_64, whereas a function call implies a lot of overhead (register spills, etc).

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+static const uint8_t bytes_6 = 6 * 8;
+static const uint8_t bytes_5 = 5 * 8;
+static const uint8_t bytes_4 = 4 * 8;
+static const uint8_t bytes_3 = 3 * 8;
+static const uint8_t bytes_2 = 2 * 8;
+static const uint8_t bytes_1 = 1 * 8;
+static const uint8_t mask = 0xff;
+
+
+int aws_hex_encode(const uint8_t *to_encode, size_t to_encode_len, char *output, size_t *output_size) {
```
bdonlan: Declaring these pointers as `type *restrict varname` will help the compiler in these kinds of buffer-heavy operations, by declaring that the input and output cannot alias each other (nor, for that matter, can they alias the encoding table). If the compiler is sufficiently clever it may also enable vectorization.

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+        output[written++] = hex_chars[to_encode[i] & 0x0f];
+    }
+
+    output[written++] = '\0';
+    *output_size = written;
+
+    return AWS_OP_SUCCESS;
+}
+
+static uint8_t char_to_int(char character) {
```
bdonlan: make it clear with this function name that this is specifically for hex decoding

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+    if (character >= 'a' && character <= 'f') {
+        return (uint8_t)(10 + (character - 'a'));
+    }
+
+    if (character >= 'A' && character <= 'F') {
+        return (uint8_t)(10 + (character - 'A'));
+    }
+
+    return (uint8_t)(character - '0');
```
bdonlan: What do we do with invalid hex characters?

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+}
+
+static size_t calculate_base64_decoded_length(const char *input, size_t len) {
+    if (len == 0) {
+        return 0;
+    }
+
+    size_t padding = 0;
+
+    if (len >= 2 && input[len - 1] == '=' && input[len - 2] == '=') { /*last two chars are = */
```
bdonlan: What happens if the padding is invalid?

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+        padding = 2;
+    }
+    else if (input[len - 1] == '=') { /*last char is = */
+        padding = 1;
+    }
+
+    return (len * 3 / 4 - padding);
```
bdonlan: What behavior do we expect for invalid input? Specifically input that has length not divisible by 4.

Note also that the `* 3` here may overflow.

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+}
+
+static size_t calculate_base64_encoded_length(size_t buffer_len) {
+    return 4 * ((buffer_len + 2) / 3) + 1; /* plus one for the NULL terminator */
```
bdonlan: What happens if `buffer_len + 2` overflows?

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+}
+
+int aws_base64_encode(const uint8_t *to_encode, size_t to_encode_len, char *output, size_t *output_size) {
+
+    size_t encoded_length = calculate_base64_encoded_length(to_encode_len);
+    if(!output) {
+        *output_size = encoded_length;
+        return AWS_OP_SUCCESS;
```
bdonlan: We should probably have clearer semantics around our return value. For example, we could declare that `*output_size` is always set, and `AWS_OP_SUCCESS` is returned only if the input is fully converted to the output.

Or, probably better would be to have a separate function to compute the length required for the output buffer based solely on input length, and just fail if the output buffer is too small here.

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+            output[block_count * 4 - 2] = '=';
+        }
+    }
+
+    return AWS_OP_SUCCESS;
+}
+
+static void init_decoding_table() {
+    if(!base64_decoding_table_init) {
+        for(size_t i = 0; i < (sizeof(base64_encoding_table) / sizeof(char)); ++i)
```
bdonlan: Let's make this a compile-time-constant table. It's a minor pain but it'll make the actual code at runtime simpler and safer.

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+        {
+            size_t index = (size_t)base64_encoding_table[i];
+            base64_decoding_table[index] = (uint8_t)i;
+        }
+
+        base64_decoding_table[(size_t)'='] = base64_sentinal_value;
```
bdonlan: What about all the illegal characters?

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+    for(size_t i = 0; i < block_count; ++i)
+    {
+        uint32_t value1 = base64_decoding_table[(size_t)to_decode[string_index++]];
+        uint32_t value2 = base64_decoding_table[(size_t)to_decode[string_index++]];
+        uint32_t value3 = base64_decoding_table[(size_t)to_decode[string_index++]];
+        uint32_t value4 = base64_decoding_table[(size_t)to_decode[string_index++]];
+
+        size_t buffer_index = i * 3;
+        output[buffer_index++] = (uint8_t)((value1 << 2) | ((value2 >> 4) & 0x03));
+        if(value3 != base64_sentinal_value)
```
bdonlan: The main loop shouldn't be making all these tests; we should make the loop go for `block_count - 1` iterations and do all these tests at the final iteration only.

source/encoding.c
```diff
@@ -0,0 +1,316 @@
...
+                output[buffer_index] = (uint8_t)((value3 & 0x03) << 6 | value4);
+            }
+        }
+    }
+
+    return AWS_OP_SUCCESS;
+}
+
+uint8_t *aws_add_uint64_to_buffer(uint8_t *buffer, uint64_t value) {
+    buffer[0] = (uint8_t)(value >> bytes_7) & mask;
```
bdonlan: This is the sort of thing that needs optimization.

Here's a portable optimized version for this variant, but we should optimize the rest as well:

```
uint32_t lo = htonl(value);
uint32_t hi = htonl(value >> 32);
memcpy(&buffer[0], &hi, 4);
memcpy(&buffer[0], &lo, 4);
```

It's possible to do slightly better than this on x86_64 by using the 64-bit `BSWAP` instruction.

tests/encoding_test.c
```diff
@@ -0,0 +1,368 @@
...
+    ASSERT_SUCCESS(aws_hex_encode((uint8_t *)test_str, test_str_size - 1, NULL, &output_size), "encode call should have succeeded");
+    ASSERT_INT_EQUALS(expected_size, output_size, "Output size on string should be %d", expected_size);
+
+    output_str = (char *)aws_mem_acquire(alloc, output_size);
+    memset(output_str, 0, output_size);
+
+    ASSERT_SUCCESS(aws_hex_encode((uint8_t *)test_str, test_str_size - 1, output_str, &output_size), "encode call should have succeeded");
+    ASSERT_INT_EQUALS(expected_size, output_size, "Output size on an empty string should be %d", expected_size);
+
+    ASSERT_BIN_ARRAYS_EQUALS(expected, expected_size, output_str, output_size, "Encode output should have been %s.", expected);
```
bdonlan: I'd recommend having a test which confirms that we don't overwrite or access memory before or after the buffer.

The portable way to do this is to have sentinel values before and after the range that should be written.

A more sophisticated variant (which catches overreads) involves aligning the buffer to the edge of a page with all access disabled. However there's no portable way to do this.

tests/encoding_test.c
```diff
@@ -0,0 +1,368 @@
...
+    char *output_str = NULL;
+
+    ASSERT_SUCCESS(aws_hex_encode((uint8_t *)test_str, test_str_size - 1, NULL, &output_size), "encode call should have succeeded");
+    ASSERT_INT_EQUALS(expected_size, output_size, "Output size on string should be %d", expected_size);
+
+    output_str = (char *)aws_mem_acquire(alloc, output_size);
+    memset(output_str, 0, output_size);
+
+    ASSERT_SUCCESS(aws_hex_encode((uint8_t *)test_str, test_str_size - 1, output_str, &output_size), "encode call should have succeeded");
+    ASSERT_INT_EQUALS(expected_size, output_size, "Output size on an empty string should be %d", expected_size);
```
bdonlan: ... is `test_str` always going to be empty?

tests/encoding_test.c
```diff
@@ -0,0 +1,368 @@
...
+    char *output_str = NULL;
+
+    ASSERT_SUCCESS(aws_base64_encode((uint8_t *)test_str, test_str_size - 1, NULL, &output_size), "encode call should have succeeded");
+    ASSERT_INT_EQUALS(expected_size, output_size, "Output size on string should be %d", expected_size);
+
+    output_str = (char *)aws_mem_acquire(alloc, output_size);
+    memset(output_str, 0, output_size);
+
+    ASSERT_SUCCESS(aws_base64_encode((uint8_t *)test_str, test_str_size - 1, output_str, &output_size), "encode call should have succeeded");
+    ASSERT_INT_EQUALS(expected_size, output_size, "Output size on an empty string should be %d", expected_size);
```
bdonlan: same comment re empty string

tests/encoding_test.c
```diff
@@ -0,0 +1,368 @@
...
+    return 0;
+}
+
+AWS_TEST_CASE(base64_encoding_invalid_buffer_size_test, base64_encoding_invalid_buffer_size_test_fn)
+
+static int uint64_buffer_test_fn(struct aws_allocator *alloc, void *ctx) {
+
+    uint64_t test_value = 0x1020304050607080;
+    uint8_t buffer[8] = { 0 };
+    uint8_t *new_ptr = aws_add_uint64_to_buffer(buffer, test_value);
```
bdonlan: Please test reads and writes to both aligned and non-aligned buffers.

source/encoding.c
marcomagdy: But `htonl` is not portable.
You can avoid the shifting and the bitwise Anding by doing something like this:
```C
uint8_t *write_uint64(uint8_t *buffer, uint64_t value) {
    union {
        uint8_t b0;
        uint8_t b1;
        //....
        uint8_t b7;
        uint64_t all;
    };
    all = value;
    buffer[0] = b7;
    buffer[1] = b6;
    // ...
    buffer[7] = b0;
    return buffer + sizeof(value);
}
```
This is basically 9 assignments.

include/aws/common/encoding.h
JonathanHenson: If we do static inline, we can never expose this interface in the form of bindings for other languages.

I think I agree on the first statement.

source/encoding.c
JonathanHenson: apparently htonl is portable

https://msdn.microsoft.com/en-us/library/windows/desktop/ms740069(v=vs.85).aspx

and it goes without saying it's in posix as well.

tests/encoding_test.c
JonathanHenson: no, that's left over from when this function was part of a particular test and I copy and pasted it during a refactor.

include/aws/common/encoding.h
JonathanHenson: what in particular would you be wanting in that case? base64 that has been url encoded? In that case no.

include/aws/common/encoding.h
bdonlan: That's a good point. We can however arrange to `static inline` but also emit an actual implementation, even without the C99 `extern inline`. For example, we could have, instead of `static inline`:

```
#ifndef AWS_INLINE
#define AWS_INLINE static inline
#endif
```

Then, in an `inlines.c`:

```
#define AWS_INLINE
#include <aws/common/encoding.h>
#include <aws/common/list.h>
// ...
#endif
```

C code will use the `static inline` definition; however, an implementation of the function will still be in the library for use by FFIs.

source/encoding.c
bdonlan: Worst case we can probe for htonl in CMake; it's a bit of work but reading/writing big-endian values is worth it I think.

include/aws/common/encoding.h
JonathanHenson: on windows it needs to be dll_export etc... to get exported. I think this is the right direction, I just need to think on what the macro trickery would be. Largely, I want to avoid different assemblies based on the consumer.

source/encoding.c
JonathanHenson: As far as I can tell, it's supported everywhere that matters.

JonathanHenson: I'll note the header files are different but that's easy enough to deal with.

bdonlan: Embedded would be my only real suspect for missing htonl and friends.

source/encoding.c
JonathanHenson: That's going to make it more expensive (since we'll need a branch inside the loop), but I've updated it.

source/encoding.c
bdonlan: Well, if length is not a multiple of 4 bytes, or if you have more than two `=`s at the end, do we report an error?

JonathanHenson: length is a fairly benign check, to verify the invalid characters, is more expensive and would be caught in decode anyways. My vote is to do the length check, but leave the illegal sequence update till the decode run.

JonathanHenson: I have tests verifying this behavior now.

include/aws/common/encoding.h
```diff
@@ -0,0 +1,175 @@
...
+ * Base 16 (hex) decodes the contents of to_decode and stores the result in output.
+ * If output is NULL, output_size will be set to what the output_size should be.
+ */
+AWS_COMMON_API int aws_hex_decode(const char *restrict to_decode, size_t to_decode_len, uint8_t *restrict output, size_t output_size);
+
+/*
+ * Computes the length necessary to store the output of aws_base64_encode call. returns -1 on failure, and 0 on success.
+ * encoded_length will be set on success.
+ */
+AWS_COMMON_API int aws_compute_base64_encoded_len(size_t to_encode_len, size_t *encoded_len);
```
bdonlan: Would recommend using consistent prefixes - b64 related functions should all start with `aws_base64_`. This makes grepping easier and reduces cognitive burden.

source/encoding.c
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    /* it's a string add the null terminator. */
+    output[encoded_length - 1] = 0;
+
+    return AWS_OP_SUCCESS;
+}
+
+static inline int base64_get_decoded_value(char to_decode, uint8_t *value, int8_t allow_sentinal) {
+    if (isalnum(to_decode) || to_decode == '/' || to_decode == '+' || (allow_sentinal && to_decode == '=')) {
```
bdonlan: This would be faster if you used a flag value in the decoding table to mark illegal base64 characters rather than this collection of four conditions (including one internal table lookup)

JonathanHenson: Good idea will update

PR #7, opened 2018-02-05
priority_queue implementation

marcomagdy:


source/priority_queue.c
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    aws_array_list_get_at_ptr(&queue->container, &item, right);
+    if (right < len && queue->compare(item, swap_item)) {
+        swap_candidate = right;
+    }
+
+    if (swap_candidate != index) {
+        aws_array_list_swap(&queue->container, index, swap_candidate);
+        make_heap_from_top(queue, swap_candidate);
```
marcomagdy: thinking about replacing this with an iterative version. I realize that this is tail recursive and _should_ be optimized by the compiler. Thoughts?

bdonlan: Tail recursion optimizations are a bit spotty, avoid relying on it if you can avoid it.

source/array_list.c
```diff
@@ -260,3 +260,34 @@ int aws_array_list_set_at(struct aws_array_list *list, const void *val, size_t i
...
 
+void aws_array_list_swap(struct aws_array_list *list, size_t a, size_t b) {
+    assert(a < list->length);
+    assert(b < list->length);
+
+    void *item1, *item2;
+    aws_array_list_get_at_ptr(list, &item1, a);
+    aws_array_list_get_at_ptr(list, &item2, b);
+
+    /* copy bytes sizeof(size_t) at a time */
```
bdonlan: Why not more?

source/array_list.c
```diff
@@ -260,3 +260,34 @@ int aws_array_list_set_at(struct aws_array_list *list, const void *val, size_t i
...
+    size_t sb_pointer_size_count = list->item_size / sizeof(size_t);
+    uint8_t temp[sizeof(size_t)];
+    for(size_t i = 0; i < sb_pointer_size_count; i++) {
+        memcpy(temp, item1, sizeof(size_t));
```
bdonlan: As a performance note: The compiler has to assume that these items may overlap arbitrarily, which prevents it from doing optimizations like autovectorization or loop unrolling.

I'd suggest you 1) bail out if `a == b`, and then delegate to a helper `void aws_memswap(void * restrict a, void * restrict b, size_t len)`, which will tell the compiler that these two objects cannot overlap.

source/array_list.c
```diff
@@ -260,3 +260,34 @@ int aws_array_list_set_at(struct aws_array_list *list, const void *val, size_t i
...
+        memcpy(item1, item2, sizeof(size_t));
+        memcpy(item2, temp, sizeof(size_t));
+        item1 = (size_t*)item1 + 1;
+        item2 = (size_t*)item2 + 1;
+    }
+
+    /* copy the rest one byte at a time */
```
bdonlan: Why not memcpy the entire remainder into a buffer?

source/priority_queue.c
```diff
@@ -0,0 +1,135 @@
...
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/priority_queue.h>
+#include <string.h>
+
+static void make_heap_from_top(struct aws_priority_queue *queue, int index) {
+    int left = (index << 1) + 1;
```
bdonlan: Let's have helpers or macros for left/right/parent.

source/priority_queue.c
```diff
@@ -0,0 +1,135 @@
...
+        aws_array_list_swap(&queue->container, index, swap_candidate);
+        make_heap_from_top(queue, swap_candidate);
+    }
+}
+
+static void make_heap_from_bottom(struct aws_priority_queue *queue, int index) {
+    int parent = 0;
+
+    if (index & 0x01) {
+        parent = (index - 1) >> 1;
```
bdonlan: Equivalent to `index >> 1`

include/aws/common/priority_queue.h
```diff
@@ -0,0 +1,88 @@
...
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/common.h>
+#include <aws/common/array_list.h>
+
+struct aws_priority_queue {
+    int(*compare)(void *a, void *b);
```
bdonlan: Do we need to pass a context pointer? How would one implement a compare function that, say, calls back into java?

source/priority_queue.c
```diff
@@ -0,0 +1,135 @@
...
+    int right = left + 1;
+    int swap_candidate = index;
+    int len = aws_array_list_length(&queue->container);
+
+
+    void *item, *swap_item;
+    aws_array_list_get_at_ptr(&queue->container, &item, left);
+    aws_array_list_get_at_ptr(&queue->container, &swap_item, swap_candidate);
+
+    if (left < len && queue->compare(item, swap_item)) {
```
bdonlan: What are the semantics of compare? Are we swapping whenever the items are nonequal here?

source/priority_queue.c
```diff
@@ -0,0 +1,135 @@
...
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/priority_queue.h>
+#include <string.h>
+
+static void make_heap_from_top(struct aws_priority_queue *queue, int index) {
```
bdonlan: Would recommend defining preconditions for these functions. These aren't 'make heap' operations as they can't start from an arbitrary order of elements and end up with a heap.

source/priority_queue.c
bdonlan: Generally don't rely on tail recursion, as compilers don't necessarily apply it consistently. In particular, certain debug flags may disable tail recursion in favor of maintaining an accurate stack trace.

source/priority_queue.c
```diff
@@ -0,0 +1,135 @@
...
+}
+
+static void make_heap_from_bottom(struct aws_priority_queue *queue, int index) {
+    int parent = 0;
+
+    if (index & 0x01) {
+        parent = (index - 1) >> 1;
+    }
+    else {
+        parent = (index - 2) >> 1;
```
bdonlan: This can go negative if we have only a single element (the root).

source/priority_queue.c
```diff
@@ -0,0 +1,135 @@
...
+        return AWS_ERROR_PRIORITY_QUEUE_EMPTY;
+    }
+
+    int err = aws_priority_queue_top(queue, item);
+
+    if (err) {
+        return err;
+    }
+
+    err = aws_array_list_pop_front(&queue->container);
```
bdonlan: This will shift all elements down one, which is not what you want when doing a heap as it'll effectively destroy the entire heap tree structure. You need to swap in the element currently at the end, then fix the heap with a down-heap operation.

tests/priority_queue_test.c
```diff
@@ -0,0 +1,77 @@
...
+#include <aws_test_harness.h>
+
+static int int_compare(void *a, void *b) {
+    int *a_val = (int*)a;
+    int *b_val = (int*)b;
+
+    return *a_val < *b_val;
+}
+
+static int test_priority_queue_preserves_order(struct aws_allocator *alloc, void *ctx) {
```
bdonlan: There was a major bug in the implementation of pop, which suggests that the test case is not comprehensive enough. I would recommend adding a randomized test - say, maintain a set of 16 +- 4 elements, and randomly push/pop, verifying that the correct elements are returned by comparing against a very simple unsorted array that you scan to find the maximum.

source/priority_queue.c
marcomagdy: doh! good catch.

include/aws/common/priority_queue.h
JonathanHenson: It seems to me, that you'd just pass a function pointer, where the implementation knows to callback into java.

include/aws/common/priority_queue.h
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Initializes a priority queue struct for use. This mode will grow memory automatically (exponential model)
+     * Default size is the inital size of the queue
+     * item_size is the size of each element in bytes. Mixing items types is not supported by this API.
+     * compare() is the function that will be used to determine priority.
+     * If compare returns a positive integer, a is of higher priority than b.
+     */
+    AWS_COMMON_API int aws_priority_queue_dynamic_init(struct aws_priority_queue *queue,
+            struct aws_allocator *alloc, size_t default_size, size_t item_size, int(*compare)(void *a, void *b));
```
JonathanHenson: compare may be better as a typedef? I don't feel particularly strongly about it though.

source/array_list.c
```diff
@@ -260,3 +260,40 @@ int aws_array_list_set_at(struct aws_array_list *list, const void *val, size_t i
...
+    uint8_t temp[SLICE];
+    for(size_t i = 0; i < slice_count; i++) {
+        memcpy(temp, item1, SLICE);
+        memcpy(item1, item2, SLICE);
+        memcpy(item2, temp, SLICE);
+        item1 = (uint8_t*)item1 + SLICE;
+        item2 = (uint8_t*)item2 + SLICE;
+    }
+
+    size_t remainder = item_size % SLICE;
```
JonathanHenson: size_t remainder = item_size & 0x7F

is likely faster, and this is a hot path, so it would probably be better.

source/array_list.c
```diff
@@ -260,3 +260,40 @@ int aws_array_list_set_at(struct aws_array_list *list, const void *val, size_t i
...
+    memcpy(temp, item1, remainder);
+    memcpy(item1, item2, remainder);
+    memcpy(item2, &temp, remainder);
```
JonathanHenson: I don't think you need &temp here.

source/priority_queue.c
```diff
@@ -0,0 +1,139 @@
...
+
+    int err = aws_priority_queue_top(queue, item);
+
+    if (err) {
+        return err;
+    }
+
+    aws_array_list_swap(&queue->container, 0, aws_array_list_length(&queue->container) - 1);
+
+    err = aws_array_list_pop_back(&queue->container);
```
JonathanHenson: nit:

/* because in this case aws_raise_error has already been called */
if (aws_array_list_*(&queue->container)) {
    return AWS_OP_ERR;
}

tests/priority_queue_test.c
```diff
@@ -0,0 +1,114 @@
...
+    struct aws_priority_queue queue;
+
+    aws_priority_queue_dynamic_init(&queue, alloc, 10, sizeof(int), int_compare);
+
+    int first = 45, second = 67, third = 80, fourth = 120, fifth = 10000;
+
+    aws_priority_queue_push(&queue, &third);
```
JonathanHenson: ASSERT_SUCCESS

tests/priority_queue_test.c
```diff
@@ -0,0 +1,114 @@
...
+    aws_priority_queue_pop(&queue, &pop_val);
+    ASSERT_INT_EQUALS(fourth, pop_val, "Fourth element returned should have been %d but was %d", fourth, pop_val);
+    ASSERT_INT_EQUALS(pop_val, top_val, "Popped element should have been the top element. expected %d but was %d",
+            pop_val, top_val);
+
+    aws_priority_queue_top(&queue, &top_val);
+    aws_priority_queue_pop(&queue, &pop_val);
+    ASSERT_INT_EQUALS(fifth, pop_val, "Fifth element returned should have been %d but was %d", fifth, pop_val);
+
+    ASSERT_INT_EQUALS(AWS_ERROR_PRIORITY_QUEUE_EMPTY, aws_priority_queue_pop(&queue, &pop_val), "Popping from empty queue should result in error");
```
JonathanHenson: ASSERT_ERROR, we shouldn't be returning error codes, we return AWS_OP_ERR, and aws_last_error() exposes the error. the ASSERT_ERROR macro will do this check for you.

source/priority_queue.c
```diff
@@ -0,0 +1,139 @@
...
+
+    sift_up(queue, aws_array_list_length(&queue->container) - 1);
+
+    return AWS_OP_SUCCESS;
+}
+
+int aws_priority_queue_pop(struct aws_priority_queue *queue, void* item) {
+
+    if(0 == aws_array_list_length(&queue->container)) {
+        return AWS_ERROR_PRIORITY_QUEUE_EMPTY;
```
JonathanHenson: return aws_raise_error(AWS_ERROR_PRIORITY_QUEUE_EMPTY);

source/array_list.c
JonathanHenson: or better yet, 

size_t remainder = item_size & (SLICE - 1);

JonathanHenson: also clang and msvc love to complain about these not being cast to (void *)

source/array_list.c
marcomagdy: oops. That was copy/pasta.

source/priority_queue.c
marcomagdy: ok

source/priority_queue.c
marcomagdy: sure

tests/priority_queue_test.c
marcomagdy: ok

tests/priority_queue_test.c
marcomagdy: ok thanks

include/aws/common/priority_queue.h
marcomagdy: I will typedef it and document its semantics as @bdonlan mentioned.

marcomagdy:
The exponential growth behavior and the static behavior are relayed to the array_list and we already have tests for those.

I think what I'm missing is mixing push and pop as well as unit testing the swap operation in array_list.

JonathanHenson:
That's a fair point, my only concern is that array_list is an implementation detail, the growth behavior is part of the contract. How do we verify the contract didn't break if we ever swap out the implementation?

JonathanHenson:
A unit test, in addition to verifying behavior, sets the contract in explicit terms

include/aws/common/priority_queue.h
```diff
@@ -0,0 +1,99 @@
...
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/common.h>
+#include <aws/common/array_list.h>
+
+typedef int(*compare)(void *a, void *b);
```
JonathanHenson: This needs to be namespaced, it’s in a header file.

include/aws/common/priority_queue.h
```diff
@@ -0,0 +1,99 @@
...
+
+struct aws_priority_queue {
+    /**
+     * compare is a predicate that returns true (non-zero) if its first argument should come before its second argument.
+     * NOTE: priority_queue pops its largest element first. So, elements that come before all others will be out of the
```
JonathanHenson: Maybe “highest priority” instead of “largest”?

include/aws/common/priority_queue.h
```diff
@@ -0,0 +1,99 @@
...
+     * Copies the element of the highest priority, and removes it from the queue.. Complexity: O(log(n)).
+     * If queue is empty, AWS_ERROR_PRIORITY_QUEUE_EMPTY will be raised.
+     */
+    AWS_COMMON_API int aws_priority_queue_pop(struct aws_priority_queue *queue, void* item);
+
+    /**
+     * Copies the element of the highest priority. Complexity: constant time.
+     * If queue is empty, AWS_ERROR_PRIORITY_QUEUE_EMPTY will be raised.
+     */
+    AWS_COMMON_API int aws_priority_queue_top(struct aws_priority_queue *queue, void* item);
```
JonathanHenson: We don’t really need the over head of a copy on this right?

tests/priority_queue_test.c
```diff
@@ -0,0 +1,151 @@
...
+    ASSERT_SUCCESS(aws_priority_queue_push(&queue, &fourth), "Push operation failed for item %d", fourth);
+    ASSERT_SUCCESS(aws_priority_queue_push(&queue, &second), "Push operation failed for item %d", second);
+    ASSERT_SUCCESS(aws_priority_queue_push(&queue, &fifth), "Push operation failed for item %d", fifth);
+    ASSERT_SUCCESS(aws_priority_queue_push(&queue, &first), "Push operation failed for item %d", first);
+
+    size_t num_elements = aws_priority_queue_size(&queue);
+    ASSERT_INT_EQUALS(5, num_elements, "Priority queue size should have been %d but was %d", 5, num_elements);
+
+    int pop_val, top_val;
+    aws_priority_queue_top(&queue, &top_val);
```
JonathanHenson: ASSERT_SUCCESS

include/aws/common/priority_queue.h
bdonlan: Ah, but if you have multiple compare functions on the java side, you only have one trampoline (unless you do runtime code generation). How does the trampoline know which java-side function _object_ to call, for instance?

tests/priority_queue_test.c
```diff
@@ -41,32 +42,42 @@ static int test_priority_queue_preserves_order(struct aws_allocator *alloc, void
     ASSERT_INT_EQUALS(5, num_elements, "Priority queue size should have been %d but was %d", 5, num_elements);
 
     int pop_val, top_val;
-    aws_priority_queue_top(&queue, &top_val);
-    aws_priority_queue_pop(&queue, &pop_val);
+    err = aws_priority_queue_top(&queue, &top_val);
```
JonathanHenson: NIT, I see why you think this was a pain in the ass now:

All you had to do was:
ASSERT_SUCCESS(aws_priority_queue_dynamic_init(&queue, alloc, 10, sizeof(int), int_compare), "Priority queue initialization failed with error %d", err);

This is fine though.

source/array_list.c
bdonlan: Normally compilers can optimize `%` on powers of two, so in general I don't worry about this too much (unless it's not actually a power of two).

include/aws/common/priority_queue.h
```diff
@@ -0,0 +1,97 @@
...
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/common.h>
+#include <aws/common/array_list.h>
+
+typedef int(*aws_priority_queue_compare)(void *a, void *b);
```
bdonlan: Ergonomics: is `aws_priority_queue_` too long for a prefix?

tests/priority_queue_test.c
marcomagdy: If I had done that, the error message wouldn't include the error code. Which makes it less useful.
Anyways, vim makes that kind of stuff easy.

include/aws/common/priority_queue.h
bdonlan: Also, please precisely define the semantics of this operation. The semantics appear to differ from normal C comparator semantics (which is probably a bad idea). Ideally this should be a comparator that would also work in, say, qsort.

The pointers should also be `const void *`.

include/aws/common/priority_queue.h
marcomagdy: What are you suggesting to do instead?

source/priority_queue.c
```diff
@@ -0,0 +1,133 @@
...
+    while(left < len) {
+        aws_array_list_get_at_ptr(&queue->container, &left_item, left);
+
+        size_t right = left + 1;
+        if(right < len) {
+            aws_array_list_get_at_ptr(&queue->container, &right_item, right);
+
+            /* choose the larger/smaller of the two in case of a max/min heap respectively */
+            if (queue->pred(left_item, right_item)) {
+                left = right;
```
bdonlan: Don't call it left if it's not left.

include/aws/common/priority_queue.h
marcomagdy: I did document the semantics. https://github.com/awslabs/aws-c-common/pull/7/files#diff-dc4707d66f9b3bc969cb897313db7982R25

It's not the same semantics of a normal C comparator. In this case we need a predicate not the full comparison that returns negative/0/positive.

bdonlan: True, but if we define "The comparator should return a positive value if `a` should be returned before `b`; if `b` should come before `a` or the order is unspecified, it should return either a negative value or zero" then the semantics are defined precisely but are also compatible with generic sort routines.

source/priority_queue.c
marcomagdy: meh, it's a convenience, having left, right and swap made the code less readable.

include/aws/common/priority_queue.h
marcomagdy: > Ergonomics: is aws_priority_queue_ too long for a prefix?

Not with a good text editor 😛 
>The pointers should also be const void *.

Will change it.

>The comparator should return a positive value if a should be returned before b; if b should come before a or the order is unspecified, it should return either a negative value or zero"

Fine, I'll change them. But note that, like in std::priority_queue providing std::less<> results in a max heap. So, I'll reword what you suggested here to fit the same semantics while making it compatible with qsort

bdonlan: Rewording is fine, I'm just trying to avoid people shooting themselves in the foot when they accidentally pass a qsort-compatible function to this

marcomagdy: updated. And you're right. It's better this way.

include/aws/common/priority_queue.h
JonathanHenson: discussed offline, for perf reasons, let's have top fill in a pointer address, and pop make a copy.

Typical use case is, call top() evaluate if I actually want to pop(), then pop(). There's no need for a copy from top() in that use case

PR #8, opened 2018-02-12
Treat warnings as errors on the test project

marcomagdy:
*Description of changes:*
Treat warnings as errors in gcc/clang/msvc.
Test source files are #included in main.c so changed CMake to only build main.c instead of globbing all *.c files, otherwise we get many errors/warnings

### DIFF
--- a/tests/priority_queue_test.c
+++ b/tests/priority_queue_test.c
@@ -103,7 +103,7 @@ static int test_priority_queue_random_values(struct aws_allocator *alloc, void *
     int storage[SIZE], err;
     aws_priority_queue_static_init(&queue, storage, SIZE, sizeof(int), compare_ints);
     int values[SIZE];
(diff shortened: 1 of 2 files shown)


tests/priority_queue_test.c
```diff
@@ -103,7 +103,7 @@ static int test_priority_queue_random_values(struct aws_allocator *alloc, void *
     int storage[SIZE], err;
     aws_priority_queue_static_init(&queue, storage, SIZE, sizeof(int), compare_ints);
     int values[SIZE];
-    srand((unsigned)&queue);
+    srand((size_t)&queue);
```
JonathanHenson: I don't think this is right on 64 bit systems. It should probably be:

size_t queue_addr = (size_t)&queue;
srand((unsigned)queue_addr);

PR #9, opened 2018-02-13
Task scheduler implementation

marcomagdy:
*Description of changes:*


include/aws/common/task_scheduler.h
```diff
@@ -0,0 +1,101 @@
...
+    uint64_t owning_thread;
+    aws_task_scheduler_clock clock;
+};
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    /**
+     * Initializes a task scheduler instance with a clock function and config. A priority queue will be allocated
```
JonathanHenson: Doc is out of date, config should be allocator

include/aws/common/task_scheduler.h
```diff
@@ -0,0 +1,101 @@
...
+extern "C" {
+#endif
+
+    /**
+     * Initializes a task scheduler instance with a clock function and config. A priority queue will be allocated 
+     * in dynamic mode with initial size of 7
+     * The recommended clock functions are in <aws/common/clock.h>
+     *
+     * Note: This needs to be called from the thread that will own the scheduler. Cross-thread access is not allowed 
+     * and will always result in an error code being returned.
```
JonathanHenson: Error codes are now raised not returned

include/aws/common/task_scheduler.h
```diff
@@ -0,0 +1,101 @@
...
+     */
+    AWS_COMMON_API int aws_task_scheduler_init_default(struct aws_task_scheduler *scheduler, 
+            aws_task_scheduler_clock clock, struct aws_allocator *alloc);
```
JonathanHenson: By convention allocator is usually the second arg

source/task_scheduler.c
```diff
@@ -0,0 +1,123 @@
...
+
+struct task_container {
+    uint64_t timestamp;
+    struct aws_task task;
+};
+
+static int compare_timestamps(const void *a, const void *b) {
+    uint64_t a_time = ((struct task_container*)a)->timestamp;
+    uint64_t b_time = ((struct task_container*)b)->timestamp;
+    return a_time > b_time; // min-heap
```
JonathanHenson: Avoid c++ style comments

source/task_scheduler.c
```diff
@@ -0,0 +1,123 @@
...
+    return aws_priority_queue_dynamic_init(&scheduler->queue, alloc, DEFAULT_QUEUE_SIZE, sizeof(struct task_container), 
+            &compare_timestamps);
+}
+
+void aws_task_scheduler_init_with_queue(struct aws_task_scheduler *scheduler,
+        struct aws_priority_queue *queue, aws_task_scheduler_clock clock) {
+
+    scheduler->alloc = NULL;
+    scheduler->clock = clock;
+/* I feel icky about copying the queue. If we don't copy it, then who owns the queue? this is making me question this whole API. Need feedback */
```
JonathanHenson: I wouldn’t copy, I’d just take a reference

source/task_scheduler.c
```diff
@@ -0,0 +1,123 @@
...
+    scheduler->queue = *queue;
+    scheduler->owning_thread = aws_thread_current_thread_id();
+}
+
+void aws_task_scheduler_clean_up(struct aws_task_scheduler *scheduler) {
+     /* do we want to add a xthread check here? what happens if another thread attempts to clean up? */
+    aws_priority_queue_clean_up(&scheduler->queue);
```
JonathanHenson: If the user passed the queue we shouldn’t clean it up

tests/task_scheduler_test.c
```diff
@@ -0,0 +1,168 @@
...
+    aws_task_scheduler_init_default(&scheduler, aws_high_res_clock_get_ticks, alloc);
+
+    uint64_t current_clock_time;
+    ASSERT_SUCCESS(aws_high_res_clock_get_ticks(&current_clock_time), "Failed to get ticks from high-res clock.");
+
+    struct aws_task task2;
+    task2.fn = (aws_task_fn)2;
+    task2.arg = (void *)2;
+
+    //run 250 ms in the future.
```
JonathanHenson: Cpp style comments

include/aws/common/task_scheduler.h
```diff
@@ -0,0 +1,101 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/common.h>
+#include <aws/common/priority_queue.h>
+#include <stdint.h>
+
+typedef void *aws_task_arg;
```
bdonlan: Does this really need a typedef?

include/aws/common/task_scheduler.h
```diff
@@ -0,0 +1,101 @@
...
+    aws_task_scheduler_clock clock;
+};
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    /**
+     * Initializes a task scheduler instance with a clock function and config. A priority queue will be allocated 
+     * in dynamic mode with initial size of 7
```
bdonlan: Don't document initial sizes and other such implementation details in the .h file. These are not part of your API contract.

include/aws/common/task_scheduler.h
bdonlan: Don't commit to actively detecting illegal x-thread access and reporting it. This can be expensive, depending on the platform.

Furthermore, _why_ is it illegal to move a task scheduler around? Couldn't I use it safely on multiple threads if I added appropriate locking around it? Should all the other datastructures guard against cross-thread access too?

include/aws/common/task_scheduler.h
```diff
@@ -0,0 +1,101 @@
...
+     * The recommended clock functions are in <aws/common/clock.h>
+     *
+     * Note: This needs to be called from the thread that will own the scheduler. Cross-thread access is not allowed 
+     * and will always result in an error code being returned.
+     */
+    AWS_COMMON_API int aws_task_scheduler_init_default(struct aws_task_scheduler *scheduler,
```
bdonlan: We keep writing into preallocated structures everywhere. Are we really okay with throwing ABI forward-compatibility to the wind?

include/aws/common/task_scheduler.h
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Initializes a task scheduler instance with a priority queue, clock function and config.
+     * The recommended clock functions are in <aws/common/clock.h>
+     *
+     * Note: This needs to be called from the thread that will own the scheduler. Cross-thread access is not allowed 
+     * and will always result in an error code being returned.
+     */
+    AWS_COMMON_API void aws_task_scheduler_init_with_queue(struct aws_task_scheduler *scheduler, 
+            struct aws_priority_queue *queue, aws_task_scheduler_clock clock);
```
bdonlan: Why are we exposing implementation details like the use of a prioqueue?

include/aws/common/task_scheduler.h
```diff
@@ -0,0 +1,101 @@
...
+     * Gets the next task that is ready for execution. Tasks in the future will not be returned, 
+     * even if they are the highest priority. next_run_time, if it is not null, and there are any tasks scheduled
+     * will be set to the timestamp for the highest priority task. This is a useful hint for setting timeouts on
+     * event loops, or thread sleeps. If no tasks are scheduled, this value will be set to 0. task is copied
+     */
+    AWS_COMMON_API int aws_task_scheduler_next_task(struct aws_task_scheduler *scheduler, struct aws_task *task, 
+            uint64_t *next_run_time);
+
+    /**
+     * Schedules a task to run at whatever the current return value of the clock fn is. task is copied
```
bdonlan: const the task argument

include/aws/common/task_scheduler.h
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Cleans up any memory allocated, and prepares the instance for reuse or deletion.
+     */
+    AWS_COMMON_API void aws_task_scheduler_clean_up(struct aws_task_scheduler *scheduler);
+
+    /**
+     * Gets the next task that is ready for execution. Tasks in the future will not be returned, 
+     * even if they are the highest priority. next_run_time, if it is not null, and there are any tasks scheduled
+     * will be set to the timestamp for the highest priority task. This is a useful hint for setting timeouts on
+     * event loops, or thread sleeps. If no tasks are scheduled, this value will be set to 0. task is copied
```
bdonlan: How do we indicate when no task is ready for execution?

include/aws/common/task_scheduler.h
```diff
@@ -0,0 +1,101 @@
...
+    AWS_COMMON_API int aws_task_scheduler_next_task(struct aws_task_scheduler *scheduler, struct aws_task *task, 
+            uint64_t *next_run_time);
+
+    /**
+     * Schedules a task to run at whatever the current return value of the clock fn is. task is copied
+     */
+    AWS_COMMON_API int aws_task_scheduler_schedule_now(struct aws_task_scheduler *queue, struct aws_task *task);
+
+    /**
+     * Schedules a task to run at time_to_run relative to the clock fn. task is copied
```
bdonlan: Don't talk about the clock fn in every function. Just mention in the initialize calls that it is used to query for the current time, and explain elsewhere that e.g. `task is queued to execute time_to_run time units after the current time`

source/task_scheduler.c
bdonlan: This is one reason why putting complex datastructures in caller-allocated fields can be troublesome. How do you safely move them? Do they have internal pointers-to-self that would make a simple copy unsafe? You don't know, unless you make assumptions of the implementation of the datastructure in question.

It's much cleaner to just always allocate these datastructures on the heap; platforms with difficulty allocating on the heap can use a custom allocator as an escape hatch here.

source/task_scheduler.c
```diff
@@ -0,0 +1,123 @@
...
+            return AWS_OP_ERR;
+        }
+
+        if (possible_task->timestamp <= now) {
+            struct task_container container;
+            if(aws_priority_queue_pop(&scheduler->queue, (void *)&container)) {
+                return AWS_OP_ERR;
+            }
+            *task = container.task;
+            return AWS_OP_SUCCESS;
```
bdonlan: next_run_time is incorrect at this point, since the 'next run time' would be that of the new top of the heap.

tests/task_scheduler_test.c
```diff
@@ -0,0 +1,168 @@
...
+    struct aws_task task_to_run;
+
+    ASSERT_SUCCESS(aws_task_scheduler_next_task(&scheduler, &task_to_run, 0), 
+            "Task pop on a now() task, should return on the first try");
+
+    ASSERT_TRUE(task1.fn == task_to_run.fn, "Popped task should have been task 1.");
+    ASSERT_TRUE(task1.arg == task_to_run.arg, "Popped task arg should have been task 1.");
+
+    uint64_t timestamp = 0;
+    while (aws_task_scheduler_next_task(&scheduler, &task_to_run, &timestamp) != AWS_OP_SUCCESS) {
```
bdonlan: Use a mock clock, so you don't need to do busy loops like this.

tests/task_scheduler_test.c
```diff
@@ -0,0 +1,168 @@
...
+    aws_task_scheduler_init_default(&scheduler, aws_high_res_clock_get_ticks, alloc);
+
+    uint64_t current_clock_time;
+    ASSERT_SUCCESS(aws_high_res_clock_get_ticks(&current_clock_time), "high res clock ticks failed");
+
+    struct aws_task task;
+    task.fn = (aws_task_fn)0;
+    task.arg = (void *)0;
+
+    //clocks are in nanoseconds. Add 10 milliseconds to give the test time to run.
```
bdonlan: You can easily get a 10ms hiccup in your test if your thread is preempted. A mock clock helps avoid this

include/aws/common/task_scheduler.h
marcomagdy: I already removed this API locally. It doesn't make sense because besides exposing implementation details, it requires the caller/owner-of-the-queue to know details about items inserted into the queue. This was a carry over, that doesn't fit in the new design.

include/aws/common/task_scheduler.h
marcomagdy: So between answering this question and fixing the bug you mentioned below (incorrect timestamp) this API has become too complicated for my taste. There're four possible outcomes to the two output parameters. The edge case is when there's only one task left in the queue, what should the value of the "next_run_time" be. Ideally return a different error code. But the implementation becomes too "branchy".

It's better to split those two functions. aws_task_scheduler_next_task() and aws_task_scheduler_next_run_time().

Thoughts?

include/aws/common/task_scheduler.h
marcomagdy: Not really. I was wondering myself when porting the code. @JonathanHenson thoughts on this?

JonathanHenson: I'd typedef the function, but not the arg.

include/aws/common/task_scheduler.h
JonathanHenson: The task scheduler design is exclusively for single-threaded scenarios, and where threads are being crossed, we want something like an event loop to do the messaging. This will make broken code look broken at debug and development time. 

Arguably we could forgo this and do a wrapper in aws-c-io, that does the check, but some other products at Amazon passed on their wisdom to me from some COEs that asserting or erroring on an accidental thread jump will keep it from accidentally occurring.

include/aws/common/task_scheduler.h
JonathanHenson: I don't think it's necessarily being thrown to the wind, there's other mechanisms for establishing ABI compat than heap allocating everything.

Besides, it's not like the problem magically goes away with a _new call anyways, you'd still encounter different libs making the same call. You'd still have to do size checks everywhere. 

I'd rather nail these data structure APIs and not need to change them than start the habbit of forcing everything to dynamic memory and fragmented allocations all the way at the bottom of the dependency tree.

include/aws/common/task_scheduler.h
JonathanHenson: That's not what the purpose of that API was, the purpose was to give the user the ability to control how the priority queue is initialized (e.g. with preallocated memory in static mode).

source/task_scheduler.c
JonathanHenson: I honestly don't know what's complicated about taking a reference to the queue and not copying it. If the user passed, it the user is responsible for cleaning it up at the appropriate time.

include/aws/common/common.h
```diff
@@ -113,6 +113,10 @@ typedef enum aws_common_error {
     AWS_ERROR_LIST_STATIC_MODE_CANT_SHRINK,
     AWS_ERROR_PRIORITY_QUEUE_FULL,
     AWS_ERROR_PRIORITY_QUEUE_EMPTY,
+    AWS_ERROR_TASK_SCHEDULER_NO_TASKS,
```
JonathanHenson: I don't see where you've added the error strings to common.c

PR #13, opened 2018-02-23
Implementation of Robin Hood hashing for AWS C Common

bdonlan:
*Issue #, if available:*

*Description of changes:*
Implements a generic hash table for the AWS C Common library, based on robinhood hashing. Bob Jenkins' lookup3 is used for generic hash functions. This also includes some checked multiply routines to help avoid integer overflow on allocations within the hash table resize logic.


include/aws/common/hash_table.h
```diff
@@ -0,0 +1,249 @@
...
+
+#include <aws/common/common.h>
+#include <aws/common/error.h>
+#include <stdint.h>
+
+#define AWS_COMMON_HASH_TABLE_ITER_CONTINUE 0
+#define AWS_COMMON_HASH_TABLE_ITER_DELETE 1
+#define AWS_COMMON_HASH_TABLE_ITER_STOP 2
+
+/**
```
JonathanHenson: This documentation and api is excellent

include/aws/common/math.h
```diff
@@ -0,0 +1,189 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+/**
+ * Multiplies a * b. If the result overflows, returns 2^64 - 1.
+ */
+static inline uint64_t aws_common_mul_u64_saturating(uint64_t a, uint64_t b) {
+#if defined(__x86_64__) && defined(__GNUC__)
```
JonathanHenson: As much as I love this, you need to convert this to msvc intrinsics for Windows builds.

include/aws/common/math.h
```diff
@@ -0,0 +1,189 @@
...
+    }
+#endif
+}
+
+/**
+ * Multiplies a * b. If the result overflows, returns 2^32 - 1.
+ */
+static inline uint32_t aws_common_mul_u32_saturating(uint32_t a, uint32_t b) {
+#if (defined(__i386__) || defined(__x86_64__)) && defined(__GNUC__)
+    // We can use inline assembly to do this efficiently on x86-64 and x86.
```
JonathanHenson: Nit, let’s avoid c++ style comments for consistency.

include/aws/common/math.h
```diff
@@ -0,0 +1,189 @@
...
+
+    // we specify edx as an output, rather than a clobber, because we want to allow it
+    // to be allocated as an input register
+    uint32_t edx;
+    __asm__(
```
JonathanHenson: For compatibility should these be just asm instead of __asm__?

include/aws/common/math.h
```diff
@@ -0,0 +1,189 @@
...
+ */
+static inline int aws_common_mul_u32_checked(uint32_t a, uint32_t b, uint32_t *r) {
+#if (defined(__i386__) || defined(__x86_64__)) && defined(__GNUC__)
+    // We can use inline assembly to do this efficiently on x86-64 and x86.
+    uint32_t result;
+    int flag;
+    // Note: We use SETNO which only takes a byte register. To make this easy,
+    // we'll write it to dl (which we throw away anyway) and mask off the high
+    // bits.
+    __asm__(
```
JonathanHenson: Since these are in a static inline function (and thus could be used inside an inner loop) should they by volatile as well?

JonathanHenson: I should have said, will be substituted as is

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+                e->element.key, e->element.value, e->hash_code,
+                (i - e->hash_code) & state->mask);
+        }
+    }
+}
+#endif
+
+#if 0
+// Not currently exposed as an API. Should we have something like this? Useful for benchmarks
+AWS_COMMON_API void aws_common_hash_table_print_stats(struct aws_common_hash_table *table) {
```
JonathanHenson: Even if it was exposed, you don’t need the dll_export in the c file, you can delete AWS_COMMON_API

JonathanHenson: Also it may be better to take a FILE* param

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+    template.alloc = alloc;
+
+    template.entry_count = 0;
+    template.max_load_factor = 0.95; // TODO - make configurable?
+
+    update_template_size(&template, size);
+    map->pImpl = alloc_state(&template);
+
+    if (!map->pImpl) {
+        return AWS_ERROR_OOM;
```
JonathanHenson: These should be using aws_raise_error

tests/hash_table_test.c
```diff
@@ -0,0 +1,411 @@
...
+
+AWS_TEST_CASE(test_hash_table_put_get, test_hash_table_put_get_fn)
+static int test_hash_table_put_get_fn(struct aws_allocator *alloc, void *ctx) {
+    struct aws_common_hash_table hash_table;
+    int err_code = aws_common_hash_table_init(&hash_table, alloc, 10, aws_common_hash_string,
+        aws_common_string_eq, NULL);
+    struct aws_common_hash_element *pElem;
+    int was_created;
+
+    ASSERT_INT_EQUALS(AWS_ERROR_SUCCESS, err_code,
```
JonathanHenson: Once you fix the raise error issue these would all be ASSERT_SUCCESS AND ASSERT_ERROR

include/aws/common/math.h
JonathanHenson: Same comments as above for this.... just transpose them for 32 bit arithmetic.

include/aws/common/math.h
```diff
@@ -0,0 +1,189 @@
...
+
+static inline int aws_common_mul_size_checked(size_t a, size_t b, size_t *r) {
+    if ((uint64_t)SIZE_MAX == (uint64_t)~(uint32_t)0) {
+        return aws_common_mul_u32_checked(a, b, (uint32_t*)r);
+    } else {
+        return aws_common_mul_u64_checked(a, b, (uint64_t*)r);
+    }
+}
+
+
```
JonathanHenson: why so many newlines?

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
+/*
+* Copyright 2010-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
JonathanHenson: 2010-2018

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+    (void)hashlittle;
+    (void)hashbig;
+}
+
+struct hash_table_entry {
+    struct aws_common_hash_element element;
+    uint64_t hash_code; // hash code (0 signals empty)
+};
+
+struct hash_table_state {
```
JonathanHenson: AWS_CACHE_ALIGN ?

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+#include <aws/common/linked_list.h>
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <limits.h>
+
+// Include lookup3.c so we can (potentially) inline it and make use of the mix() macro.
+#include "private/lookup3.c"
+
+static void suppress_unused_lookup3_func_warnings() {
```
JonathanHenson: should we inline this? This would avoid the function overhead just for a warning suppression in the hash_for fn

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+    return (int)(entry - map->slots);
+}
+
+#if 0
+// Useful debugging code for anyone working on this in the future
+static uint64_t distance(struct hash_table_state *state, int index) {
+    return (index - state->slots[index].hash_code) & state->mask;
+}
+
+void hash_dump(struct aws_common_hash_table *tbl) {
```
JonathanHenson: rather than printf, it's probably more usable if we do fprintf etc... and take a FILE *fd parameter, I can always pass stdout, stderr, or an actual log file to it that way.

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+    struct hash_table_state *state = tbl->pImpl;
```
JonathanHenson: Is this missing a cast?

source/hash_table.c
JonathanHenson: also alloc_state can return null if the requested size is too large. do we need a more granular error code for that?

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+    template.max_load_factor = 0.95; // TODO - make configurable?
+
+    update_template_size(&template, size);
+    map->pImpl = alloc_state(&template);
+
+    if (!map->pImpl) {
+        return AWS_ERROR_OOM;
+    }
+
+    return AWS_ERROR_SUCCESS;
```
JonathanHenson: AWS_OP_SUCCESS

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+// Returns AWS_ERROR_SUCCESS if the item existed (leaving it in *entry),
+// or AWS_ERROR_HASHTBL_ITEM_NOT_FOUND if it did not (putting its destination
+// in *entry). Note that this does not take care of displacing whatever was in that entry before.
+
+// probe_idx is set to the probe index of the entry.
+
+static int find_entry1(struct hash_table_state *state, uint64_t hash_code, const void *key, struct hash_table_entry **p_entry, int *p_probe_idx);
+
+// Inlined fast path: Check the first slot, only.
+static int find_entry(struct hash_table_state *state, uint64_t hash_code, const void *key, struct hash_table_entry **p_entry, int *p_probe_idx) __attribute__((always_inline));
```
JonathanHenson: this will only compile on GCC and CLang, let's add a macro definition that handles msvc as well in Common.h

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+static int find_entry(struct hash_table_state *state, uint64_t hash_code, const void *key, struct hash_table_entry **p_entry, int *p_probe_idx) {
+    struct hash_table_entry *entry = &state->slots[hash_code & state->mask];
+
+    if (entry->hash_code == 0) {
+        if (p_probe_idx) *p_probe_idx = 0;
+        *p_entry = entry;
+        return AWS_ERROR_HASHTBL_ITEM_NOT_FOUND;
```
JonathanHenson: return aws_raise_error(AWS_ERROR_HASHTBL_ITEM_NOT_FOUND);

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+    }
+
+    if (entry->hash_code == hash_code && state->equals_fn(key, entry->element.key)) {
+        if (p_probe_idx) *p_probe_idx = 0;
+        *p_entry = entry;
+        return AWS_ERROR_SUCCESS;
```
JonathanHenson: AWS_OP_SUCCESS

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+    // but if not we'll overwrite the deleted entry).
+    int probe_cap = INT_MAX;
+
+    int rv;
+    struct hash_table_entry *entry;
+    do {
+        int index = (hash_code + probe_idx) & state->mask;
+        entry = &state->slots[index];
+        if (!entry->hash_code) {
+            rv = AWS_ERROR_HASHTBL_ITEM_NOT_FOUND;
```
JonathanHenson: rv = aws_raise_error(AWS_ERROR_HASHTBL_ITEM_NOT_FOUND) ?

Alternatively, do this, and have a branch at the end that echecks if rv is 0 or -1.

JonathanHenson: same comment for each branch in this loop.

Also it may make since to use the AWS_LIKELY/UNLIKELY in this inner loop.

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+    return rv;
+}
+
+int aws_common_hash_table_find(struct aws_common_hash_table *map,
+    const void *key, struct aws_common_hash_element **pElem) {
+    struct hash_table_state *state = map->pImpl;
+    uint64_t hash_code = hash_for(state, key);
+    struct hash_table_entry *entry;
+
+    int rv = find_entry(state, hash_code, key, &entry, NULL);
```
JonathanHenson: if(!find_entry...) {
    *pElem = &entry->element;
    return AWS_OP_SUCCESS;
}

return AWS_OP_ERR;

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+    }
+
+    return initial_placement;
+}
+
+static int expand_table(struct aws_common_hash_table *map) {
+    struct hash_table_state *old_state = map->pImpl;
+    struct hash_table_state template = *old_state;
+
+    update_template_size(&template, template.size * 2);
```
JonathanHenson: error code check?

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+
+    struct hash_table_state *new_state = alloc_state(&template);
+    if (!new_state) {
+        return AWS_ERROR_OOM;
```
JonathanHenson: This could be cleaner and future granularity proof, if alloc_state called aws_raise_error and if this returns NULL then just return AWS_OP_ERR  and the error code will already be set.

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+int aws_common_hash_table_create(struct aws_common_hash_table *map,
+    const void *key, struct aws_common_hash_element **pElem, int *was_created
+) {
+    struct hash_table_state *state = map->pImpl;
+    uint64_t hash_code = hash_for(state, key);
+    struct hash_table_entry *entry;
+    int probe_idx;
+    int ignored;
+    if (!was_created) {
+        was_created = &ignored;
```
JonathanHenson: should ignored have been initialized at this point?

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+        index = next_index;
+    }
+
+    // Clear the entry we shifted out of
+    memset(&state->slots[index], 0, sizeof(*entry));
+
+    return index;
+}
+
+int aws_common_hash_table_remove(struct aws_common_hash_table *map,
```
JonathanHenson: If the system running on this hash table hits a burst of activity this table could get large, and then when the system goes back to a normal state, all of this memory will still be laying around.

You could argue that we should optimize for the heavy load path not the light load path, which is fair; but it's also possible that different parts of the system (like in a video game) encounter heavy loads at different times and we wouldn't release the memory we no longer need, but another part of the system desperately needs. 

Should we at least have an API that allows a user to check the load factor, and free up space down to the ideal load factor?

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+    } else if (state->destroy_fn) {
+        state->destroy_fn(entry->element);
+    }
+
+    remove_entry(state, entry);
+
+    return AWS_ERROR_SUCCESS;
+}
+
+int aws_common_hash_table_foreach(struct aws_common_hash_table *map,
```
JonathanHenson: Maybe add in the API documentation, that any iteration ordering is purely an accidental result of the implementation and should not be relied upon, for any reason, ever.

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+        struct hash_table_entry *entry = &state->slots[i];
+
+        if (!entry->hash_code) {
+            continue;
+        }
+
+        switch (callback(baton, &entry->element)) {
+            case AWS_COMMON_HASH_TABLE_ITER_CONTINUE:
+                break; // out of switch
+            case AWS_COMMON_HASH_TABLE_ITER_DELETE: {
```
JonathanHenson: what if I want to delete AND stop?

JonathanHenson: maybe a bit field would be better?

include/aws/common/math.h
bdonlan: I'm not set up to test those currently.

include/aws/common/math.h
bdonlan: Nope. The difference is `asm` will result in compile failures in a strictly-conforming mode in GCC (e.g. `--std=c99 --pedantic`).

include/aws/common/math.h
```diff
@@ -0,0 +1,189 @@
...
+static inline uint64_t aws_common_mul_u64_saturating(uint64_t a, uint64_t b) {
+#if defined(__x86_64__) && defined(__GNUC__)
+    // We can use inline assembly to do this efficiently on x86-64 and x86.
+
+    // we specify rdx as an output, rather than a clobber, because we want to
+    // allow it to be allocated as an input register
+
+    uint64_t rdx;
+    __asm__(
+        "mulq %q2\n" // rax * b, result is in RDX:RAX, OF=CF=(RDX != 0)
```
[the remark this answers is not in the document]
bdonlan: That appears to be a msvc intrinsic, unavailable on GCC. Though on GCC I think there's a 128-bit integer type I might be able to use.

include/aws/common/math.h
```diff
@@ -0,0 +1,189 @@
...
+ * overflows, returns 0, else returns 1.
+ */
+static inline int aws_common_mul_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
+#if defined(__x86_64__) && defined(__GNUC__)
+    // We can use inline assembly to do this efficiently on x86-64 and x86.
+
+    int flag;
+    uint64_t result;
+    __asm__(
+        "mulq %q3\n" // rax * b, result is in RDX:RAX, OF=CF=(RDX != 0)
```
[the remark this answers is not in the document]
bdonlan: Note that using `mulq` lets us skip a comparison as it sets `OF` implicitly.

include/aws/common/math.h
bdonlan: I don't understand, why is volatile necessary in any of this?

source/hash_table.c
bdonlan: GCC, at least, will inline a static void function with only one call site automatically.

source/hash_table.c
bdonlan: This is a heap-allocated structure, adding an alignment annotation is a promise to the compiler, but not something that impacts the allocation directly.

It might be useful to align `slots` I suppose, but if we make that promise to the compiler then we need to have an aligned allocation function to use to allocate the state table.

source/hash_table.c
bdonlan: C implicitly casts to/from `void*`

source/hash_table.c
bdonlan: I don't think a more granular error is needed, the semantics of `AWS_ERROR_OOM` or `NULL` return from allocate are already "You asked for more memory than I can give you"

source/hash_table.c
bdonlan: Nope. We're writing into it, not reading from it.

source/hash_table.c
bdonlan: Yes, we could have a `_shrink` function. I've left it out as I'm presenting the required part of the API first.

source/hash_table.c
bdonlan: Well, you could return `_STOP` next time through the loop I suppose.

include/aws/common/math.h
JonathanHenson: Too keep it from getting moved by the compiler if it’s inline in certain scenarios we haven’t predicted?

source/hash_table.c
JonathanHenson: Fair enough

source/hash_table.c
JonathanHenson: We probable need one of those anyways

include/aws/common/math.h
bdonlan: If you're referring to `asm volatile`, this is only needed if the code has desired side effects other than its out parameters. Here, though, we're only interested in those out parameters; we're perfectly happy to have the compiler do things like delete it as part of dead value elimination, so we should avoid `volatile`.

source/hash_table.c
JonathanHenson: I thought about that too. I'm not opposed. I'm not sure which is less messy though. A Bit field seems cleanest to me.

include/aws/common/math.h
```diff
@@ -0,0 +1,189 @@
...
+    // to be allocated as an input register
+    uint32_t edx;
+    __asm__(
+        "mull %k3\n" // eax * b, result is in EDX:EAX, OF=CF=(EDX != 0)
+#ifdef __x86_64__
+        "cmovc %k4, %%eax\n"
+#else
+        // cmov isn't guaranteed to be available on x86-32
+        "jno .1f\n"
+        "mov $0xFFFFFFFF, %%eax\n"
```
bdonlan: Note to self: use %k4 here

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+// deletions, see:
+// http://codecapsule.com/2013/11/17/robin-hood-hashing-backward-shift-deletion/
+
+#include <aws/common/hash_table.h>
+#include <aws/common/common.h>
+#include <aws/common/math.h>
+#include <aws/common/linked_list.h>
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
```
marcomagdy: stdio only needed for debugging functions. Guard it with if 0? or include it in the debugging code section.

source/hash_table.c
```diff
@@ -0,0 +1,582 @@
...
+    printf("\n");
+}
+#endif
+
+// Given a header template, allocates space for a hash table of the appropriate size, and copies the state header
+// into this allocated memory, which is returned.
+static struct hash_table_state *alloc_state(const struct hash_table_state *template) {
+    size_t elemsize;
+
+    if (!aws_common_mul_size_checked(template->size - 1, sizeof(template->slots[0]), &elemsize)) {
```
marcomagdy: can you clarify why -1?

JonathanHenson:
can you resolve the conflicts for us? I like the formatting updates Marco made, but they need to be merged with your updates.

marcomagdy:
closing this in favor of https://github.com/awslabs/aws-c-common/pull/17

PR #20, opened 2018-03-19
String split implementations (needed for http as well as alpn lists).

JonathanHenson:
String split implementations (needed for http as well as alpn lists).


include/aws/common/byte_buf.h
```diff
@@ -35,5 +36,49 @@ static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t c_str[],
...
+ * if the string begins with split_on, an empty string will be the first entry in output
+ * if the input string has two adjacent split_on tokens, an empty string will be inserted into the output.
+ * if the input string ends with split_on, it will be ignored.
+ *
+ * It is the user's responsibility to properly initialize output. Recommended number of preallocated elements from output
+ * is your most likely guess for the upper bound of the number of elements resulting from the split.
+ *
+ * It is the user's responsibility to make sure the input buffer stays in memory long enough to use the results.
+ *
+ * We assume split_on is a literal and will call strlen() on it.
```
bdonlan: What about binary strings?

source/byte_buf.c
```diff
@@ -0,0 +1,109 @@
...
+
+int aws_string_split_on_char(struct aws_byte_buf *input_str, char split_on, struct aws_array_list *output) {
+    assert(input_str);
+    assert(output);
+    assert(output->item_size >= sizeof(struct aws_byte_buf));
+
+    size_t current_pos = 0;
+    size_t last_offset = 0;
+
+    for(; current_pos < input_str->len; ++current_pos) {
```
bdonlan: Use `memchr`

source/byte_buf.c
```diff
@@ -0,0 +1,109 @@
...
+
+        if (AWS_UNLIKELY(aws_array_list_push_back(output, (const void *)&buffer))) {
+            return AWS_OP_ERR;
+        }
+    }
+
+    return AWS_OP_SUCCESS;
+}
+
+int aws_string_split_on_str(struct aws_byte_buf *input_str, char *split_on,
```
bdonlan: What is the intended behavior when `input_str` contains NULs?

source/byte_buf.c
```diff
@@ -0,0 +1,109 @@
...
+                if (AWS_UNLIKELY(aws_array_list_push_back(output, (const void *)&buffer))) {
+                    return AWS_OP_ERR;
+                }
+
+                last_offset = current_pos + 1;
+                possible_match_idx = 0;
+            }
+        }
+        else {
+            possible_match_idx = 0;
```
bdonlan: This will fail to identify strings with a repeated prefix. Here's an example:

```
input_str: "FOFOFOO"
split_on: "FOFOO"
```

After matching `FOFO`, we have `current_pos = 4`, `possible_match_idx = 4`. We then see that `'O' != 'F'` and reset `possible_match_idx = 0`. We then retry matching `FOFOO` starting at `current_pos = 5 (OO)`, instead of back at `current_pos = 2 (FOFOO)`

source/byte_buf.c
```diff
@@ -0,0 +1,109 @@
...
+    assert(token_len);
+
+    size_t current_pos = 0;
+    size_t last_offset = 0;
+
+    int8_t possible_match_idx = 0;
+
+    for(; current_pos < input_str->len; ++current_pos) {
+
+        if (current_pos < input_str->len - 2 && input_str->buffer[current_pos] == (uint8_t)split_on[possible_match_idx]) {
```
bdonlan: Substring search is tricky to do both correctly and efficiently. Use `strstr` (standard) or `memmem` (nonstandard).

If you intend to implement NULs in the `input_str` but not `split_on`, you can use `strstr` and restart the search if a NUL is encountered before end of string. If you intend to implement NULs in both, you may need to implement the Boyer-Moore string search algorithm (or similar).

source/byte_buf.c
```diff
@@ -0,0 +1,109 @@
...
+                last_offset = current_pos + 1;
+                possible_match_idx = 0;
+            }
+        }
+        else {
+            possible_match_idx = 0;
+        }
+    }
+
+    /* in this case we're only here if we didn't read the entire split token so we don't account for the token length */
```
bdonlan: This is a confusing comment. Maybe just `/* Because we did not encounter the split token for the last chunk, we do not need to subtract token length here */`

include/aws/common/byte_buf.h
```diff
@@ -35,5 +36,49 @@ static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t c_str[],
...
+/**
+ * No copies, no string allocations. Fills in output with a list of aws_byte_buf instances where buffer is
+ * an offset into the input_str and len is the length of that string in the original buffer.
+ *
+ * Edge case rules are as follows:
+ * if the string begins with split_on, an empty string will be the first entry in output
+ * if the input string has two adjacent split_on tokens, an empty string will be inserted into the output.
+ * if the input string ends with split_on, it will be ignored.
+ *
+ * It is the user's responsibility to properly initialize output. Recommended number of preallocated elements from output
```
bdonlan: What happens if output is fixed-sized and too small to accommodate all of the tokens?

Also, it's very useful to have a split with a limited number of tokens, where any additional tokens after the first N are left appended to the last token.

include/aws/common/byte_buf.h
JonathanHenson: it wasn't really designed for that use case, but I don't see any reason we couldn't change that interface to use aws_byte_buf for split_on instead of a const char *

source/byte_buf.c
JonathanHenson: good idea.

source/byte_buf.c
JonathanHenson: If NULLS are in input str, we don't pay attention to them unless they're in the split criteria.

source/byte_buf.c
JonathanHenson: that is better, will update.

source/byte_buf.c
bdonlan: The reason I ask is that `strstr` will not deal with embedded NULs.

include/aws/common/byte_buf.h
JonathanHenson: AWS_OP_ERR will be returned and the error will be in aws_last_error().

I'm not sure I see the usecase for your second <p/>

source/byte_buf.c
JonathanHenson: how would strstr() be helpful if we are allowing binary splits?

JonathanHenson: I think this whole thing get's simpler if we don't support NULL and I don't have a use case for binary data. For binary data, I'd typically not be reading and then splitting. It would usually have some sort of framing.... I'd be content to just make this a string-only based API.

source/byte_buf.c
JonathanHenson: and I think that's a great point.... see below. I don't have a usecase in hand for supporting NULLs, and like the simplicity of applying YAGNI here.

include/aws/common/byte_buf.h
bdonlan: Example usecase: I have a S3 path and I want to split it into bucket and key. I have a CSV line and I really only care about the third field. Etc.

source/byte_buf.c
JonathanHenson: done

source/byte_buf.c
JonathanHenson: I'm deleting the string search. Basically I refuse to put unsafe code in our common-lib, and strstr is fundamentally unsafe. I don't want to waste Amazon's money by spending hours reemplementing boyer-moore. And now that I think about it, this function wouldn't be that helpful for http anyways (seeing as how I don't have a way of knowing where the body would begin ahead of time anyways.

I'm deleting that function, will add it when we need it.

source/byte_buf.c
```diff
@@ -0,0 +1,55 @@
...
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/byte_buf.h>
+#include <assert.h>
+
+int aws_string_split_on_char(struct aws_byte_buf *input_str, char split_on, struct aws_array_list *output) {
```
bdonlan: I still hold that having a 'first N splits' mode would be very useful

tests/string_test.c
```diff
@@ -0,0 +1,194 @@
...
+    ASSERT_BIN_ARRAYS_EQUALS(expected, strlen(expected), value.buffer,  value.len, "Value 0 should have been %s", expected);
+
+    ASSERT_SUCCESS(aws_array_list_get_at(&output, &value, 1), "Retrieving element at 1 failed.");
+    expected = "testb";
+    ASSERT_BIN_ARRAYS_EQUALS(expected, strlen(expected), value.buffer,  value.len, "Value 1 should have been %s", expected);
+
+    ASSERT_SUCCESS(aws_array_list_get_at(&output, &value, 2), "Retrieving element at 2 failed.");
+    expected = "testc";
+    ASSERT_BIN_ARRAYS_EQUALS(expected, strlen(expected), value.buffer,  value.len, "Value 2 should have been %s", expected);
+
```
bdonlan: Should there be an empty split at the end? It seems strange that "a;b" and "a;b;" should have the same results.

tests/string_test.c
```diff
@@ -0,0 +1,194 @@
...
+    ASSERT_SUCCESS(aws_array_list_get_at(&output, &value, 3), "Retrieving element at 3 failed.");
+    expected = "testc";
+    ASSERT_BIN_ARRAYS_EQUALS(expected, strlen(expected), value.buffer,  value.len, "Value 3 should have been %s", expected);
+
+    aws_array_list_clean_up(&output);
+
+    return 0;
+}
+
+AWS_TEST_CASE(test_char_split_adj_tokens, test_char_split_adj_tokens_fn)
```
bdonlan: Missing edge case: Output array list is too small and can't be expanded.

tests/string_test.c
david-koenig: I agree that having an empty string at the end seems to be the most logically consistent. That being said, he did clearly document that he was going the opposite way. I tend to side with Bryan on this, but I don't feel that strongly and wouldn't block on this issue.

source/byte_buf.c
JonathanHenson: it's an extra branch, but that's fine, implemented.

JonathanHenson: wrong comment one sec.

tests/string_test.c
JonathanHenson: it's an extra branch, but that's fine, implemented.

tests/string_test.c
JonathanHenson: updated.

include/aws/common/byte_buf.h
```diff
@@ -39,6 +40,34 @@ static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t c_str[],
     return (struct aws_byte_buf){.buffer = (uint8_t *)&c_str[0], .len = len};
 }
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * No copies, no string allocations. Fills in output with a list of aws_byte_buf instances where buffer is
```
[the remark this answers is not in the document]
JonathanHenson: That's not my intention with byte_buffer. It was merely to join len and array into a single object and to not pass c strings around.

source/byte_buf.c
JonathanHenson: added.

include/aws/common/byte_buf.h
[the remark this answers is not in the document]
david-koenig: The way the aws_byte_cursor_advance* functions are written, they are changing the value of the pointer and updating the length to be the length remaining in the array. @bdonlan was originally using the byte_buffer struct to keep that pointer and length together as a single entity, but that concerned me, because if we're dealing with heap-allocated memory we would want to keep a pointer to the beginning of it for freeing later. Also, I felt that the name "buffer" implied that this was the entire array of memory and not just something pointing to somewhere in the middle of it. I could easily see someone who picked this code up later getting confused and trying to free the memory somewhere in the middle of the array and causing a crash.

This is why we changed those functions to be working with the byte cursor, where the struct for the cursor is essentially the same (a pointer and a length), but the pointer might point to anywhere in the middle of the array, and so that we knew never to free the memory through the cursor. Whereas the plan was to keep the byte buffer objects always pointing to the beginning of the allocated memory with the length field representing their full length.

I understand these concerns and decisions were not what motivated the initial creation of the byte_buffer struct, but those design decisions have already made their way into the byte_buf.h code in this package, specifically in those byte cursor advance functions.

I can see it either way here, the substrings shouldn't be freed themselves, but you might want to keep them as byte buffers and then only create byte cursor objects when you were reading through them. (We've been in the habit of *always* reading the data in our byte buffers with the byte cursor advance functions to ensure that we are not reading off the edges.)  Or you might want to call them byte cursors from the beginning to differentiate that they shouldn't be freed.

Mostly, I just want to make sure that we agree on the conventions so that we don't end up working at cross-purposes on the same code base.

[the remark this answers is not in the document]
david-koenig: Let me add that I won't hold up the pull request on this issue. Just want to make sure you think about it.

include/aws/common/byte_buf.h
```diff
@@ -61,6 +62,65 @@ static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t *c_str,
...
+*
+* The type that will be stored in output is struct aws_byte_cursor (you'll need this for the item size param).
+*
+* It is the user's responsibility to make sure the input buffer stays in memory long enough to use the results.
+*/
+AWS_COMMON_API int aws_string_split_on_char_n(struct aws_byte_buf *input_str, char split_on,
+    struct aws_array_list *output, size_t n);
+
+/**
+ * Copies from to to. If to is too small, AWS_ERROR_DEST_COPY_TOO_SMALL will be returned.
```
bdonlan: If the destination is too small, what is the state of the destination buffer after the call?

source/byte_buf.c
```diff
@@ -0,0 +1,108 @@
...
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/byte_buf.h>
+#include <assert.h>
+#include <stdarg.h>
+
+#ifdef _MSC_VER
+#pragma warning(disable:4204)
```
bdonlan: Document which warnings we're disabling, please.

source/byte_buf.c
```diff
@@ -0,0 +1,108 @@
...
+        struct aws_byte_buf *buffer = va_arg(ap, struct aws_byte_buf *);
+
+        if (buffer->len > dest_cursor.len) {
+            va_end(ap);
+            return aws_raise_error(AWS_ERROR_DEST_COPY_TOO_SMALL);
+        }
+
+        memcpy(dest_cursor.ptr, buffer->buffer, buffer->len);
+        aws_byte_cursor_advance(&dest_cursor, buffer->len);
+    }
```
bdonlan: Should we report the size of the data written somehow?

tests/string_test.c
```diff
@@ -0,0 +1,310 @@
...
+
+static int test_char_split_empty_fn(struct aws_allocator *allocator, void *ctx) {
+    const char str_to_split[] = "";
+
+    struct aws_byte_buf to_split = aws_byte_buf_from_literal(str_to_split);
+
+    struct aws_array_list output;
+    ASSERT_SUCCESS(aws_array_list_init_dynamic(&output, allocator, 4, sizeof(struct aws_byte_cursor)));
+    ASSERT_SUCCESS(aws_string_split_on_char(&to_split, ';', &output));
+    ASSERT_INT_EQUALS(0, aws_array_list_length(&output));
```
bdonlan: This behavior seems a bit inconsistent. I would argue that the number of elements in the output should be equal to the number of splits plus one. Having a weird edge case where it's zero seems like something that could trip someone up and lead to NPEs and similar.

tests/string_test.c
```diff
@@ -0,0 +1,310 @@
...
+    struct aws_byte_buf str1 = aws_byte_buf_from_literal("testa");
+    struct aws_byte_buf str2 = aws_byte_buf_from_literal(";testb");
+    struct aws_byte_buf str3 = aws_byte_buf_from_literal(";testc");
+
+
+    const char expected[] = "testa;testb;testc";
+
+    struct aws_byte_buf destination;
+    ASSERT_SUCCESS(aws_byte_buf_alloc(allocator, &destination, str1.len + str2.len + str3.len));
+    ASSERT_SUCCESS(aws_byte_buf_cat(&destination, 3, &str1, &str2, &str3));
```
bdonlan: Given that we have to manually compute the sum of the input buffers here, would it make more sense to move the allocation into cat to make it more convenient to use?

tests/string_test.c
```diff
@@ -0,0 +1,310 @@
...
+}
+
+AWS_TEST_CASE(test_buffer_cat_dest_too_small, test_buffer_cat_dest_too_small_fn)
+
+static int test_buffer_cpy_fn(struct aws_allocator *allocator, void *ctx) {
+    struct aws_byte_buf from = aws_byte_buf_from_literal("testa");
+    struct aws_byte_buf destination;
+
+    ASSERT_SUCCESS(aws_byte_buf_alloc(allocator, &destination, from.len));
+    ASSERT_SUCCESS(aws_byte_buf_copy(&destination, &from));
```
bdonlan: I'm also wondering if we should have the concept of a buffer with both a _maximum_ size and a _current_ size to make it easier to expose the result size from methods like this.

include/aws/common/byte_buf.h
JonathanHenson: It isn’t touched as the bounds check happens first.

source/byte_buf.c
JonathanHenson: I knew I forgot something.... I’ll add them. This is for non const declared initializers

source/byte_buf.c
JonathanHenson: I see two options.

1, add an output param
2, update the len value of dest.

Which do you prefer?

tests/string_test.c
JonathanHenson: I like this, this also removes a branch. Will do

tests/string_test.c
JonathanHenson: I’d rather not make the allocation decision for the user here. For example, I could already have preallocated buffers I want to use.

tests/string_test.c
JonathanHenson: I’m curious what value that would have since it’s always going to be from->len

JonathanHenson: done.

source/byte_buf.c
```diff
@@ -0,0 +1,113 @@
...
+    assert(output);
+    assert(output->item_size >= sizeof(struct aws_byte_cursor));
+
+    size_t max_splits = n > 0 ? n : SIZE_MAX;
+    size_t last_offset = 0, split_count = 0;
+    uint8_t *new_location = NULL;
+
+    struct aws_byte_cursor current_pos = aws_byte_cursor_from_buf(input_str);
+
+    while (split_count < max_splits &&
```
singku: I am refactoring the split function in CPP SDK, just checking out here.

You may want `split_count < max_splits - 1` here, or you'd better refactoring what `N `means int the comments.

If you don't hit the end when the while loop ended, the `split_count` will be `max_splits`, then you add the remaining to array. The results will have `n+1` splits(This is aligned with your test below, but the comments says, N is the max number of splits)

source/byte_buf.c
```diff
@@ -55,12 +56,11 @@ int aws_string_split_on_char_n(struct aws_byte_buf *input_str, char split_on,
         return aws_array_list_push_back(output, (const void *)&current_pos);
     }
 
-    if (input_str->buffer[input_str->len - 1] == split_on) {
-        struct aws_byte_cursor cursor = aws_byte_cursor_from_array(NULL, 0);
+    /* if we get here, we either hit the end and we need to add an empty split */
```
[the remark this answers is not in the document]
JonathanHenson: outdated comment. Will update

source/common.c
```diff
@@ -74,7 +74,7 @@ static struct aws_error_info errors[] = {
         AWS_DEFINE_ERROR_INFO(aws_error_condition_variable_failed, AWS_ERROR_COND_VARIABLE_ERROR_UNKNOWN, "condition variable unknown error.", AWS_LIB_NAME),
         AWS_DEFINE_ERROR_INFO(aws_error_clock_failure, AWS_ERROR_CLOCK_FAILURE, "clock get ticks operation failed", AWS_LIB_NAME),
         AWS_DEFINE_ERROR_INFO(aws_error_list_empty, AWS_ERROR_LIST_EMPTY, "empty list", AWS_LIB_NAME),
-        AWS_DEFINE_ERROR_INFO(aws_error_list_dest_copy_too_small, AWS_ERROR_LIST_DEST_COPY_TOO_SMALL, "destination of list copy is too small", AWS_LIB_NAME),
```
bdonlan: Is this error code too specific? Should we just use the short buffer error instead?

tests/string_test.c
```diff
@@ -0,0 +1,366 @@
...
+static int test_char_split_output_too_small_fn(struct aws_allocator *allocator, void *ctx) {
+    const char str_to_split[] = "testa;testb;testc;";
+
+    struct aws_byte_buf to_split = aws_byte_buf_from_literal(str_to_split);
+
+    struct aws_array_list output;
+    struct aws_byte_buf output_array[3] = {{0}};
+    ASSERT_SUCCESS(aws_array_list_init_static(&output, output_array, 3, sizeof(struct aws_byte_cursor)));
+    ASSERT_ERROR(AWS_ERROR_LIST_EXCEEDS_MAX_SIZE, aws_string_split_on_char(&to_split, ';', &output));
+
```
bdonlan: Should we assert on what is in the array now?

source/common.c
JonathanHenson: I dunno, I use it for array copies too. It doesn't matter to me.

tests/string_test.c
JonathanHenson: good idea.

PR #21, opened 2018-03-22
Add SPECTRE-resistant byte_buf bounds-check & advancement functions

bdonlan:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -98,10 +98,20 @@ static inline size_t aws_nospec_index(size_t index, size_t bound) {
      * that bit (and everything else), shift it over so it's the only bit in the
      * ones position, and multiply across the entire register.
      *
-     * Note that GCC is smart enough to optimize this multiply into an arithmetic
-     * right shift operation on x86.
+     * First, extract the (inverse) top bit and move it to the lowest bit.
+     * Because there's no standard SIZE_BIT in C99, we'll divide by a mask with
+     * just the top bit set instead.
      */
-    combined_mask = ((~combined_mask) >> (AWS_SIZE_BITS - 1)) * ~(size_t)0;
+    
+    combined_mask = (~combined_mask) / (~(size_t)0 - ((~(size_t)0) >> 1));
+
+    /* 
+     * Now multiply it to replicate it across all bits.
+     *
+     * Note that GCC is smart enough to optimize the divide-and-multiply into
+     * an arithmetic right shift operation on x86.
+     */
+    combined_mask = combined_mask * ~(size_t)0;
 
     return index & combined_mask;
 }
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -103,7 +103,7 @@ static inline size_t aws_nospec_index(size_t index, size_t bound) {
      * just the top bit set instead.
      */
     
-    combined_mask = (~combined_mask) / (~(size_t)0 - ((~(size_t)0) >> 1));
+    combined_mask = (~combined_mask) / (SIZE_MAX - (SIZE_MAX >> 1));
 
     /* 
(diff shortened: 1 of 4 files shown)


include/aws/common/byte_buf.h
```diff
@@ -35,5 +39,147 @@ static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t c_str[],
...
+    size_t combined_mask = negative_mask | toobig_mask;
+
+    /*
+     * combined_mask needs to have its sign bit OFF for us to be in range.
+     * Invert it and sign-extend it across the entire field.
+     *
+     * Note that this technically depends on UB - we need right-shift to
+     * perform sign extension on a negative input - but in practice most
+     * architectures and compilers do this. As an alternative we could do a
+     * multiply, if it were to be necessary.
```
marcomagdy: My vote is to never depend on UB because:
a) tools like valgrind and clang sanitizer will flag this (hopefully)
b) this lib will be distributed to AWS clients and we shouldn't make assumptions on what machines they will run on.

include/aws/common/byte_buf.h
```diff
@@ -35,5 +39,140 @@ static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t c_str[],
...
+    
+    combined_mask = (~combined_mask) / (~(size_t)0 - ((~(size_t)0) >> 1));
+
+    /* 
+     * Now multiply it to replicate it across all bits.
+     *
+     * Note that GCC is smart enough to optimize the divide-and-multiply into
+     * an arithmetic right shift operation on x86.
+     */
+    combined_mask = combined_mask * ~(size_t)0;
```
marcomagdy: Can we replace all the `~(size_t)0` with `SIZE_MAX`?

include/aws/common/byte_buf.h
```diff
@@ -35,5 +35,133 @@ static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t c_str[],
...
+     * SPECTRE mitigation - we compute a mask that will be zero if len < 0
+     * or len >= buf->len, and all-ones otherwise, and AND it into the index.
+     * It is critical that we avoid any branches in this logic.
+     */
+
+     /*
+      * We need to know the number of bits in a size_t for this. Unfortunately
+      * there doesn't seem to be a standard define for this...
+      */
+# if (UINTPTR_MAX >> 63 == 1)
```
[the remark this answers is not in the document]
bdonlan: Fixed in a later revision (and not quite technically correct)

include/aws/common/byte_buf.h
```diff
@@ -35,5 +39,140 @@ static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t c_str[],
...
+/**
+ * Tests if the given aws_byte_cursor has at least len bytes remaining. If so,
+ * *buf is advanced by len bytes (incrementing ->ptr and decrementing ->len),
+ * and an aws_byte_cursor referring to the first len bytes of the original *buf is
+ * returned. Otherwise, an aws_byte_cursor with ->ptr = NULL, ->len = 0 is returned.
+ *
+ * Note that if len is above (SIZE_MAX / 2), this function will also treat it as a buffer
+ * overflow, and return NULL without changing *buf.
+ */
+static inline struct aws_byte_cursor aws_byte_cursor_advance(struct aws_byte_cursor *cursor, size_t len) {
```
david-koenig: This construction is a little bit confusing to me, though admittedly my opinion may be swayed by the way you were using the cursor construct in the aws-encryption-sdk-c code. But if we've already checked that the original buffer we were inspecting had at least len bytes, I don't see what the advantage of returning a new struct aws_byte_cursor with len bytes is. The only thing that cursor can be used for is a redundant length check.

PR #22, opened 2018-04-02
aws_byte_cursor_from_array

david-koenig:
*Issue #, if available:*

*Description of changes:*

Add aws_byte_cursor_from_array

### DIFF
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -28,15 +28,15 @@ struct aws_byte_buf {
 };
 
 static inline struct aws_byte_buf aws_byte_buf_from_literal(const char *literal) {
-    return (struct aws_byte_buf){ .buffer = (uint8_t *)literal, .len = strlen(literal)};
+    return (struct aws_byte_buf){(uint8_t *)literal, strlen(literal)};
 }
 
 static inline struct aws_byte_buf aws_byte_buf_from_c_str(const char *c_str, size_t len) {
-    return (struct aws_byte_buf){.buffer = (uint8_t *)c_str, .len = len};
+    return (struct aws_byte_buf){(uint8_t *)c_str, len};
 }
 
 static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t *c_str, size_t len) {
-    return (struct aws_byte_buf){.buffer = (uint8_t *)c_str, .len = len};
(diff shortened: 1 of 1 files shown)


include/aws/common/byte_buf.h
```diff
@@ -51,6 +51,10 @@ static inline struct aws_byte_cursor aws_byte_cursor_from_buf(const struct aws_b
     return (struct aws_byte_cursor){.ptr = buf->buffer, .len = buf->len};
 }
 
+static inline struct aws_byte_cursor aws_byte_cursor_from_array(const uint8_t c_str[], size_t len) {
```
bdonlan: Style: I prefer `uint8_t *` in arguments; array syntax comes with some conceptual baggage that doesn't apply in argument lists.

david-koenig: I agree 100%. I always write it that way when doing my own code, but I had gone the other way to match the style of aws_byte_buf_from_array in the same file. I will change both of them to match the style you requested.

Incidentally, I prefer to keep spaces on both sides of the * as well when I am declaring a pointer, but it looks like I am losing that battle in this code.

include/aws/common/byte_buf.h
```diff
@@ -51,6 +51,10 @@ static inline struct aws_byte_cursor aws_byte_cursor_from_buf(const struct aws_b
     return (struct aws_byte_cursor){.ptr = buf->buffer, .len = buf->len};
 }
 
+static inline struct aws_byte_cursor aws_byte_cursor_from_array(const uint8_t *c_str, size_t len) {
+    return (struct aws_byte_cursor){.ptr = (uint8_t *)c_str, .len=len};
```
marcomagdy: I don't think designated initializers work on msvc (unless I'm missing a flag) https://godbolt.org/g/1a7h2e

(sorry i missed that in earlier reviews)

JonathanHenson:
They do work in vs2015 and later, but I’d recommend against having them in header files.

JonathanHenson:
No, there isn’t but I’ve been using them and building successfully for a long time. In cmake I do specify the language standard though.

marcomagdy:
I just checked it and the compiler flag is `/TC` that's what CMake uses when specifying the language standard.
Anyways, good to know for future godbolting.

PR #24, opened 2018-04-09
aws_task_scheduler_run_all() reentrancy bug fix, now scheduling a tas…

JonathanHenson:
…k for now inside a task will not run the task on that that call (necessary for fairness in aws-c-io). Also added timestamp output to make use in event loop simpler.


include/aws/common/task_scheduler.h
```diff
@@ -89,8 +89,16 @@ extern "C" {
 
     /**
      * Sequentially execute all tasks that are ready until either the queue is empty or no ready tasks are available.
+     * time before next task is the time in nanoseconds (based on the configured aws_task_scheduler_clock) before the
```
david-koenig: This sentence starts with a lower case letter. It is also unclear. There is an argument to the function called "next_task_time" but nothing called "time before next task". Is this an out of date description?

include/aws/testing/aws_test_harness.h
```diff
@@ -35,6 +36,8 @@ the AWS_UNSTABLE_TESTING_API compiler flag
 #define AWS_TESTING_REPORT_FD stderr
 #endif
 
+static struct aws_mutex mutex;
```
david-koenig: I think I would give this a more descriptive name than "mutex"

tests/task_scheduler_test.c
```diff
@@ -198,7 +198,61 @@ static int test_scheduler_rejects_xthread_access(struct aws_allocator *alloc, vo
...
+};
+
+static void reentrancy_fn(void *arg) {
+
+    struct task_scheduler_reentrancy_args *reentrancy_args = (struct task_scheduler_reentrancy_args *)arg;
+
+    if (reentrancy_args->reentrant_args) {
+        struct aws_task task;
+        task.fn = reentrancy_fn;
+        task.arg = reentrancy_args->reentrant_args;
```
david-koenig: The naming convention is a bit counterintuitive to me. struct task_scheduler_reentrancy_args has an element called reentrant_args. What does this mean, and can these names be made less cumbersome?

include/aws/common/task_scheduler.h
JonathanHenson: will fix.

include/aws/testing/aws_test_harness.h
JonathanHenson: sure, I'll rename it something like, allocator_lock.

tests/task_scheduler_test.c
JonathanHenson: it's a unit test, I don't really care that much, 

but it's self-referential so that's why it's named the same.

include/aws/testing/aws_test_harness.h
```diff
@@ -35,6 +36,8 @@ the AWS_UNSTABLE_TESTING_API compiler flag
 #define AWS_TESTING_REPORT_FD stderr
 #endif
 
+static struct aws_mutex allocator_lock;
```
bdonlan: Having static data members in a test header is pretty weird. Why isn't the malloc harness part of a library .c file?

include/aws/testing/aws_test_harness.h
```diff
@@ -322,6 +329,7 @@ static int aws_run_test_case(struct aws_test_harness *harness) {
 
 #define AWS_RUN_TEST_CASES(...)                                                                                        \
     struct aws_test_harness *tests[] = { __VA_ARGS__ };                                                                \
+    aws_mutex_init(&allocator_lock, aws_default_allocator());                                                          \
```
bdonlan: We should probably have a static initializer macro. Both pthreads and windows mutexes/SRW locks support this

include/aws/testing/aws_test_harness.h
```diff
@@ -331,21 +339,23 @@ static int aws_run_test_case(struct aws_test_harness *harness) {
...
-        return -1;                                                                                                     \
     }                                                                                                                  \
     else {                                                                                                             \
         for(size_t i = 0; i < test_count; ++i) {                                                                       \
             ret_val |= aws_run_test_case(tests[i]);                                                                    \
         }                                                                                                              \
     }                                                                                                                  \
                                                                                                                        \
+    aws_mutex_clean_up(&allocator_lock);                                                                               \
```
bdonlan: If we have a static initializer we can probably just leak the static mutex.

source/task_scheduler.c
```diff
@@ -122,14 +127,28 @@ int aws_task_scheduler_schedule_future(struct aws_task_scheduler *scheduler,
...
-        int err = aws_task_scheduler_next_task(scheduler, &task_to_run, 0);
-        if(AWS_OP_SUCCESS != err) {
-            return err;
+
+        if (get_next_task(scheduler, &task_to_run,  now, next_task_time)) {
+            int err_code = aws_last_error();
+            if (err_code == AWS_ERROR_TASK_SCHEDULER_NO_READY_TASKS  || err_code == AWS_ERROR_TASK_SCHEDULER_NO_TASKS) {
+                return AWS_OP_SUCCESS;
+            }
+            else if (err_code == AWS_ERROR_TASK_SCHEDULER_NO_MORE_TASKS) {
```
bdonlan: It's kind of strange to report an error when you have a successful condition. Perhaps next_task_time should just be `(uint64_t)-1` in this case?

tests/task_scheduler_test.c
```diff
@@ -198,7 +198,63 @@ static int test_scheduler_rejects_xthread_access(struct aws_allocator *alloc, vo
...
+
+    ASSERT_SUCCESS(aws_task_scheduler_schedule_now(&scheduler, &task), "Failed while scheduling task.");
+
+    ASSERT_SUCCESS(aws_task_scheduler_run_all(&scheduler, NULL), "Executing tasks failed.");
+
+    ASSERT_TRUE(task1_args.executed, "Task 1 should have been executed but was not.");
+    ASSERT_FALSE(task2_args.executed, "Task 2 should not have been executed but was.");
+
+    ASSERT_SUCCESS(aws_task_scheduler_run_all(&scheduler, NULL), "Executing tasks failed.");
+    ASSERT_TRUE(task2_args.executed, "Task 2 should not have been executed but was.");
```
bdonlan: Wrong assert string.

As an aside, I'd suggest leaving off assert strings when they state the same thing as the actual assert expression - the default assert message will print e.g. `Expected condition to be true: task2_args.executed [task_scheduler.test.c:250]`, which says the same thing as your assert string would say (if it wasn't wrong due to a copy-paste error), but with less typing and no chance of being inconsistent with the actual condition being tested.

include/aws/testing/aws_test_harness.h
JonathanHenson: Having to link in a test lib will end up being a pain for users. Just look at google test and their move to amalgamated header file.

Plus it ends up being static to the main test entry point which I think is fine.

include/aws/testing/aws_test_harness.h
JonathanHenson: Yeah that allocator isn’t actually necessary anyways IIRC

include/aws/testing/aws_test_harness.h
bdonlan: It could be just part of the main aws-c-commons library.

tests/task_scheduler_test.c
JonathanHenson: I'm going to add a coding style/standards section to the README. 

But the guidance I think I'd like to go with is,

Use the strings when you want to give reasoning for why x is a failure condition, but not just to print out the failure condition.

include/aws/common/mutex.h
```diff
@@ -36,10 +35,12 @@ struct aws_mutex {
 extern "C" {
 #endif
 
+AWS_COMMON_API struct aws_mutex aws_mutex_static_init (void);
```
bdonlan: This is not ideal. That still requires code to be run to initialize the mutex, and it assumes that mutexes are moveable structures (which is probably the case for static initialized things, but might not always be true).

Instead do something like

```
#ifdef _WIN32
#define AWS_MUTEX_INIT SRWLOCK_INIT
#else
#define AWS_MUTEX_INIT PTHREAD_MUTEX_INITIALIZER
#endif
```

This makes it a compile time constant, so mutexes can be defined at the file level pre-initialized.

include/aws/common/task_scheduler.h
```diff
@@ -66,10 +66,9 @@ extern "C" {
      * Gets the next task that is ready for execution. Tasks in the future will not be returned, 
      * even if they are the highest priority. next_run_time, if it is not null, and there are any tasks scheduled
      * will be set to the timestamp for the highest priority task. This is a useful hint for setting timeouts on
-     * event loops, or thread sleeps. If no tasks are scheduled, this value will be set to 0.
+     * event loops, or thread sleeps. If no tasks are scheduled, this value will be set to -0.
```
bdonlan: `-0`?

include/aws/common/mutex.h
JonathanHenson: This breaks the abstraction/ isolation of interface from implementation. I think that’s less ideal.

JonathanHenson: Actually scratch that, this abstraction is already broken for mutex. Will update

PR #25, opened 2018-04-10
Allocate an error code range at 0x4000

bdonlan:
*Issue #, if available:*

*Description of changes:*


source/error.c
```diff
@@ -26,7 +26,7 @@ static AWS_THREAD_LOCAL aws_error_handler thread_handler = NULL;
 AWS_THREAD_LOCAL void *thread_handler_context = NULL;
 
 #ifndef AWS_MAX_ERROR_SLOTS
-#define AWS_MAX_ERROR_SLOTS 16
+#define AWS_MAX_ERROR_SLOTS 32
```
JonathanHenson: I'm not sure why this needs to grow any. We only have 5 planned libs we're releasing for the SDK, plus yours that makes 6.

PR #27, opened 2018-04-11
host to network and network to host byte order functions for 16/32/64…

david-koenig:
… bit ints

*Issue #, if available:*

*Description of changes:*
Our own implementation of common byte order swapping functions. This has the advantage of not relying on any preprocessor symbols such as ```__LITTLE_ENDIAN__``` to detect endianness, which are inconsistent across compilers and systems. If compiled at -O1 or higher, it does not introduce any branching.

### DIFF
--- a/include/aws/common/byte_order.h
+++ b/include/aws/common/byte_order.h
@@ -18,85 +18,62 @@
 
 #include <stdint.h>
 
+#ifdef _WIN32
+#include <winsock2.h>
+#else
+#include <netinet/in.h>
+#endif /*_WIN32 */
+
+
 /**
  * Returns 1 if machine is big endian, 0 if little endian.
  * If you compile with even -O1 optimization, this check is completely optimized out at compile
- * time and code which calls "if (is_big_endian())" will do the right thing without branching.
+ * time and code which calls "if (aws_is_big_endian())" will do the right thing without branching.
  */
-static inline int is_big_endian() {
+static inline int aws_is_big_endian() {
     const uint16_t z = 0x100;
     return *(const uint8_t*)&z;
 }
 
 /**
- * Swaps bytes of 64 bit integer.
- */
-static inline uint64_t byteswap64(uint64_t x) { return
-        (x << 56) | (x >> 56) |
-        ((x & 0x0000ff00ULL) << 40) | ((x >> 40) & 0x0000ff00ULL) |
-        ((x & 0x00ff0000ULL) << 24) | ((x >> 24) & 0x00ff0000ULL) |
-        ((x & 0xff000000ULL) <<  8) | ((x >>  8) & 0xff000000ULL);
-}
-
-/**
- * Swaps bytes of 32 bit integer.
- */
-static inline uint32_t byteswap32(uint32_t x) { return
-        (x << 24) | (x >> 24) |
-        ((x & 0xff00) << 8 ) | ((x >> 8) & 0xff00);
-}
-
-/**
- * Swaps bytes of 16 bit integer.
- */
-static inline uint16_t byteswap16(uint16_t x) { return (x << 8) | (x >> 8); }
-
-/**
  * Convert 64 bit integer from host to network byte order.
  */
-static inline uint64_t hton64(uint64_t x) {
-    if (is_big_endian()) return x;
-    return byteswap64(x);
+static inline uint64_t aws_hton64(uint64_t x) {
+    if (aws_is_big_endian()) return x;
+#if defined(__x86_64__) && (defined(__GNUC__) || defined(__clang__))
+    uint64_t v;
+    __asm__("bswap %q0" : "=r" (v) : "0" (x));
+    return v;
+#else
+    uint32_t low = x;
+    uint32_t high = x >> 32;
+    return ((uint64_t)htonl(low)) << 32 | htonl(high);
+#endif
 }
 
 /**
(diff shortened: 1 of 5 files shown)


JonathanHenson:
How is this different from:

https://github.com/awslabs/aws-c-common/blob/master/include/aws/common/encoding.h#L86

?

david-koenig:
@JonathanHenson and I talked about this, and decided this is a better approach and to make the encoding code use this. I've updated the PR with the change.

tests/byte_order_test.c
```diff
@@ -0,0 +1,54 @@
...
+    uint64_t x = 0x1122334455667788ULL;
+    uint64_t rev_x = 0x8877665544332211ULL;
+
+    uint32_t y = 0xaabbccdd;
+    uint32_t rev_y = 0xddccbbaa;
+
+    uint16_t z = 0xeeff;
+    uint16_t rev_z = 0xffee;
+
+    // Since I know of no more reliable way to detect endianness
```
JonathanHenson: oh yeah, forgot to mention, no C++ comments.

only 
/* */ style.

david-koenig:
I see no mention of winsock in the cmake files

include/aws/common/byte_order.h
```diff
@@ -0,0 +1,102 @@
...
+        (x << 56) | (x >> 56) |
+        ((x & 0x0000ff00ULL) << 40) | ((x >> 40) & 0x0000ff00ULL) |
+        ((x & 0x00ff0000ULL) << 24) | ((x >> 24) & 0x00ff0000ULL) |
+        ((x & 0xff000000ULL) <<  8) | ((x >>  8) & 0xff000000ULL);
+}
+
+/**
+ * Swaps bytes of 32 bit integer.
+ */
+static inline uint32_t byteswap32(uint32_t x) { return
```
bdonlan: This will be significantly slower than `htonl`/`ntohl` on x86. x86 has specific assembly instructions for byteswapping 16-bit (`ROR $8, %ax`), 32-bit (`BSWAP %eax`) and 64-bit (`BSWAP %rax`) registers.

include/aws/common/byte_order.h
```diff
@@ -0,0 +1,102 @@
...
+static inline int is_big_endian() {
+    const uint16_t z = 0x100;
+    return *(const uint8_t*)&z;
+}
+
+/**
+ * Swaps bytes of 64 bit integer.
+ */
+static inline uint64_t byteswap64(uint64_t x) { return
+        (x << 56) | (x >> 56) |
```
bdonlan: I would expect using htonl for 32-bit chunks to be faster on x86-32 as it can at least use the 32-bit bswap instruction to avoid most of the masking.

Ideal would be to use the 64-bit bswap routines in (linux-only) `endian.h`. This would need a CMake-time probe however as they're unavailable on OS X/windows. If unavailable, we'd need to fall back to inline assembler or using 32-bit htonl/ntohl.

include/aws/common/byte_order.h
```diff
@@ -0,0 +1,102 @@
...
+ */
+static inline uint32_t byteswap32(uint32_t x) { return
+        (x << 24) | (x >> 24) |
+        ((x & 0xff00) << 8 ) | ((x >> 8) & 0xff00);
+}
+
+/**
+ * Swaps bytes of 16 bit integer.
+ */
+static inline uint16_t byteswap16(uint16_t x) { return (x << 8) | (x >> 8); }
```
bdonlan: `is_big_endian() ? x : htons(x)` to use bit rotates instead of masking.

include/aws/common/byte_order.h
```diff
@@ -0,0 +1,102 @@
...
+ */
+static inline uint64_t ntoh64(uint64_t x) {
+    if (is_big_endian()) return x;
+    return byteswap64(x);
+}
+
+/**
+ * Convert 32 bit integer from host to network byte order.
+ */
+static inline uint32_t hton32(uint32_t x) {
```
bdonlan: Use htonl instead internally. This will provide a more efficient, machine-specific implementation.

tests/byte_order_test.c
bdonlan: How about we have some explicit byte arrays and test ntoh results?

```
uint8_t val[] = { 0x01, 0x02, 0x03, 0x04 };
uint32_t val2;
memcpy(&val2, val, 4);

ASSERT_UINT_EQUALS(ntoh32(val2), 0x01020304);
```

include/aws/common/byte_order.h
```diff
@@ -0,0 +1,79 @@
...
+    uint32_t low = x;
+    uint32_t high = x >> 32;
+    return ((uint64_t)htonl(low)) << 32 | htonl(high);
+#endif
+}
+
+/**
+ * Convert 64 bit integer from network to host byte order.
+ */
+static inline uint64_t aws_ntoh64(uint64_t x) { return aws_hton64(x); }
```
bdonlan: It's worth noting that this only works on big endian and little endian systems - on mixed endian systems it might not work properly. Probably not worth worrying too much about though.

david-koenig:
@JonathanHenson Functions now have aws_ prefix and all comments are C-style. I did not remove the winsock dependency because after incorporating @bdonlan's feedback, this retained the dependency on htonl/htons family of functions. @bdonlan has approved this PR and I believe it is now ready to be merged.

PR #28, opened 2018-04-11
Initial reference, and coding guidlines etc....

JonathanHenson:
This doesn't need to be perfect or all inclusive, it's just a start, we can add as we go. But at least take some time to raise things you disagree with or want reworded for more clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,86 @@ Core c99 package for AWS SDK for C. Includes cross-platform primitives, configur
 
 This library is licensed under the Apache 2.0 License. 
 
-## Error Codes
+## Usage
+### Building
+aws-c-common uses CMake for setting up build environments. This library has now non-kernel dependencies so the build is quite
+simple. 
+
+For example:
+
+    git clone git@github.com:awslabs/aws-c-common.git aws-c-common
+    mkdir aws-c-common-build
+    cd aws-c-common-build
+    cmake ../aws-c-common
+    make -j 12
+    make test
+    sudo make install
+    
+Keep in mind that CMake supports multiple build systems, so for each platform you can pass your own build system
+as the `-G` option. For example:        
+
+    cmake -GNinja ../aws-c-common
+    ninja build
+    ninja test
+    sudo ninja install
+    
+Or on windows,     
+
+    cmake -G "Visual Studio 14 2015 Win64" ../aws-c-common
+    msbuild.exe ALL_BUILD.vcproj    
+    
+### API style and conventions
+Every API has a specific set of styles and conventions. We'll outline them here. These conventions are followed in every
+library in the AWS C SDK ecosystem.
+
+#### Error handling
+Every function that returns an `int` type, returns `AWS_OP_SUCCESS` ( 0 ) or `AWS_OP_ERR` (-1) on failure. To retrieve
+the error code, use the function `aws_last_error()`. Each error code also has a corresponding error string that can be 
+accessed via the `aws_error_str()` function.
+
+In addition, you can install both a global and a thread local error handler by using the `aws_set_global_error_handler_fn()`
+and `aws_set_thread_local_error_handler_fn()` functions.
+
+All error functions are in the `include/aws/common/error.h` header file.
+
+#### Naming
+Any function that allocates and initializes an object will be suffixed with `new`. Similarly, these objects will always 
+have a corresponding function with a `destroy` suffix. The `new` suffixed functions will return `NULL` on failure. 
+To respond to the error, call `aws_last_error()`.
+
+Any function that simply initializes an object will be suffixed with `init`. These objects will have a corresponding 
+`clean_up` function. In these cases, you are responsible for making the decisions for how your object is allocated.
+
+## Contributing
+
+If you are contributing to this code-base, first off, THANK YOU!. There are a few things to keep in mind to minimize the 
+pull request turn around time.
+
+### Coding "guidelines"
+These "guidelines" are followed in every library in the AWS C SDK ecosystem.
+
+#### Memory Management
+* All APIs that need to be able to allocate memory, must take an instance of `aws_allocator` and use that. No `malloc()` or
+`free()` calls should be made directly.
+* If an API does not allocate the memory, it does not free it. All allocations and deallocations should take place at the same level.
+For example, if a user allocates memory, the user is responsible for freeing it. There will inevitably be a few exceptions to this
+rule, but they will need significant justification to make it through the code-review.
+* All functions that allocate memory must raise an `AWS_ERROR_OOM` error code upon allocation failures. If it is a `new()` function
+it should return NULL. If it is an `init()` function, it should return `AWS_OP_ERR`.
+
+#### Threading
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,7 +94,52 @@ have pre-slotted error codes for each library. The currently allocated error ran
...
-| [0x1000, 0x2000) |aws-c-amazon-flow | 
+| [0x1000, 0x2000) |aws-c-eventstream | 
 | [0x2000, 0x2800) | (reserved for future project) |
 
-Each library should begin its error codes at the beginning of its range and follow in sequence (don't skip codes).
+Each library should begin its error codes at the beginning of its range and follow in sequence (don't skip codes). Upon
+adding an AWS maintained library, an error code range must be approved and added to the above table.
+
+### Testing
+PRs without unit tests will not be acknowledged. A couple of tips:
```
bdonlan: We should still acknowledge and request changes...

README.md
```diff
@@ -15,7 +94,52 @@ have pre-slotted error codes for each library. The currently allocated error ran
...
+We provide a test harness for writing unit tests. This includes an allocator that will fail your test if you have any 
+memory leaks, as well as some `ASSERT` macros. To write a test:
+
+* Create a *.c test file in the tests directory of the project. 
+* Implement one or more tests with the signature `int test_case_name(struct aws_allocator *, void *ctx)`
+* Use the `AWS_TEST_CASE` macro to declare the test.
+* Include your test in the `tests/main.c` file.
+* Include yur test in the `tests/CMakeLists.txt` file. 
+
+### Coding Style
```
bdonlan: Do we have any automated coding style enforcement?

README.md
JonathanHenson: by contract with the user, no, but to be nice and productive (probably).

We'll probably say something like, let me know when you've tested this and I'll have a look

README.md
JonathanHenson: I'll add cpp style to the travis runs at some point.

cppcheck also does some of this. Marco added this recently but we backed it out because we had so much in flight. We can add it back shortly, since the churn has turned down a bit.

README.md
JonathanHenson: maybe the wording should be something like:

"We will not evaluate a PR without unit tests or if the unit tests aren't passing other than to ask you to please write unit tests."

bdonlan: Even that feels a bit strong as an absolute. What if someone wants review of an API sketch before they implement it? What if it's a PR that just does formatting changes, or improves efficiency but doesn't change behavior and thus doesn't have any unit tests?

It's better to just say that we have a high bar for test coverage, and PRs fixing bugs or introducing new functionality need to have tests before they will be accepted.

JonathanHenson: sure, will update

README.md
```diff
@@ -15,7 +94,53 @@ have pre-slotted error codes for each library. The currently allocated error ran
...
+* Indent is 4 spaces
+* K & R style for braces
+* Space after if, before the `(`
+* Avoid C99 features in header files
+* Avoid C++ style comments e.g. `//`
+* All public API functions need C++ guards and Windows dll semantics
+* Use Unix line endings
+* SNAKE_UPPER_CASE constants, macros, and enum members.
+* snake_lower_case everything else.
+* do not use typedef struct idiom.
```
david-koenig: I don't understand why we put the kibosh on this. If we're typedeffing function pointers, it also seems to make sense to me to typedef struct pointers when we want users to handle them but we don't want users to be able to declare their own instances of the structs. (i.e., opaque pointers with forward declaration of structs, so that we can change their implementation later)

README.md
```diff
@@ -15,7 +94,53 @@ have pre-slotted error codes for each library. The currently allocated error ran
...
+* typedef function pointer definitions if you intend to expose them to the user
+* typedef enums
+* every source and header file must have a copyright header (The standard AWS one for apache 2).
+* Use standard include guards (e.g. #IFNDEF HEADER_NAME #define HEADER_NAME etc...)
```
david-koenig: I'm personally in favor of using "#pragma once" because otherwise we can easily run into issues such as the current include/aws/common/byte_buf.h file having an include guard of AWS_STRING_H due to changed filenames. If we don't maintain those include guards well, we can easily end up breaking things due to include guard naming collisions. Also "#pragma once" is one line of code, instead of three for the include guard.

README.md
JonathanHenson: typedef structs is broadly seen as problematic. I'm not opposed to it in the ideal, but I do want to standardize on one or the other. 

I also don't mind amending this standard when we have a good use case for using it.

README.md
JonathanHenson: #pragma once expects a C99 language standard, which we don't want to expose in a header file, for non C99 programs.

I also "prefer" it, but I'd rather not hamstring a user without a really good reason. (ergonomics isn't a good enough reason).

README.md
david-koenig: I've never heard of typedef struct as being problematic, and I'm wondering what the case for that is. That being said, what I am advocating for is slightly different, i.e., typedeffing a pointer to a struct, not the struct itself. I can understand the argument that typedef struct might obscure the actual type detail to a programmer who needs to know it, but the case I am thinking about is specifically the one where we want our users *not* to know the details of the struct, i.e., when we return or modify a struct in certain API calls that we give them, and they have to in turn send those structs to future API calls. In those cases, we want to reserve the possibility to change our implementation of the struct later, so we don't want them ever dereferencing the pointer and doing something to the memory underneath. By typedeffing, we encourage our users not to think about the object as a pointer type, but simply as a thing they need to pass around from one function to another as a whole. So I might amend the wording to "Only use typedef for struct pointers, not structs themselves, when they are being forward declared in end user header files."

README.md
david-koenig: It works fine for me to compile code with #pragma once and -std=c89 using gcc/clang on both my Linux dev desktop and my Mac laptop. I even used a redundant include to make sure that #pragma once was doing what it was supposed to be doing:

(18-04-13 16:34:14) <0> [~/sandbox]
 % cat once.h
#pragma once

struct foo {
    int x, y;
};

(18-04-13 16:34:24) <0> [~/sandbox]
 % cat once.c
#include "once.h"
#include "once.h"
#include <stdio.h>

int main() {
    struct foo f = {5, 7};
    printf("%i\t%i\n", f.x, f.y);
    return 0;
}

(18-04-13 16:34:25) <0> [~/sandbox]
 % gcc -Wall -Werror -pedantic -std=c89 -o once once.c

(18-04-13 16:34:31) <0> [~/sandbox]
 % ./once
5	7

That was on Linux. Results on Mac were the same. If there really is a portability problem with #pragma once on other systems/compilers, then I agree with your reasoning.

JonathanHenson: the main goal here is that we want to support this from any compiler (even ancient ones using ANSI C), adding #pragma once is similar to using C99 syntax in a header file.

README.md
JonathanHenson: http://yarchive.net/comp/linux/typedefs.html

PR #29, opened 2018-04-13
Condition variable

JonathanHenson:
Condition variable implementation and tests.


include/aws/common/condition_variable.h
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Notifies/Wakes all waiting threads.
+ */
+AWS_COMMON_API int aws_condition_variable_notify_all (struct aws_condition_variable *condition_variable);
+
+/**
+ * Waits the calling thread on a notification from another thread.
+ */
+AWS_COMMON_API int aws_condition_variable_wait (struct aws_condition_variable *condition_variable, struct aws_mutex *mutex);
```
marcomagdy: You need another parameter for a predicate to handle the spurious wakeup case.

include/aws/common/condition_variable.h
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Waits the calling thread on a notification from another thread. Times out after time_to_wait. time_to_wait is in nanoseconds.
+ */
+AWS_COMMON_API int aws_condition_variable_wait_until (struct aws_condition_variable *condition_variable, struct aws_mutex *mutex,
```
marcomagdy: **Please** rename this `aws_condition_variable_wait_for ` instead of `aws_condition_variable_wait_until ` to avoid confusion with the C++ api.

source/posix/condition_variable.c
```diff
@@ -0,0 +1,90 @@
...
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/condition_variable.h>
+#include <aws/common/mutex.h>
+
+#include <errno.h>
+
+#define NANOS_PER_SEC 1000000000
```
marcomagdy: should be marked with `ull` suffix otherwise division will be handles as integral and we'll have to handle the overflow/underflow case.

source/windows/condition_variable.c
```diff
@@ -0,0 +1,56 @@
...
+#include <aws/common/condition_variable.h>
+#include <aws/common/mutex.h>
+
+int aws_condition_variable_init(struct aws_condition_variable *condition_variable) {
+    InitializeConditionVariable(&condition_variable->condition_handle);
+    return AWS_OP_SUCCESS;
+}
+
+void aws_condition_variable_clean_up(struct aws_condition_variable *condition_variable) {
+    (void *)&condition_variable;
```
marcomagdy: I'd remove this line and remove the name of the parameter to avoid unused variable warnings.

source/windows/condition_variable.c
```diff
@@ -0,0 +1,56 @@
...
+    if (SleepConditionVariableSRW(&condition_variable->condition_handle, &mutex->mutex_handle, INFINITE, 0)) {
+        return AWS_OP_SUCCESS;
+    }
+
+    return aws_raise_error(AWS_ERROR_COND_VARIABLE_ERROR_UNKNOWN);
+}
+
+int aws_condition_variable_wait_until(struct aws_condition_variable *condition_variable, struct aws_mutex *mutex,
+    uint64_t time_to_wait) {
+    DWORD time_ms = (DWORD)(time_to_wait / 1000000);
```
marcomagdy: add unsigned suffix here too.
#define MILLIS_PER_SEC?

source/windows/condition_variable.c
JonathanHenson: that's not allowed in C for function definitions (only declarations).

source/posix/condition_variable.c
JonathanHenson: good catch.

include/aws/common/condition_variable.h
marcomagdy: >When using condition variables there is always a Boolean predicate involving shared variables associated with each condition wait that is true if the thread should proceed. Spurious wakeups from the pthread_cond_timedwait() or pthread_cond_wait() functions may occur. Since the return from pthread_cond_timedwait() or pthread_cond_wait() does not imply anything about the value of this predicate, the predicate should be re-evaluated upon such return.

https://linux.die.net/man/3/pthread_cond_wait

JonathanHenson: damn it. I didn't want to do this. But you're correct. Will update.

include/aws/common/condition_variable.h
JonathanHenson: good call, will update.

source/posix/condition_variable.c
JonathanHenson: ah, nevermind, these are supposed to be signed AND these are not subject to overflow due to usage.

PR #32, opened 2018-05-15
change type names that start with 'aws_common_' prefix to 'aws_'

david-koenig:
also make map arg const in hash table find function


include/aws/common/common.h
```diff
@@ -95,7 +95,7 @@ AWS_COMMON_API void aws_load_error_strings(void);
 #define AWS_OP_SUCCESS 0
 #define AWS_OP_ERR -1
 
-typedef enum aws_common_error {
```
JonathanHenson: These probably need to stay prefixed with common

PR #38, opened 2018-05-30
Add byte cursor read/write utility functions.

bdonlan:
*Issue #, if available:*

*Description of changes:*


bdonlan:
I consider the error returns here to be too much of a 'normal' condition to be an error code that requires the whole thread-local-and-callback error return framework; bringing that framework in has significant overhead.

Also, the test list was updated in the prior commit; the subsequent change is to add in a missing file. I'll squash them when I rebase on master.

bdonlan:
Yes, that is what the change uses now.

PR #40, opened 2018-05-31
Byte buf allocators

david-koenig:
*Issue #, if available:*

*Description of changes:*


david-koenig:
The only reason that this is failing the windows build at the moment is the scheduler_reentrant_safe test. This is not related to this code change and is currently being worked on by another developer.

include/aws/common/byte_buf.h
```diff
@@ -20,44 +20,55 @@
...
 static inline struct aws_byte_buf aws_byte_buf_from_literal(const char *literal) {
     struct aws_byte_buf buf;
     buf.buffer = (uint8_t *)literal;
     buf.len = strlen(literal);
+    buf.allocator = NULL;
     return buf;
 }
 
-static inline struct aws_byte_buf aws_byte_buf_from_c_str(const char *c_str, size_t len) {
+static inline struct aws_byte_buf aws_byte_buf_from_c_str(struct aws_allocator * allocator, const char *c_str, size_t len) {
```
JonathanHenson: why would I want an allocator here?

include/aws/common/byte_buf.h
```diff
@@ -20,44 +20,55 @@
...
+    buf.allocator = allocator;
     return buf;
 }
 
-static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t *c_str, size_t len) {
+static inline struct aws_byte_buf aws_byte_buf_from_array(struct aws_allocator * allocator, const uint8_t *c_str, size_t len) {
     struct aws_byte_buf buf;
     buf.buffer = (uint8_t *)c_str;
     buf.len = len;
+    buf.allocator = allocator;
```
JonathanHenson: same here, this is memory we don't own.

include/aws/common/byte_buf.h
david-koenig: I thought there might be a use case where we had already allocated the array or string and then wanted to hand over that ownership to the byte buffer and I figured it was easiest to give the caller the option of whether to pass in an allocator for deallocation or to pass in NULL. But I'll change this and the other one to just not take the allocator argument and set the allocator to NULL.

PR #42, opened 2018-06-01
Fixed bug in task scheduler caused by our code being so damned fast, …

JonathanHenson:
…it outran the clock.

### DIFF
--- a/source/task_scheduler.c
+++ b/source/task_scheduler.c
@@ -37,6 +37,7 @@ int aws_task_scheduler_init(struct aws_task_scheduler *scheduler, struct aws_all
         aws_task_scheduler_clock clock) {
     scheduler->alloc = alloc;
     scheduler->clock = clock;
+    scheduler->latest_run_time = 0;
     return aws_priority_queue_dynamic_init(&scheduler->queue, alloc, DEFAULT_QUEUE_SIZE, sizeof(struct task_container),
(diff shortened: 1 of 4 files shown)


source/task_scheduler.c
```diff
@@ -122,6 +124,16 @@ int aws_task_scheduler_schedule_future(struct aws_task_scheduler *scheduler,
...
     struct task_container container;
 
+    /* this serves the purpose of handling reentrant scheduling while the tasks are being run.
+       if the clock tick is on the same nanosecond (with microsecond precision) as the run was kicked
+       off, it will still be run. As a result, if time_to_run is before or on the same clock tick, increment it
+       by one nanosecond to avoid the issue. However this does assume that the clock ticks are at least greater than 0. Seeing as how
+       boot times take longer than 1 nanosecond and a system clock will always be greater than 0, I think this is a safe assumption 
+       in a real world scenario. */
+    if (AWS_UNLIKELY(time_to_run <= scheduler->latest_run_time)) {
+        time_to_run = scheduler->latest_run_time + 1;
```
graebm: Later in this review you had to change a unit test to start its clock at 1 instead of 0.

If this variable was `scheduler->run_time_min` instead of `scheduler->latest_run_time`, then we could allow tasks at time 0.

PR #45, opened 2018-06-01
Updated test to make it not fail due to fast clocks on windows.

JonathanHenson:

### DIFF
--- a/tests/task_scheduler_test.c
+++ b/tests/task_scheduler_test.c
@@ -206,9 +206,6 @@ static int test_scheduler_reentrant_safe(struct aws_allocator *alloc, void *ctx)
 
     ASSERT_FALSE(task2_args.executed);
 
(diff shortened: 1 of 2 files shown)


tests/task_scheduler_test.c
```diff
@@ -206,6 +206,9 @@ static int test_scheduler_reentrant_safe(struct aws_allocator *alloc, void *ctx)
 
     ASSERT_FALSE(task2_args.executed);
 
+    /* these clocks are REALLLY fast on windows depending on the hardware,
```
graebm: you could allow subsequent runs in the same nanosecond to succeed by modifying `_run_all()` like so:
```
... get time
if (now < scheduler->min_run_at) {
    now = scheduler->min_run_at;
}
... proceed to set scheduler->min_run_at = now + 1;
```

PR #53, opened 2018-06-04
Renamed bytebuf size to capacity, fixed some unit tests that were inc…

JonathanHenson:
…orrectly using len.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/encoding.c
+++ b/source/encoding.c
@@ -78,6 +78,7 @@ int aws_hex_encode(const struct aws_byte_buf *AWS_RESTRICT to_encode, struct aws
     }
 
     output->buffer[written] = '\0';
+    output->len = encoded_len;
 
     return AWS_OP_SUCCESS;
 }
@@ -153,6 +154,8 @@ int aws_hex_decode(const struct aws_byte_buf *AWS_RESTRICT to_decode, struct aws
         value |= low_value;
         output->buffer[written++] = value;
     }
+
+    output->len = decoded_length;
 
     return AWS_OP_SUCCESS;
 }
--- a/source/encoding.c
+++ b/source/encoding.c
@@ -66,7 +66,7 @@ int aws_hex_encode(const struct aws_byte_buf *AWS_RESTRICT to_encode, struct aws
         return AWS_OP_ERR;
     }
 
-    if (AWS_UNLIKELY(output->len < encoded_len)) {
+    if (AWS_UNLIKELY(output->capacity < encoded_len)) {
         return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
     }
 
@@ -124,7 +124,7 @@ int aws_hex_decode(const struct aws_byte_buf *AWS_RESTRICT to_decode, struct aws
         return aws_raise_error(AWS_ERROR_OVERFLOW_DETECTED);
     }
 
-    if (AWS_UNLIKELY(output->len < decoded_length)) {
+    if (AWS_UNLIKELY(output->capacity < decoded_length)) {
         return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
     }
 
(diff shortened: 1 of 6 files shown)


source/encoding.c
```diff
@@ -66,7 +66,7 @@ int aws_hex_encode(const struct aws_byte_buf *AWS_RESTRICT to_encode, struct aws
         return AWS_OP_ERR;
     }
 
-    if (AWS_UNLIKELY(output->len < encoded_len)) {
+    if (AWS_UNLIKELY(output->capacity < encoded_len)) {
         return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
     }
```
david-koenig: Just before this function succeeds it should set output->len = encoded_len

source/encoding.c
```diff
@@ -124,7 +124,7 @@ int aws_hex_decode(const struct aws_byte_buf *AWS_RESTRICT to_decode, struct aws
         return aws_raise_error(AWS_ERROR_OVERFLOW_DETECTED);
     }
 
-    if (AWS_UNLIKELY(output->len < decoded_length)) {
+    if (AWS_UNLIKELY(output->capacity < decoded_length)) {
         return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
     }
```
david-koenig: this one should set output->len = decoded_length before returning successfully.

tests/encoding_test.c
```diff
@@ -27,33 +27,35 @@ static int run_hex_encoding_test_case(struct aws_allocator *alloc, const char *t
...
-    memset(allocation, 0xdd, output_size + 2);
-
     struct aws_byte_buf to_encode = aws_byte_buf_from_c_str(test_str, test_str_size - 1);
-    struct aws_byte_buf output = aws_byte_buf_from_c_str(output_str, output_size);
+    struct aws_byte_buf output;
+    ASSERT_SUCCESS(aws_byte_buf_init(alloc, &output, output_size + 2));
+    memset(output.buffer, 0xdd, output_size + 2);
+
+    output.capacity = output_size;
+    output.buffer += 1;
```
david-koenig: I get what you're doing in terms of putting a byte in front of and in back of the buffer to make sure that you're not writing past the end of it, but it's a very unnatural thing to change the buffer pointer, which forces you to change it back before freeing the memory, as you do in line 57.

While the test looks correct, I often think of test code as illustrating to my users how to use the functions, and this is something we wouldn't want users ever doing. I might do this in a slightly different way which is closer to what you did before, but using the new buffer library functions:

struct aws_byte_buf allocation;
ASSERT_SUCCESS(aws_byte_buf_init(alloc, allocation, output_size + 2));
struct aws_byte_buf output = aws_byte_buf_from_array(allocation.buffer + 1, output);
Then you can run the encode/decode functions on output only, but verify that there is no corruption on allocation.buffer[0] and allocation.buffer[allocation.capacity - 1].

I would also then add a check that output.len has been set properly by the encode/decode functions. 

Furthermore, you can run aws_byte_buf_clean_up on both buffers, since the function knows not to deallocate anything on output. This increases the chances that if someone messes up any of the aws_byte_buf_init, aws_byte_buf_from_array, and aws_byte_buf_clean_up functions in the future, that it will most likely break the tests.

tests/encoding_test.c
```diff
@@ -28,34 +28,38 @@ static int run_hex_encoding_test_case(struct aws_allocator *alloc, const char *t
     ASSERT_INT_EQUALS(expected_size, output_size, "Output size on string should be %d", expected_size);
 
     struct aws_byte_buf to_encode = aws_byte_buf_from_c_str(test_str, test_str_size - 1);
-    struct aws_byte_buf output;
-    ASSERT_SUCCESS(aws_byte_buf_init(alloc, &output, output_size + 2));
-    memset(output.buffer, 0xdd, output_size + 2);
 
-    output.capacity = output_size;
-    output.buffer += 1;
+    struct aws_byte_buf allocation;
```
david-koenig: These changes look great. I would make the same changes to run_base64_encoding_test_case and it will be good to go.

JonathanHenson: ok

PR #57, opened 2018-06-08
Add placeholder logging functions

marcomagdy:
*Description of changes:*

Placeholder logging functions, so we can start using them in PRs. They're no-op as of now.

### DIFF
--- a/include/aws/common/logger.h
+++ b/include/aws/common/logger.h
@@ -14,33 +14,35 @@
 * express or implied. See the License for the specific language governing
 * permissions and limitations under the License.
 */
+#include <aws/common/common.h>
+#include <stdarg.h>
 
 enum aws_log_level
 {
-    aws_off = 0,
-    aws_fatal = 1,
-    aws_error = 2,
-    aws_warn = 3,
-    aws_info = 4,
-    aws_debug = 5,
-    aws_trace = 6
+    AWS_LOG_LEVEL_OFF = 0,
+    AWS_LOG_LEVEL_FATAL = 1,
+    AWS_LOG_LEVEL_ERROR = 2,
+    AWS_LOG_LEVEL_WARN = 3,
+    AWS_LOG_LEVEL_INFO = 4,
+    AWS_LOG_LEVEL_DEBUG = 5,
+    AWS_LOG_LEVEL_TRACE = 6
 };
 
 static inline const char *aws_get_log_level_name(enum aws_log_level level) {
     switch(level) {
-        case aws_off:
+        case AWS_LOG_LEVEL_OFF:
             return "OFF";
-        case aws_fatal:
+        case AWS_LOG_LEVEL_FATAL:
             return "FATAL";
-        case aws_error:
+        case AWS_LOG_LEVEL_ERROR:
             return "ERROR";
-        case aws_warn:
+        case AWS_LOG_LEVEL_WARN:
             return "WARN";
-        case aws_info:
+        case AWS_LOG_LEVEL_INFO:
             return "INFO";
-        case aws_debug:
+        case AWS_LOG_LEVEL_DEBUG:
             return "DEBUG";
-        case aws_trace:
+        case AWS_LOG_LEVEL_TRACE:
             return "TRACE";
         default:
             return "UNKNOWN";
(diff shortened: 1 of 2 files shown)


include/aws/common/logger.h
```diff
@@ -0,0 +1,102 @@
...
+*
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+enum aws_log_level
+{
+    aws_off = 0,
```
JonathanHenson: Should be something like AWS_LOG_LEVEL_OFF etc...

include/aws/common/logger.h
```diff
@@ -0,0 +1,102 @@
...
+        case aws_debug:
+            return "DEBUG";
+        case aws_trace:
+            return "TRACE";
+        default:
+            return "UNKNOWN";
+    }
+}
+
+static inline void aws_log(enum aws_log_level level, struct aws_allocator *alloc, const char *tag, const char *msg,
```
JonathanHenson: I think what we want is some macros that do nothing at the moment... something like

AWS_LOG_F(level, format, ...)
AWS_LOG(level, str)

And just have it do nothing for the moment.
This way it compiles out to nothing in the future when logging is disabled.

marcomagdy: I didn't want to use macros because:
1. I didn't want to conflict with the C++ SDK ones (couldn't come up with a better name quickly)
1. You can still get them to compile out to nothing if you define them with empty body.

include/aws/common/logger.h
marcomagdy: OK

include/aws/common/logger.h
```diff
@@ -0,0 +1,104 @@
...
+    AWS_LOG_LEVEL_OFF = 0,
+    AWS_LOG_LEVEL_FATAL = 1,
+    AWS_LOG_LEVEL_ERROR = 2,
+    AWS_LOG_LEVEL_WARN = 3,
+    AWS_LOG_LEVEL_INFO = 4,
+    AWS_LOG_LEVEL_DEBUG = 5,
+    AWS_LOG_LEVEL_TRACE = 6
+};
+
+static inline const char *aws_get_log_level_name(enum aws_log_level level) {
```
graebm: Trivial, but I don't think `static` does anything here. This is a header file

include/aws/common/logger.h
```diff
@@ -0,0 +1,104 @@
...
+            return "TRACE";
+        default:
+            return "UNKNOWN";
+    }
+}
+
+static inline void aws_log(enum aws_log_level level, struct aws_allocator *alloc, const char *tag, const char *msg,
+        va_list args) {
+
+    /* TODO: implement this */
```
graebm: Don't we want the user setting the function which ultimately handles logging? Kind of like how the user ultimately sets an error handler via `aws_set_global_error_handler_fn()`

Then we wouldn't need to pass an allocator around with every logging call

include/aws/common/logger.h
bdonlan: If static is not provided, the symbol will be exported (at least at the library level) and will clash with any other .c files which included this header. In addition, it'll be generated even if it's not used.

That said this should probably not be static or inline.

include/aws/common/logger.h
bdonlan: I agree that setting the allocator should be part of global logging configuration. It may well differ from the allocator meant to be used from the calling context (e.g. what if the logging framework wants to ship the message off to a logging writeback thread, but the calling code wants to destroy its non-thread-safe allocator?)

tests/main.c
```diff
@@ -20,6 +20,7 @@
 #pragma warning(disable:4204)
 #endif
 
+#include <aws/common/logger.h>
```
bdonlan: ?

include/aws/common/logger.h
```diff
@@ -0,0 +1,104 @@
...
+        case AWS_LOG_LEVEL_DEBUG:
+            return "DEBUG";
+        case AWS_LOG_LEVEL_TRACE:
+            return "TRACE";
+        default:
+            return "UNKNOWN";
+    }
+}
+
+static inline void aws_log(enum aws_log_level level, struct aws_allocator *alloc, const char *tag, const char *msg,
```
bdonlan: The logging logic should be out-of-line for now.

Later on we might want to change it to do the bare minimum needed to check if the logging category is enabled before passing code to an out-of-line function that actually does the logging.

graebm:
Logger is being implemented here: https://github.com/awslabs/aws-c-common/pull/131

RandyGaul:
Well, the memory pool is implemented there. The logger is, for now, being implemented in the `mini_logger` branch: https://github.com/awslabs/aws-c-common/tree/mini_logger

PR #58, opened 2018-06-08
New aws_string type for immutable strings which can be used as keys and values for hash table

david-koenig:
-Reorganized byte buffer, byte cursor, split, and string tests
-Updated wording of split functions so as not to refer to strings
-Change hash table to use different destroy functions for key and value
-Helper functions for string use in hash table
-Unit tests for new string and hash table functionalities

*Issue #, if available:*

*Description of changes:*
Implementation of hash table was not usable with binary strings because if they are not null terminated we have no way to determine the length once we retrieve the pointer. The byte buffer implementation was not suitable for this either, because the struct itself is generally declared on the stack and only its data bytes are allocated.

This implementation of a string type solves these problems and also includes added functionality and tests in the hash table code to allow for hash tables with strings as keys and/or values.

There is also a significant amount of cleanup and reorganization of the test code, as our tests for byte buffers, cursors, and split functions had become scattered and the file names were not always accurate.


include/aws/common/string.h
```diff
@@ -0,0 +1,101 @@
...
+ * held in dynamic memory, such as when held by a struct aws_hash_table. The data bytes
+ * themselves are always held in contiguous memory immediately after the end of the
+ * struct aws_string, and the memory for both the header and the data bytes
+ * is allocated together.
+ *
+ * Use the aws_string_bytes function to access the data bytes.
+ */
+struct aws_string {
+    struct aws_allocator * allocator;
+    size_t len;
```
JonathanHenson: Why not just put a uint8_t* at the end of this instead of the point arithmetic?

david-koenig: This is used in a situation where we need to run malloc (or aws_mem_acquire) on the struct itself. If we allocated memory to a pointer within the struct, we would have to do two allocations, one for the struct and one for the pointer. The way it is coded now is an optimization to require only one malloc for the entire string (header struct and data bytes).

I suppose it would be possible to include a pointer to the data bytes and always set that pointer to the value of aws_string_bytes() upon creation. That would have the advantage of making the implementation opaque to the users and more like the byte buffer object, though it seems like an unnecessary use of memory, since the pointer will always do the same thing as the aws_string_bytes() function. (Not only that, but I would argue that using the pointer to always go the same number of bytes forward rather than the function is less safe from a security perspective.)

@bdonlan and I had a detailed conversation about this, and one insight he had was that in cases where we wanted to be able to reallocate the data bytes, it makes sense to allocate the struct and the data bytes separately, so that we can reallocate the memory within the string (such as when appending bytes beyond the current capacity) without losing the pointer to the original struct, which can be important when that struct is held in other container objects by that pointer. The optimization we made of allocating them all together makes sense in the context of an immutable string, and this use case of an immutable string is what is most useful to us in a hash table.

david-koenig:
I made a very minor update to the AWS_STATIC_STRING_FROM_LITERAL macro, but otherwise this PR is still unchanged from before. Unfortunately, I mishandled my git branches, so the commits that I had originally all squashed together on this got added to the git history making it looks much hairier than it is.

include/aws/common/hash_table.h
```diff
@@ -104,17 +103,19 @@ extern "C" {
...
+     * Either or both may be NULL if a callback is not desired in this case.
      */
 
     AWS_COMMON_API int aws_hash_table_init(
         struct aws_hash_table *map,
         struct aws_allocator *alloc, size_t size,
         aws_hash_fn_t hash_fn,
         aws_equals_fn_t equals_fn,
-        aws_hash_element_destroy_t destroy_fn);
+        aws_hash_element_destroy_t destroy_key_fn,
```
bdonlan: While we're making changes, do we need a context parameter on the destroy fn? This may not be very useful in pure-C applications, but it might be useful when doing language bindings to pass a pointer to the foreign function to be invoked via a shim.

include/aws/common/string.h
```diff
@@ -0,0 +1,100 @@
...
+};
+
+static inline const uint8_t * aws_string_bytes(const struct aws_string * hdr) {
+    return (const uint8_t *)(hdr + 1);
+}
+
+static inline const struct aws_string * aws_string_from_literal_new(struct aws_allocator * allocator, const char * literal) {
+    size_t len = strlen(literal);
+    struct aws_string * hdr = aws_mem_acquire(allocator, sizeof(struct aws_string) + len);
+    if (!hdr) {aws_raise_error(AWS_ERROR_OOM); return NULL;}
```
bdonlan: Do we have a style guide? I prefer not putting multi-statement blocks on a single line, for readability.

include/aws/common/string.h
```diff
@@ -0,0 +1,100 @@
...
+    hdr->allocator = allocator;
+    hdr->len = len;
+    memcpy((void *)aws_string_bytes(hdr), literal, len);
```
bdonlan: This loses the null terminator. Should we preserve it (without reporting it as part of the length) - or, more in general, guarantee that `aws_string_bytes(hdr)[hdr->len] == 0`, as with C++ `std::string`? Doing so makes it much easier to interoperate with traditional C APIs.

include/aws/common/string.h
```diff
@@ -0,0 +1,100 @@
...
+static inline const struct aws_string * aws_string_from_array_new(struct aws_allocator * allocator, const uint8_t * bytes, size_t len) {
+    struct aws_string * hdr = aws_mem_acquire(allocator, sizeof(struct aws_string) + len);
+    if (!hdr) {aws_raise_error(AWS_ERROR_OOM); return NULL;}
+    hdr->allocator = allocator;
+    hdr->len = len;
+    memcpy((void *)aws_string_bytes(hdr), bytes, len);
+    return hdr;
+}
+
+static inline const struct aws_string * aws_string_from_c_str_new(struct aws_allocator * allocator, const char * bytes, size_t len) {
```
bdonlan: This is identical to aws_string_from_array_new only with a `const char *`. Consider using `const void *` instead; or at least delegate from one to the other.

include/aws/common/string.h
```diff
@@ -0,0 +1,100 @@
...
+ */
+void aws_string_destroy(void * buf);
+
+/**
+ * Defines a (static const struct aws_string *) with name specified in first argument
+ * that points to constant memory and has data bytes containing the string literal in the second argument.
+ */
+#define AWS_STATIC_STRING_FROM_LITERAL(name, literal)                              \
+    static const struct {struct aws_string hdr; uint8_t data[sizeof(literal)];}    \
+        _ ## name ## _s = {                                                        \
```
bdonlan: Identifiers starting with `_` are reserved for the compiler and C library and cannot be (portably) used.

include/aws/common/string.h
```diff
@@ -0,0 +1,100 @@
...
+    hdr->len = len;
+    memcpy((void *)aws_string_bytes(hdr), bytes, len);
+    return hdr;
+}
+
+/* Takes a void * so it can be used as a destructor function for struct aws_hash_table.
+ * That is also why it is not inlined.
+ */
+void aws_string_destroy(void * buf);
+
```
bdonlan: An `aws_string_clone` would probably be useful too.

include/aws/common/string.h
```diff
@@ -0,0 +1,100 @@
...
+    struct aws_string * hdr = aws_mem_acquire(allocator, sizeof(struct aws_string) + len);
+    if (!hdr) {aws_raise_error(AWS_ERROR_OOM); return NULL;}
+    hdr->allocator = allocator;
+    hdr->len = len;
+    memcpy((void *)aws_string_bytes(hdr), bytes, len);
+    return hdr;
+}
+
+/* Takes a void * so it can be used as a destructor function for struct aws_hash_table.
+ * That is also why it is not inlined.
```
bdonlan: In general I'd argue that allocate/deallocate functions are high enough overhead (due to the allocator call) that inlining is not necessary in the first place for the allocate side either.

source/hash_table.c
```diff
@@ -556,12 +566,31 @@ int aws_hash_table_foreach(struct aws_hash_table *map,
 
 void aws_hash_table_clear(struct aws_hash_table *map) {
     struct hash_table_state *state = map->pImpl;
-    if (state->destroy_fn) {
+    if (state->destroy_key_fn) {
```
bdonlan: There's a lot of duplicate code in this loop. I'd do an `if (state->destroy_key_fn || state->destroy_value_fn)` at the top level and test for each individual one on the inner loop. Yes, it's very slightly slower, but in practice the call to whichever destructor exists will be much higher overhead than the inner loop guards.

tests/hash_table_test.c
```diff
@@ -60,7 +60,132 @@ static int test_hash_table_put_get_fn(struct aws_allocator *alloc, void *ctx) {
...
+    const struct aws_string * key_2 = aws_string_from_literal_new(alloc, "Over many a quaint and curious");
+    AWS_STATIC_STRING_FROM_LITERAL(val_2, "volume of forgotten lore--");
+    const struct aws_string * key_3 = aws_string_from_literal_new(alloc, "While I nodded, nearly napping,");
+    AWS_STATIC_STRING_FROM_LITERAL(val_3, "suddenly there came a tapping,");
+
+    const struct aws_string * dyn_keys[] = {key_1, key_2, key_3};
+    const struct aws_string * static_vals[] = {val_1, val_2, val_3};
+
+    int ret = aws_hash_table_init(&hash_table, alloc, 10, aws_hash_string, aws_string_eq,
+                                  aws_string_destroy, NULL); /* destroy keys not values */
```
bdonlan: I assume this is implicitly tested via the debug allocator

tests/string_test.c
```diff
@@ -1,382 +1,63 @@
...
+                             "Data bytes should have been set correctly.");
 
-    ASSERT_ERROR(AWS_ERROR_DEST_COPY_TOO_SMALL, aws_byte_buf_cat(&destination, 3, &str1, &str2, &str3));
-
-    aws_byte_buf_clean_up(&destination);
-
-    return 0;
-}
+    /* Test: string creation works. */
+    const struct aws_string * test_string_2 = aws_string_from_c_str_new(alloc, "foofaraw", 8);
```
bdonlan: The `c_str` part would, to me, imply a nul-terminated string, not a length delimited argument.

include/aws/common/string.h
JonathanHenson: agreed, we do have a style guide but this isn't explicitly called out. It's part of the reference to "k&r" style.

include/aws/common/hash_table.h
david-koenig: I do not have any insight on this.

include/aws/common/string.h
david-koenig: Will fix.

include/aws/common/string.h
david-koenig: I feel strongly that we should not report the null terminator as part of the length. Not including it was consistent with our approach on byte buffers.

I don't have a huge objection to including it, but something to keep in mind is that the aws_string object is designed to hold either textual or binary data. In fact, the ability to use it to hold binary data is much more the justification for the need for the new data type in the context of a hash table. If we were only using null-terminated strings in the hash table, the pointers we pass in could be just pointers to the C-strings and we could always use strlen to determine their length. But this data type allows us to have keys and values within the hash table where the null byte is an ordinary character within the string. If we are treating the null byte as an ordinary character within the string, it feels a little unnatural to stick an extra null byte on the end. Furthermore, in that case getting a c-string back from the aws_string object would not give the expected result, because that c-string would end right before the first null byte.

JonathanHenson: My argument any api we want to use that only takes a cstr argument without length is an API we shouldn’t be using in the first place.

include/aws/common/string.h
david-koenig: This was not my design decision. The different constructors for aws_string were the same as what we already had for aws_byte_buf. I have noticed that we have warnings on strongly enough that trying to call the function that expects unsigned characters on arrays of signed characters, or the reverse, causes an error, which is probably why this was built this way in the first place. I don't have a super strong feeling about what the right set of constructor functions is, but it does seem to me that they should be the same for both aws_byte_buf and aws_string. Opinions @JonathanHenson ?

include/aws/common/string.h
david-koenig: Will fix.

include/aws/common/string.h
david-koenig: I don't feel strongly about this. Happy to change it. Does that mean that we also want the aws_byte_buf_init function not to be inlined?

include/aws/common/string.h
david-koenig: Not sure I agree. Since these are supposed to be immutable, I'd rather refer to the same string in memory than make multiple copies of the same immutable data. This seems like something we should wait to add until we have a need for it.

include/aws/common/string.h
david-koenig: On further thought, I think this is a better naming convention:

aws_string_from_literal_new -> aws_string_from_c_str_new
aws_string_from_array_new -> aws_string_from_uint8_arr_new
aws_string_from_c_str_new-> aws_string_from_int8_arr_new

To me "c_str" means null terminated string, which is how we have been using the term literal before.

david-koenig: Also, I will do a test of whether making the bytes argument in the last two functions a void * obviates the need for two separate functions for the signed and unsigned cases.

david-koenig: Yes, it looks like @bdonlan was right, and we can avoid making separate functions for unsigned and signed character arrays by making the arguments (void *).

New constructors where we don't specify length but get it from strlen will be aws_string_from_c_str_new and aws_byte_buf_from_c_str. (What used to be _literal.) New constructors where we do specify length will be aws_string_from_array_new and aws_byte_buf_from_array. These changes will be included on the next PR update.

source/hash_table.c
david-koenig: I agree that making this change can make for slightly simpler code, but in my opinion the efficiency gain is important here. Hash tables can get very large. Doing redundant checks of the same flag every iteration through is wasteful. Since this is a general purpose library that anyone who is using hash tables of any size while working with AWS in C might use, it makes sense to me to do a best possible implementation rather than sacrifice quality for a little bit of extra clarity.

Furthermore, this function isn't that long or hard to read as is. I got rid of the blank lines, so now the entire function is under 30 lines. I also looked it over with another coworker and on her advice added a few extra comments to improve the clarity.

tests/string_test.c
david-koenig: I agree with this. As I mentioned above, this was not my naming convention. I have changed the naming convention to match what you want here.

tests/hash_table_test.c
david-koenig: Not quite sure what you mean here. I haven't used the debug allocator, but I could look into adding some more tests with it. I did run all of these tests through valgrind to make sure that all allocated memory was being freed.

david-koenig: I understand now that the memory tracking allocator is running by default on the tests. I verified that if the destructor function was not set properly on the init, this test will not pass because of memory not being freed. So yes, the test framework allocator is working properly to help us on this test.

include/aws/common/byte_buf.h
```diff
@@ -103,47 +110,21 @@ AWS_COMMON_API int aws_byte_buf_cat(struct aws_byte_buf *dest, size_t number_of_
...
-    buf.len = len;
-    buf.capacity = len;
+    buf.len = strlen(c_str);
+    buf.capacity = buf.len;
     buf.allocator = NULL;
     return buf;
 }
 
-static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t *c_str, size_t len) {
+static inline struct aws_byte_buf aws_byte_buf_from_array(const void * bytes, size_t len) {
```
JonathanHenson: why are we making this void * now, uint8_t * seems appropriate given 'byte_buf'. I also like forcing users to do an explicit cast from const char * to uint8_t * as a "I agree that I want you to do thiat and I approve" mechanism.

source/common.c
```diff
@@ -40,7 +40,10 @@ struct aws_allocator *aws_default_allocator() {
 }
 
 void *aws_mem_acquire(struct aws_allocator *allocator, size_t size) {
-    return allocator->mem_acquire(allocator, size);
+    void * mem = allocator->mem_acquire(allocator, size);
+    if (!mem)
```
JonathanHenson: braces or same line

JonathanHenson: If we're going to make this change, it needs to be done EVERYWHERE we allocate memory

source/string.c
```diff
@@ -0,0 +1,41 @@
...
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+#include <aws/common/string.h>
+
+const struct aws_string * aws_string_from_c_str_new(struct aws_allocator * allocator, const char * c_str) {
```
JonathanHenson: nit:

spacing between * and function name.

source/common.c
david-koenig: I thought I did handle this everywhere we allocate memory. If I missed something, please point me to it.

include/aws/common/byte_buf.h
david-koenig: I don't feel strongly about that either way, but I do feel strongly that we shouldn't have two separate functions for the (uint8_t *) case and the (char *) case. I was following @bdonlan's opinion that we should make this a void * to make it more easily usable no matter whether the original array was signed or unsigned. Your opinion is the opposite. While I do find having to do the explicit casts here a bit annoying, I agree that uint8_t * is more descriptive, so assuming mine is the tiebreaking vote, I will change it back to that, but I will not include a separate function for (char *) arrays.

source/string.c
david-koenig: Is this specified in a coding standard? I prefer spaces on both sides of the *. If not that, I definitely prefer omitting the space between the star and the type, since this is meant to be read (const struct aws_string *) as a return type. Omitting the space between the function name and the star would be what I prefer the least. But this is not a hill I will die on.

include/aws/common/string.h
bdonlan: I don't think we should report it as part of the length. I'm simply considering that someone may reasonably want to take the value in an aws_string and pass it to, say, `printf` or `fopen` or something. Currently they'd need to malloc a new buffer, memcpy the data into it, set the nul-terminator, and then free the buffer. Or, at the cost of a single byte, we can make that problem go away.

david-koenig: That's reasonable. I'll make the change to keep the extra null byte.

include/aws/common/byte_buf.h
JonathanHenson: I agree about not having both char * and uint8_t *

source/common.c
JonathanHenson: you're right I thought we also had it in the priority queue and scheduler implementations.

source/string.c
JonathanHenson: It doesn't really matter to me other than consistency. It's the defacto standard at the moment, but it's not in the README at the moment.

Issue #60, opened 2018-06-18
Potential Memory safety issue in Array List - set_at function


danielsn:
In aws_array_list_set_at, the necessary_size (set at line 238) is list->item_size too small (an off-by-one error).
The code in array_list_bug_set.c is an example of how using aws_array_list_set_at on one list (list0) can result in the written value being read out at the beginning of another list (list1).
The example only works if the index provided to aws_array_list_set_at is such that list0->data + index * list->item_size is the same as list1->data, which depends on how memory allocation works out. When I compile with clang on my machine, the example always works, but I don’t know how portable it is.

```
#include <aws/common/array_list.h>
#include <aws/common/common.h>
#include <stdlib.h>

int main (void) {
    struct aws_allocator *allocator = aws_default_allocator();
    struct aws_array_list *list0 = aws_mem_acquire(allocator, sizeof(*list0));
    struct aws_array_list *list1 = aws_mem_acquire(allocator, sizeof(*list1));
    char val0 = 20;
    char val1 = 0;
    if (aws_array_list_init_dynamic(list0, allocator, 4, sizeof(char)) == AWS_OP_SUCCESS) {
        if (aws_array_list_init_dynamic(list1, allocator, 4, sizeof(char)) == AWS_OP_SUCCESS) {
            aws_array_list_set_at(list1, (void *)&val1, 4);
            aws_array_list_set_at(list0, (void *)&val0, 16); // write 20 into list0
            // this behavior is only visible when memory allocation is such that the
            // following condition holds
            if (list0->data + 16 == list1->data) {
                aws_array_list_get_at(list1, &val1, 0);
                printf("%d\n", val1); // can read out 20 from list1
            }
        }
    }
    aws_array_list_clean_up(list0);
    aws_mem_release(allocator, list0);
    aws_array_list_clean_up(list1);
    aws_mem_release(allocator, list1);
    return 0;
}
```

Issue #59, opened 2018-06-18
Potential memory safety issue in Array List implemetation


danielsn:
In aws_array_list_init_dynamic, list->current_size is not set.
The code in array_list_bug_init.c segfaults when list->current_size is nonzero.
I never set the value of list->current_size to be a nonzero value, so whether or not there is a segfault is nondeterministic (since list->current_size might happen to be zero).

```
#include <aws/common/array_list.h>
#include <aws/common/common.h>

int main (void) {
    struct aws_allocator *allocator = aws_default_allocator();
    // dynamic aws_array_list
    struct aws_array_list *list = aws_mem_acquire(allocator, sizeof(*list));
    // current_size field is possibly nonzero
    // initial_item_allocation is zero
    if (aws_array_list_init_dynamic(list, allocator, 0, sizeof(char)) == AWS_OP_SUCCESS) {
        char val = 2;
        // segfaults if current_size field is nonzero
        aws_array_list_set_at(list, (void *)&val, 1);
    }
    aws_array_list_clean_up(list);
    aws_mem_release(allocator, list);
    return 0;
}
```

PR #62, opened 2018-06-18
Addresses issues #59 and #60. array list also now supports zero alloc…

JonathanHenson:
…ation on setup.

*Issue #, if available:*

*Description of changes:*


source/array_list.c
```diff
@@ -24,6 +24,8 @@ int aws_array_list_init_dynamic(struct aws_array_list *list,
     size_t allocation_size = initial_item_allocation * item_size;
     list->data = NULL;
     list->item_size = item_size;
+    list->current_size = 0;
+    list->length = 0;
```
graebm: a few lines below here length is set to 0. You can delete that line since you already set it above. GitHub won't let me highlight the offending line :|

source/array_list.c
```diff
@@ -137,7 +139,7 @@ int aws_array_list_pop_back(struct aws_array_list *list) {
 }
 
 void aws_array_list_clear(struct aws_array_list *list) {
-    if (list->length > 0) {
+    if (list->data && list->length > 0) {
```
graebm: you don't have to change this line, length can't be >0 when data is NULL

source/array_list.c
```diff
@@ -167,6 +171,7 @@ int aws_array_list_shrink_to_fit(struct aws_array_list *list) {
 
 int aws_array_list_copy(const struct aws_array_list *from, struct aws_array_list *to) {
     assert(from->item_size == to->item_size);
+    assert(from->data);
```
graebm: We can handle it if you modify that if-statements that follow, like so:
```
    ...
    if (to->current_size >= copy_size) {
        if (copy_size > 0) {
            memcpy(to->data, from->data, copy_size);
        }
        ...
```
This is also more efficient than the current code, which needlessly re-allocates when `to->len` is less than capacity

source/array_list.c
```diff
@@ -146,6 +148,8 @@ void aws_array_list_clear(struct aws_array_list *list) {
 }
 
 int aws_array_list_shrink_to_fit(struct aws_array_list *list) {
+    assert(list->data);
+
```
graebm: Why assert? we can handle it.

```
        if (ideal_size < list->current_size) {
            ... everything before here is same as before
            void *raw_data = NULL;
            if (ideal_size > 0) {
                raw_data = aws_mem_acquire(list->alloc, ideal_size);
                if (!raw_data) {
                    return AWS_OP_ERR;
                }

                memcpy(raw_data, list->data, ideal_size);
            }
            ... everything after here is same as before
            aws_mem_release(list->alloc, list->data);
            list->data = raw_data;
            list->current_size = ideal_size;
        }
```

source/array_list.c
JonathanHenson: done

source/array_list.c
JonathanHenson: good idea. Done.

source/array_list.c
JonathanHenson: another good catch! done.

PR #66, opened 2018-06-19
RFC: Remove main.c and have cmake generate the test entry point

bdonlan:
It's proven a bit error-prone to have tests defined in multiple places. This
change uses cmake's create_test_sourcelist to generate an entry point for the
tests; now, the only way to run a test is if you actually define it in
CMakeLists (otherwise it's completely unreachable), so if you've run the test
_at all_ it should be in the 'make test' testlist.

Note: This is built on top of the realloc PR for now to minimize merge conflict pain.


graebm:
I like this a lot. Much less copy/paste. Also it was weird to have a .c file include other .c files.

ColdenCullen:
Closing, we took #124 instead.

PR #70, opened 2018-06-22
Clang tidy fixes

ColdenCullen:
- Add .clang-tidy file with default settings
  - Disable valist.Uninitialized, it triggers a false positive.
- Add asserts to aws_array_list_mem_swap to validate parameters.
- task_schedule.c: Add asserts to validate task functions before call.
    This is required because the analyzer doesn't know that
    aws_raise_error() always returns 1, so it can't make
    branch flow assumptions.
- Remove some trailing whitespaces.

After this PR is merged, running cmake with -DCMAKE_C_CLANG_TIDY="clang-tidy" will run clang-tidy as part of the build, and fail on any new errors.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,7 +1,6 @@
 ---
 Checks:          'clang-diagnostic-*,clang-analyzer-*,-clang-analyzer-valist.Uninitialized'
 WarningsAsErrors: '*'
-FormatStyle:     none
 CheckOptions:
   - key:             readability-braces-around-statements.ShortStatementLines
     value:           '1'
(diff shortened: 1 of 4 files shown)


ColdenCullen:
Yeah I'll give it a shot. I'm not sure if it's installed by default, but we can find out!

ColdenCullen:
So it looks like we can't, because the build is running on Ubunutu 14.04 LTS, which only supports CMake 3.5. CMAKE_C_CLANG_TIDY was added in 3.6.

JonathanHenson:
this isn't useful if it isn't in our CI, you'll need to find another way to run it.

JonathanHenson:
also, it also doesn't work on Ubuntu 16.04

PR #73, opened 2018-06-22
fix -pedantic

graebm:
There are smarter ways to do this in cmake (using CHECK_CXX_COMPILER_FLAG), but I'm just going to keep this simple for now.


graebm:
btw, if anyone is curious -pedantic and -Wpedantic are aliases for the same thing.

The -W version was added late in the gcc4 lifecycle, which is why it's working on some people's machines with gcc4, but not others.

PR #76, opened 2018-06-26
Add a utility function to convert an aws_string to an aws_byte_buf

bdonlan:
*Issue #, if available:*

*Description of changes:*


include/aws/common/string.h
```diff
@@ -74,6 +74,19 @@ extern "C" {
...
+/**
+ * Creates an aws_byte_buf that references the data in an aws_string.
+ * The aws_byte_buf will not be associated with the aws_string's allocator;
+ * destroying the byte_buf is therefore a no-op.
+ *
+ * Calling code should take care not to modify the data in the returned byte
+ * buf; code generally assumes that the aws_string should be constant, and in
+ * many cases it may in fact be constant data.
+ */
+static inline struct aws_byte_buf aws_string_to_buf(const struct aws_string *s) {
```
david-koenig: I would make this return a const struct aws_byte_buf

bdonlan: That would have no effect. The constness would apply to the struct aws_byte_buf itself, which would be promptly copied to whatever the calling code assigns it to (which would lose the constness).

david-koenig: You are correct. I suppose an alternative approach would be to have a struct aws_const_byte_buf in which the pointer is a pointer to constant data.

It does seem hazardous to be handling byte buffers to what are supposed to be constant strings. What is the use case?

bdonlan:
WIthdrawing this in favor of creating an internal API elsewhere

PR #82, opened 2018-07-13
Add and install common CMake settings modules

ColdenCullen:
After this PR, these common CMake modules may be included like this:
```cmake
list(APPEND CMAKE_MODULE_PATH "${CMAKE_INSTALL_PREFIX}/lib/cmake")

include(AwsCFlags)
include(Sanitizers)
```

This reduces the ammout of copy-pasted code, enables more warnings by default, and makes it easier to enable Sanitizers.


cmake/AwsCFlags.cmake
```diff
@@ -0,0 +1,69 @@
...
+        set(EXPORT_DEFINE ${target})
+        string(TOUPPER ${EXPORT_DEFINE} EXPORT_DEFINE)
+        string(REPLACE "AWS-C" "AWS" EXPORT_DEFINE ${EXPORT_DEFINE})
+        string(REPLACE "-" "_" EXPORT_DEFINE ${EXPORT_DEFINE})
+
+        list(APPEND AWS_C_DEFINES -DUSE_IMPORT_EXPORT)
+        list(APPEND AWS_C_DEFINES -D${EXPORT_DEFINE}_EXPORTS)
+    endif()
+
+    set(AWS_C_PROPERTIES LINKER_LANGUAGE C C_STANDARD 99)
```
graebm: AWS_C_PROPERTIES isn't used anywhere but this one line

cmake/AwsCFlags.cmake
```diff
@@ -0,0 +1,69 @@
...
+    endif()
+
+    if(CMAKE_BUILD_TYPE STREQUAL "" OR CMAKE_BUILD_TYPE MATCHES Debug)
+        list(APPEND AWS_C_DEFINES -DDEBUG_BUILD)
+    endif()
+
+    if(BUILD_SHARED_LIBS AND WIN32)
+        set(EXPORT_DEFINE ${target})
+        string(TOUPPER ${EXPORT_DEFINE} EXPORT_DEFINE)
+        string(REPLACE "AWS-C" "AWS" EXPORT_DEFINE ${EXPORT_DEFINE})
```
graebm: This replace could false-positive and do something like this:
AWS-COMPRESSION" -> "AWSOMPRESSION"

maybe do the replace like:
string(REPLACE "^AWS-C-" "AWS-" ...

cmake/Sanitizers.cmake
```diff
@@ -0,0 +1,53 @@
...
+
+                # When testing for libfuzzer, if attempting to link there will be 2 mains
+                if(${sanitizer} STREQUAL "fuzzer")
+                    set(sanitizer_test_flag -fsanitize=fuzzer-no-link)
+                else()
+                    set(sanitizer_test_flag -fsanitize=${sanitizer})
+                endif()
+
+                set(sanitizer_variable HAS_SANITIZER_${sanitizer})
+                string(MAKE_C_IDENTIFIER ${sanitizer_variable} sanitizer_variable)
```
marcomagdy: I don't quite get the reason for this. Can you explain it via comments?

cmake/Sanitizers.cmake
```diff
@@ -0,0 +1,53 @@
...
+                # Need to set this here so that the flag is passed to the linker
+                set(CMAKE_REQUIRED_FLAGS ${sanitizer_test_flag})
+                check_c_compiler_flag(${sanitizer_test_flag} ${sanitizer_variable})
+                if(${${sanitizer_variable}})
+                    set(PRESENT_SANITIZERS "${PRESENT_SANITIZERS},${sanitizer}")
+                endif()
+            endforeach()
+
+            target_compile_options(${target} PRIVATE -fno-omit-frame-pointer -fsanitize=${PRESENT_SANITIZERS})
+            set_target_properties(${target} PROPERTIES LINK_FLAGS "-fno-omit-frame-pointer -fsanitize=${PRESENT_SANITIZERS}")
```
marcomagdy: You could do
```cmake
target_compile_options(${taget} PRIVATE $<$<CONFIG:Debug>:-fno-omit-frame-pointer -fsanitize=${PRESENT_SANITIZERS}>

set_target_properties(${target} PROPERTIES LINK_FLAGS_DEBUG "-fno-omit-frame-pointer -fsanitize=${PRESENT_SANITIZERS}")
```
And get rid of the if condition at the beginning.

cmake/Sanitizers.cmake
ColdenCullen: Sure!

PR #86, opened 2018-07-18
Read/Write Lock

ColdenCullen:

### DIFF
--- a/include/aws/common/rw_lock.h
+++ b/include/aws/common/rw_lock.h
@@ -1,0 +1,79 @@
+#ifndef AWS_COMMON_RW_LOCK_H
+#define AWS_COMMON_RW_LOCK_H
+
+/*
+* Copyright 2010-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License").
+* You may not use this file except in compliance with the License.
+* A copy of the License is located at
+*
+*  http://aws.amazon.com/apache2.0
+*
+* or in the "license" file accompanying this file. This file is distributed
+* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+* express or implied. See the License for the specific language governing
+* permissions and limitations under the License.
+*/
+
+#include <aws/common/common.h>
+#ifdef _WIN32
+#include <Windows.h>
+#else
+#include <pthread.h>
+#endif
+
+struct aws_rw_lock {
+#ifdef _WIN32
+    SRWLOCK lock_handle;
+#else
+    pthread_rwlock_t lock_handle;
+#endif
+};
+
+#ifdef _WIN32
+#define AWS_RW_LOCK_INIT { .lock_handle = SRWLOCK_INIT }
+#else
+#define AWS_RW_LOCK_INIT { .lock_handle = PTHREAD_RWLOCK_INITIALIZER }
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Initializes a new platform instance of mutex.
+ */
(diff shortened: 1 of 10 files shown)


tests/rw_lock_test.c
```diff
@@ -0,0 +1,104 @@
...
+    ASSERT_SUCCESS(aws_thread_join(&thread), "Thread join failed with error code %d.", aws_last_error());
+    ASSERT_INT_EQUALS(rw_lock_data.thread_fn_increments, rw_lock_data.max_counts, "Thread 2 should have written all data");
+    ASSERT_INT_EQUALS(rw_lock_data.max_counts, rw_lock_data.counter, "Both threads should have written exactly the max counts.");
+
+    aws_thread_clean_up(&thread);
+    aws_rw_lock_clean_up(&rw_lock_data.rw_lock);
+
+    return 0;
+}
+AWS_TEST_CASE(rw_lock_is_actually_rw_lock_test, test_rw_lock_is_actually_rw_lock)
```
RandyGaul: I’d add in a test to hammer the rw lock with lots of readers and a couple writers.

include/aws/common/rw_lock.h
```diff
@@ -0,0 +1,79 @@
...
+ * you should not treat it like one. On platforms it is possible for it to be non-reentrant, it will be.
+*/
+AWS_COMMON_API int aws_rw_lock_try_rdlock(struct aws_rw_lock *lock);
+AWS_COMMON_API int aws_rw_lock_try_wrlock(struct aws_rw_lock *lock);
+
+/**
+ * Releases the lock.
+ */
+AWS_COMMON_API int aws_rw_lock_rdunlock(struct aws_rw_lock *lock);
+AWS_COMMON_API int aws_rw_lock_wrunlock(struct aws_rw_lock *lock);
```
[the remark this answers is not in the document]
ColdenCullen: Sure, I was just trying to match pthread's api. But I think you're right, this'll be clearer.

ColdenCullen:
Because the implementation is platform specific, and it seems silly to make these functions not inline-able. Re-install: they are excluded. You'll notice that `AWS_COMMON_OS_HEADERS` is populated per-platform.

include/aws/common/windows/rw_lock.inl
```diff
@@ -0,0 +1,82 @@
...
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int aws_rw_lock_init(struct aws_rw_lock *lock) {
```
JonathanHenson: what's the point of putting these in cpp guards if we aren't doing dll semantics?
Even then why would we export platform specific symbols?

ColdenCullen: You're right, we probably don't need the cpp guards. Where do you see this exporting platform specific symbols?

include/aws/common/windows/rw_lock.inl
```diff
@@ -0,0 +1,82 @@
+#ifndef AWS_COMMON_WINDOWS_RW_LOCK_INL
```
JonathanHenson: I don't support of worrying about inlines here. At the point you're grabbing a lock, do you actually give a damn about a function call overhead?

ColdenCullen: I was just doing what @graebm suggested. You guys fight it out.

include/aws/common/windows/rw_lock.inl
graebm: shouldn't these implementations be marked `inline` to prevent the compiler from borking?

PR #89, opened 2018-07-19
clang-format & clang-tidy

ColdenCullen:
This PR contains two commits that should be reviewed independently. Each commit contains the config file for the appropriate clang tool, and the fixed code.

Note that clang-tidy will not make changes automatically, those were by hand. Non-conformance will fail the build.


format-check.sh
```diff
@@ -0,0 +1,22 @@
...
+do
+    $CLANG_FORMAT -output-replacements-xml $i | grep -c "<replacement " > /dev/null
+    if [ $? -ne 1 ]
+    then
+        echo "$i failed clang-format check."
+        FAIL=1
+    fi
+done
+
+exit $FAIL
```
graebm: needs newline at end of file

include/aws/common/error.h
```diff
@@ -82,27 +79,34 @@ AWS_COMMON_API void aws_reset_error(void);
...
- * temporarily. Setting this to NULL will turn off this error callback after it has been enabled.
+ * Sets an application wide error handler function. This will be overridden by
+ * the thread local handler. The previous handler is returned, this can be used
+ * for restoring an error handler if it needs to be overridden temporarily.
+ * Setting this to NULL will turn off this error callback after it has been
+ * enabled.
  */
-AWS_COMMON_API aws_error_handler aws_set_global_error_handler_fn(aws_error_handler handler, void *ctx);
+AWS_COMMON_API aws_error_handler
+    aws_set_global_error_handler_fn(aws_error_handler handler, void *ctx);
```
graebm: Could we get this so the function name is still on the 1st line?

source/array_list.c
```diff
@@ -70,7 +78,8 @@ void aws_array_list_clean_up(struct aws_array_list *list) {
 int aws_array_list_push_back(struct aws_array_list *list, const void *val) {
     int err_code = aws_array_list_set_at(list, val, list->length);
 
-    if(err_code && aws_last_error() == AWS_ERROR_INVALID_INDEX && !list->alloc) {
+    if (err_code && aws_last_error() == AWS_ERROR_INVALID_INDEX
+        && !list->alloc) {
         return aws_raise_error(AWS_ERROR_LIST_EXCEEDS_MAX_SIZE);
```
graebm: could you manually make this nicer looking:

```
if (err_code
    && aws_last_error() == AWS_ERROR_INVALID_INDEX
    && !list->alloc) {

    return aws_raise_error(AWS_ERROR_LIST_EXCEEDS_MAX_SIZE);
```

source/array_list.c
```diff
@@ -106,10 +115,12 @@ void aws_array_list_pop_front_n(struct aws_array_list *list, size_t n) {
         size_t popping_bytes = list->item_size * n;
         size_t remaining_items = list->length - n;
         size_t remaining_bytes = remaining_items * list->item_size;
-        memmove(list->data, (uint8_t *)list->data + popping_bytes, remaining_bytes);
+        memmove(
+            list->data, (uint8_t *)list->data + popping_bytes, remaining_bytes);
```
graebm: This seems weird. Either:
- we should do a line-break for each argument, if we do a line break for any arguments
- or we should try and fit as many args into 80 characters as possible

source/common.c
```diff
@@ -187,44 +217,127 @@ static struct aws_error_info errors[] = {
...
+        AWS_LIB_NAME),
+    AWS_DEFINE_ERROR_INFO(
+        AWS_ERROR_COND_VARIABLE_ERROR_UNKNOWN,
+        "condition variable unknown error.",
+        AWS_LIB_NAME),
+    AWS_DEFINE_ERROR_INFO(
+        AWS_ERROR_CLOCK_FAILURE,
+        "clock get ticks operation failed",
+        AWS_LIB_NAME),
     AWS_DEFINE_ERROR_INFO(AWS_ERROR_LIST_EMPTY, "empty list", AWS_LIB_NAME),
```
graebm: could you manually break these lines so it's like the others. thanks

source/common.c
```diff
@@ -187,44 +217,127 @@ static struct aws_error_info errors[] = {
...
+        "the caller of a mutex operation was not the owner",
+        AWS_LIB_NAME),
+    AWS_DEFINE_ERROR_INFO(
+        AWS_ERROR_MUTEX_FAILED,
+        "mutex operation failed",
+        AWS_LIB_NAME),
+    AWS_DEFINE_ERROR_INFO(
+        AWS_ERROR_COND_VARIABLE_INIT_FAILED,
+        "condition variable initialization failed.",
+        AWS_LIB_NAME),
```
graebm: [pretty please] we could eliminate all those `AWS_LIB_NAME),` lines with a local macro:
```
#define AWS_DEFINE_ERROR_INFO_COMMON(code, msg) \
    AWS_DEFINE_ERROR_INFO((code), (msg), AWS_LIB_NAME)
```
remember this morning when you taught me what currying meant

source/hash_table.c
```diff
@@ -531,9 +595,10 @@ int aws_hash_table_remove(struct aws_hash_table *map,
...
 }
 
-int aws_hash_table_foreach(struct aws_hash_table *map,
-    int (*callback)(void *baton, struct aws_hash_element *pElement), void *baton)
-{
+int aws_hash_table_foreach(
+    struct aws_hash_table *map,
+    int (*callback)(void *context, struct aws_hash_element *pElement),
+    void *context) {
     struct hash_table_state *state = map->pImpl;
```
graebm: it would be cool if there was a rule that enforced whitespace after a multiline blocking opening like this, so the args don't blend in with the function body

source/hash_table.c
```diff
@@ -599,14 +667,16 @@ bool aws_hash_iter_done(const struct aws_hash_iter *iter) {
 }
 
 void aws_hash_iter_next(struct aws_hash_iter *iter) {
-    if (!aws_hash_iter_done(iter)) { /* If already at end of table, do nothing. */
+    if (!aws_hash_iter_done(
+            iter)) { /* If already at end of table, do nothing. */
```
graebm: Can you just move this comment instead

source/hash_table.c
```diff
@@ -599,14 +667,16 @@ bool aws_hash_iter_done(const struct aws_hash_iter *iter) {
...
         get_next_element(iter, iter->slot + 1);
     }
 }
 
 void aws_hash_table_clear(struct aws_hash_table *map) {
     struct hash_table_state *state = map->pImpl;
-    if (state->destroy_key_fn) { /* Check whether we have destructors once before traversing table. */
+    if (state->destroy_key_fn) { /* Check whether we have destructors once
+                                    before traversing table. */
```
graebm: likewise, can you just move these comments

PR #96, opened 2018-07-24
Make Sanitizer LINK_FLAGS public

ColdenCullen:
Also converts all calls to `target_link_libraries` to use CMake 3.0 style `PUBLIC`/`PRIVATE` syntax.


CMakeLists.txt
```diff
@@ -93,7 +93,7 @@ file(GLOB COMMON_SRC
 add_library(${CMAKE_PROJECT_NAME} ${LIBTYPE} ${COMMON_SRC})
 aws_set_common_properties(${CMAKE_PROJECT_NAME} NO_WEXTRA)
 aws_add_sanitizers(${CMAKE_PROJECT_NAME} BLACKLIST "sanitizer-blacklist.txt")
-target_link_libraries(${CMAKE_PROJECT_NAME} ${PLATFORM_LIBS})
+target_link_libraries(${CMAKE_PROJECT_NAME} PUBLIC ${PLATFORM_LIBS})
```
marcomagdy: Please don't do this. If I link to aws-c-common, I don't want your link flags. They probably also won't work with a C++ linker.

ColdenCullen: This is not a change in functionality. When `target_link_libraries` is called without `PUBLIC`/`PRIVATE`/`INTERFACE`, flags are set to public. Note that this only includes things like `pthread` and `Kernel32`.

ColdenCullen: That is a good point. Probably something we should talk about as a group.

also, fwiw, there isn't a C++ linker, it's all `ld` all the way down :stuck_out_tongue: (unless it's mac, then it's `lld`).

JonathanHenson: Depends on what you mean by "guarantee", but on windows winsock is also included. Anyways, we won't be adding non-OS dependencies ever to any package in the aws-c* ecosystem.

PR #97, opened 2018-07-24
Lru cache to return const time number of elements.

JonathanHenson:


marcomagdy:
It might be too late for this, sorry I missed the PR when this was introduced. But anyway, here it is:

By putting the word 'lru' in the API, it becomes an implicit contract of how the cache behaves. The API should be just a cache API. This affords you the flexibility in the future to change how and when the cache evicts its contents without violating the contract. It also allows you to plug-in different cache eviction policies in the future, or let the user provide their own.

JonathanHenson:
From a design perspective I prefer doing concrete and then general when I have more than one. If I did that it would end up being something polymorphic, and plugging in the existing code would be trivial.

PR #104, opened 2018-07-30
Add aws_byte_buf_eq, aws_byte_buf_copy, aws_byte_buf_null

mutescu:
*Issue #, if available:*

*Description of changes:*
New functions for aws_byte_buf that were util for me. 


source/byte_buf.c
```diff
@@ -57,6 +57,46 @@ void aws_byte_buf_secure_clean_up(struct aws_byte_buf *buf) {
...
+    dest->capacity = 0;
+    dest->allocator = allocator;
+    if (src->buffer == NULL) {
+        dest->buffer = NULL;
+        return AWS_OP_SUCCESS;
+    }
+
+    dest->buffer = (uint8_t *) aws_mem_acquire(allocator, sizeof(uint8_t) * src->len);
+    if (dest->buffer == NULL) {
+        return aws_raise_error(AWS_ERROR_OOM);
```
JonathanHenson: You no longer need to raise this error, the allocator does it for you

source/byte_buf.c
```diff
@@ -57,6 +57,46 @@ void aws_byte_buf_secure_clean_up(struct aws_byte_buf *buf) {
...
+    dest->len = 0;
+    dest->capacity = 0;
+    dest->allocator = allocator;
+    if (src->buffer == NULL) {
+        dest->buffer = NULL;
+        return AWS_OP_SUCCESS;
+    }
+
+    dest->buffer = (uint8_t *) aws_mem_acquire(allocator, sizeof(uint8_t) * src->len);
+    if (dest->buffer == NULL) {
```
bdonlan: If the allocation fails we should set `dest->allocator` to `NULL`.

source/byte_buf.c
```diff
@@ -57,6 +57,46 @@ void aws_byte_buf_secure_clean_up(struct aws_byte_buf *buf) {
...
+    }
+
+    if (!a->buffer || !b->buffer) {
+        return (a->buffer == b->buffer);
+    }
+
+    return !memcmp(a->buffer, b->buffer, a->len);
+}
+
+int aws_byte_buf_copy(struct aws_allocator *allocator, struct aws_byte_buf *dest, const struct aws_byte_buf *src) {
```
bdonlan: I would call this `aws_byte_buf_dup` - as in `strdup` - to clarify that it's allocating the memory instead of copying into an existing buffer.

source/byte_buf.c
```diff
@@ -57,6 +57,46 @@ void aws_byte_buf_secure_clean_up(struct aws_byte_buf *buf) {
...
+    assert(allocator);
+    assert(dest);
+    assert(src);
+
+    dest->len = 0;
+    dest->capacity = 0;
+    dest->allocator = allocator;
+    if (src->buffer == NULL) {
+        dest->buffer = NULL;
+        return AWS_OP_SUCCESS;
```
bdonlan: `dest->allocator` should be NULL.

include/aws/common/byte_buf.h
```diff
@@ -154,6 +172,15 @@ static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t *bytes,
     return buf;
 }
 
+static inline struct aws_byte_buf aws_byte_buf_null() {
```
bdonlan: In C99 we can just write

    struct aws_byte_buf null_buf = { 0 };

PR #106, opened 2018-07-31
[CI] Add CodeBuild config files

ColdenCullen:
- Adds new CI scripts using shared execution logic
- Tests newly created CI jobs under the aws-sdk-common-runtime@ AWS account.

The old jobs will be removed in a separate PR once this is merged to master.


codebuild/common-trusty.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
JonathanHenson: set -e

then delete the pipes.

PR #110, opened 2018-08-02
Add cmake logic to ensure that each header builds on its own

bdonlan:
This didn't actually help with the winsock2.h issue but might come in handy in the future.

*Issue #, if available:*

*Description of changes:*


bdonlan:
The failing checks don't exist in the codebuild account, for some reason.

ColdenCullen:
After the move, you should now be logging into the `aws-sdk-common-runtime@amazon.com` account instead of the cpp sdk team's account.

CMakeLists.txt
```diff
@@ -11,12 +11,13 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-cmake_minimum_required(VERSION 3.0)
-project(aws-c-common C)
+cmake_minimum_required(VERSION 3.1)
+project(aws-c-common C CXX)
```
ColdenCullen: It looks like adding CXX support to this project broke the clang 6 build. You may need to add `export CXX=XXX` to the codebuild scripts to ensure it's using the correct compiler for both languages.

You'll also want to add `-m32` to the CXX flags for the gcc-4x-x86 build script. Currently it's trying to build the c file in 32bit and the cxx files in 64bit, and then obviously the link fails.

JonathanHenson: Xnay on adding CXX to this, this will enforce users having a C++ compiler to build.

It should be a separate project for this test (that is entirely optional) if we want to do this.

CMakeLists.txt
```diff
@@ -35,6 +35,13 @@ file(GLOB AWS_COMMON_SRC
         "source/*.c"
         )
 
+option(PERFORM_HEADER_CHECK "Performs compile-time checks that each header can be included independently. Requires a C++ compiler.")
+
+if (PERFORM_HEADER_CHECK) 
+    set(HEADERS_TO_CHECK ${AWS_COMMON_HEADERS} ${AWS_TEST_HEADERS})
```
graebm: `${AWS_COMMON_OS_HEADERS}` also needs to be checked

CMakeLists.txt
```diff
@@ -35,6 +35,13 @@ file(GLOB AWS_COMMON_SRC
...
+    add_subdirectory(cmake/header-tester)
```
graebm: should the files from cmake/header-tester be in this review? Or are they being reviewed separately?

codebuild/linux-gcc-4x-x86.yml
```diff
@@ -11,7 +11,7 @@ phases:
   build:
     commands:
       - echo Build started on `date`
-      - ./codebuild/common-linux.sh -DCMAKE_C_FLAGS="-m32" -DCMAKE_LINK_FLAGS="-m32"
+      - ./codebuild/common-linux.sh -DCMAKE_C_FLAGS="-m32" -DCMAKE_CXX_FLAGS="-m32" -DCMAKE_LINK_FLAGS="-m32"
```
ColdenCullen: I think you need to export `CXX` on this one too, looks like it's using the builtin `/usr/bin/c++`.

cmake/header-tester/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+# #includes any necessary dependencies, and that it builds under C++ as well.
+# 
+# To do so, we generate a single-line C or C++ source file that includes each
+# header, and link all of these stub source files into a test executable.
+#
+# Because we pull in a CXX compiler, this is split out into a subproject,
+# so that it can be disabled when not desired.
+
+cmake_minimum_required(VERSION 3.1)
+project(aws-c-common C CXX)
```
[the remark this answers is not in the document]
bdonlan: Ah, oops, will fix

CMakeLists.txt
bdonlan: I forgot to add these initially; they should be there now.

CMakeLists.txt
bdonlan: Added.

PR #118, opened 2018-08-07
Fix lookup3 valgrind warnings

bdonlan:
This switches lookup3 to use a marginally slower way of accessing the last 12
bytes of a string, in exchange for removing the valgrind
access-beyond-end-of-buffer warnings that arise from its normal methods.

While this is a slight pessimization, leaving this optimization in makes it
difficult for consuming applications and libraries to perform valgrind testing;
valgrind appears to attribute the uninitialized accesses to later code, making
it very difficult to accurately suppress these warnings. Considering cases
where the aws-c-common library builder may not be the same person as the
library user, it's better to have this in the valgrind-safe mode by default; we
can later make it togglable if the performance gains are needed in
ultra-high-performance systems.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/aws/common/private/lookup3.c
+++ b/include/aws/common/private/lookup3.c
@@ -3,9 +3,12 @@
  * # All functions have been made static.
  * # The self test harness has been turned off.
  * # stdint.h include removed for C89 compatibility.
+ * # VALGRIND has been enabled to suppress valgrind warnings
  *
  * The original code was retrieved from http://burtleburtle.net/bob/c/lookup3.c
  */
+
+#define VALGRIND
 
 /*
 -------------------------------------------------------------------------------


bdonlan:
Re suppressions: unfortunately, valgrind blames the code that uses the return value from lookup3, not lookup3 itself. Because of this, we'd need to suppress whatever code uses hash functions, which is too large of a blast radius (particularly since user code might use it too). Example suppressions generated by valgrind look like e.g.:

```
{
   <insert_a_suppression_name_here>
   Memcheck:Cond
   fun:s_get_next_element
   fun:aws_hash_iter_begin
   fun:aws_cryptosdk_hash_elems_array_init
   fun:aws_cryptosdk_serialize_enc_context_init
   fun:serialize_error_when_serialized_len_too_long
   fun:main
}
```

Re removing the ifdef code: I prefer leaving the imported code as close to the original upstream version as possible. This will make importing updated versions easier in the future, should it become necessary.

bdonlan:
Actually, hold off on this, I'm investigating this stuff further. It might not be from lookup3 after all...

bdonlan:
Misdiagnosed the valgrind errors. Closing this PR for now.

PR #120, opened 2018-08-08
aws_mem_acquire_many

ColdenCullen:
Adds `aws_mem_acquire_many`, a useful function for allocating lots of stuff contiguously. This is very good for cache locality. Extremely useful for pimpl pattern, i.e.:
```c
    struct aws_mqtt_client *client = NULL;
    struct aws_mqtt_client_impl *client_impl = NULL;
    aws_mem_acquire_many(allocator, 2,
        &client, sizeof(struct aws_mqtt_client),
        &client_impl, sizeof(struct aws_mqtt_client_impl));

    AWS_ZERO_STRUCT(*client);
    AWS_ZERO_STRUCT(*client_impl);

    client->impl = client_impl;
```

### DIFF
--- a/include/aws/common/common.h
+++ b/include/aws/common/common.h
@@ -172,6 +172,18 @@ void aws_wrapped_cf_allocator_destroy(CFAllocatorRef allocator);
 AWS_COMMON_API
 void *aws_mem_acquire(struct aws_allocator *allocator, size_t size);
 
+/**
+ * Allocates many chunks of bytes into a single block. Expects to be called with alternating void ** (dest), size_t
+ * (size). The first void ** will be set to the root of the allocation. Alignment is assumed to be sizeof(intmax_t).
+ *
+ * This is useful for allocating structs using the pimpl pattern, as you may allocate the public object and impl object
+ * in the same contiguous block of memory.
+ *
+ * Returns a pointer to the allocation.
+ */
+AWS_COMMON_API
+void *aws_mem_acquire_many(struct aws_allocator *allocator, size_t count, ...);
+
 /*
  * Releases ptr back to whatever allocated it.
  */
--- a/source/common.c
+++ b/source/common.c
@@ -62,7 +62,7 @@ void *aws_mem_acquire(struct aws_allocator *allocator, size_t size) {
     return mem;
 }
 
-#define AWS_ALIGN_ROUND_UP(value, alignment) (((value) + ((alignment)-1)) & -(alignment))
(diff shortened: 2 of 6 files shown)


source/common.c
```diff
@@ -61,6 +62,58 @@ void *aws_mem_acquire(struct aws_allocator *allocator, size_t size) {
     return mem;
 }
 
+#define AWS_ALIGN_ROUND_UP(value, alignment) (((value) + ((alignment) - 1)) & -(alignment))
+
+void *aws_mem_acquire_many(struct aws_allocator *allocator, size_t count, ...) {
+
+    enum { S_ALIGNMENT = sizeof(intmax_t) };
```
JonathanHenson: shouldn't we be aligning on 64 byte boundaries?

JonathanHenson: nevermind, talked offline, this is fine

include/aws/common/common.h
```diff
@@ -172,6 +172,18 @@ void aws_wrapped_cf_allocator_destroy(CFAllocatorRef allocator);
 AWS_COMMON_API
 void *aws_mem_acquire(struct aws_allocator *allocator, size_t size);
 
+/**
+ * Allocates many chunks of bytes into a single block. Expects to be called with alternating void ** (dest), size_t
+ * (size). The first void ** will be set to the root of the allocation. Alignment is assumed to be sizeof(intmax_t).
```
bdonlan: We shouldn't lock ourselves into intmax_t alignment, if we know the max alignment on our platform we should use that.

ColdenCullen: I know, I don't like either. In C11 you can do [`_Alignof`](https://en.cppreference.com/w/c/language/_Alignof)`(`[`max_align_t`](https://en.cppreference.com/w/c/types/max_align_t)`)`, but in C99 this was the closest thing I could find. If you know of another way to get default alignment, I'm all ears :smile:

That said, once we actually support aligned allocations, it'll be trivial to add another version of this that accepts trios of `ptr, size, alignment` and individually align each object.

JonathanHenson: I think we should just add an actual aws_mem_aquire_aligned function when we need it.

source/common.c
```diff
@@ -61,6 +62,58 @@ void *aws_mem_acquire(struct aws_allocator *allocator, size_t size) {
     return mem;
 }
 
+#define AWS_ALIGN_ROUND_UP(value, alignment) (((value) + ((alignment)-1)) & -(alignment))
```
marcomagdy: [trivial] prefer `~alignment`rather than `-alignment` so you're not relying on whether the machine uses 1 or 2 complement.

PR #127, opened 2018-08-17
Implement a SIMD-optimized constant-time base64 en/decoder

bdonlan:
This change implements a parallel (24-byte-at-a-time) base64 en/decoder using
AVX2 and SSSE4.1 instructions. Based on preliminary benchmarking, and depending
on CPU model, this provides a 3-4x speedup for both encode and decode for large
buffers; small buffers (<64 bytes) take double the time to process on my 2015
macbook, but on c5 class instances speed is comparable to the existing
implementation on the low end.

As a nice bonus, the new implementation avoids leaking data through cache
side-channels, as it does not use lookup tables for decoding.

Note that this change does not yet work on Windows; I need to implement a
processor feature test using CPUID directly for this to be enabled. For now, it
will be disabled at compile time.


bdonlan:
Performance on C5 hardware

bdonlan:
The above is the result of benchmarking on EC2 c5 hardware.

bdonlan:
I'll merge this and fix any weird cross-platform issues later, but putting this out for initial review now.

source/arch/encoding_simd_sse41.c
```diff
@@ -0,0 +1,396 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include "emmintrin.h"
```
JonathanHenson: why the ""?

source/arch/encoding_simd_sse41.c
```diff
@@ -0,0 +1,396 @@
...
+#include "immintrin.h"
+#include <ctype.h>
+#include <stdbool.h>
```
JonathanHenson: we don't include these. aws/common/common.h defines the types.

source/arch/encoding_simd_sse41.c
```diff
@@ -0,0 +1,396 @@
...
+    *in = _mm256_sub_epi8(tmp, _mm256_set1_epi8(1));
+
+    /* If any byte is now zero, we had a decode failure */
+    __m256i mask = _mm256_cmpeq_epi8(tmp, _mm256_set1_epi8(0));
+    return _mm256_testz_si256(mask, mask);
+}
+
+struct aligned256 {
+    unsigned char bytes[32];
+} __attribute__((aligned(32)));
```
JonathanHenson: AWS_ALIGN

source/arch/encoding_simd_sse41.c
```diff
@@ -0,0 +1,396 @@
...
+    uint64_t hi = _mm256_extract_epi64(vec, 2);
+
+    _mm_storeu_si128((__m128i *)out, lo);
+    memcpy(out + 16, &hi, sizeof(hi));
+
+    return true;
+}
+
+size_t aws_common_private_base64_decode_sse41(const unsigned char *in, unsigned char *out, size_t len) {
+    if (len % 4)
```
JonathanHenson: I prolly wrote this, but since we're on a perfomance kick here:

if (len & 0x03) {
    return AWS_OP_ERR;
}

source/arch/encoding_simd_sse41.c
```diff
@@ -0,0 +1,396 @@
...
+        unsigned char tmp_in[32];
+        unsigned char tmp_out[24];
+
+        memset(tmp_out, 0xEE, sizeof(tmp_out));
+
+        /* We need to ensure the vector contains valid b64 characters */
+        memset(tmp_in, 'A', sizeof(tmp_in));
+        memcpy(tmp_in, in, len);
+
+        size_t final_out = (3 * len) / 4;
```
JonathanHenson: could go for a >> 2 here.

source/encoding.c
```diff
@@ -17,6 +17,48 @@
...
+#else
+size_t aws_common_private_base64_decode_sse41(const unsigned char *in, unsigned char *out, size_t len) {
+    (void)in;
+    (void)out;
+    (void)len;
+
+    /*
+     * This volatile variable prevents MSVC from flagging code after the call to this stub as unreachable
+     */
+
```
JonathanHenson: extra space?

source/arch/encoding_simd_sse41.c
bdonlan: This should probably be changed.

source/arch/encoding_simd_sse41.c
bdonlan: Any decent optimizing compiler will deal with this for us.

source/arch/encoding_simd_sse41.c
bdonlan: Also will be handled by optimizing compilers.

source/arch/encoding_simd_sse41.c
ColdenCullen: I'm with Bryan on this one, `len % 4` is much clearer to the reader and per this [godbolt](https://godbolt.org/z/cDyOhw) instance, becomes a `test len, 3` at just -O1.

source/arch/encoding_simd_sse41.c
ColdenCullen: Similarly, is optimized away at -O1 ([godbolt](https://godbolt.org/z/_-fFos))

CMakeLists.txt
```diff
@@ -102,6 +102,10 @@ aws_set_common_properties(${CMAKE_PROJECT_NAME} NO_WEXTRA)
 aws_add_sanitizers(${CMAKE_PROJECT_NAME} BLACKLIST "sanitizer-blacklist.txt")
 target_link_libraries(${CMAKE_PROJECT_NAME} PUBLIC ${PLATFORM_LIBS})
 
+include(SIMD)
+
+set_source_files_properties(${CMAKE_CURRENT_SOURCE_DIR}/source/b64-simd.c PROPERTIES COMPILE_FLAGS "-mavx -mavx2 -mssse3 -msse4 -msse4.1")
```
tautschnig: Doesn't this unconditionally add all these flags?

source/arch/encoding_simd_sse41.c
```diff
@@ -0,0 +1,396 @@
...
+ */
+static inline __m256i translate_exact(__m256i in, uint8_t match, uint8_t decode) {
+    __m256i mask = _mm256_cmpeq_epi8(in, _mm256_set1_epi8(match));
+    return mask & _mm256_set1_epi8(decode);
+}
+
+/*
+ * Input: a pointer to a 256-bit vector of base64 characters
+ * The pointed-to-vector is replaced by a 256-bit vector of 6-bit decoded parts;
+ * on decode failure, returns 0, else returns 1 on success.
```
tautschnig: The return type is `bool`, hence the comment should be talking about `false` and `true`,  respectively.

source/arch/encoding_simd_sse41.c
```diff
@@ -0,0 +1,396 @@
...
+ */
+static inline bool decode_vec(__m256i *in) {
+    __m256i tmp;
+
+    tmp = translate_range(*in, 'A', 'Z', 0 + 1);
+    tmp |= translate_range(*in, 'a', 'z', 26 + 1);
+    tmp |= translate_range(*in, '0', '9', 52 + 1);
+    tmp |= translate_exact(*in, '+', 62 + 1);
+    tmp |= translate_exact(*in, '/', 63 + 1);
```
tautschnig: Don't use magic numbers. Doing arithmetic on `'/' - 'A'` is totally fine.

source/arch/encoding_simd_sse41.c
```diff
@@ -0,0 +1,396 @@
...
+ * Output: a 256-bit vector, the lower 24 bytes of which contain the packed version of the input
+ */
+static inline __m256i pack_vec(__m256i in) {
+    /*
+     * Our basic strategy is to split the input vector into three vectors, for each 6-bit component
+     * of each 24-bit group, shift the groups into place, then OR the vectors together. Conveniently,
+     * we can do this on a dword-by-dword basis.
+     *
+     * It's important to note that we're interpreting the vector as being little-endian. That is,
+     * on entry, we have 32-bit dwords that looks like this:
```
tautschnig: "32-bit dwords" is either wrong (on architectures where bytes are not 8 bits wide, but presumably we don't target these), or redundant (a dword is 4 bytes).

source/arch/encoding_simd_sse41.c
```diff
@@ -0,0 +1,396 @@
...
+
+/***** Encode logic *****/
+static inline __m256i encode_chars(__m256i in) {
+    __m256i tmp;
+
+    tmp = translate_range(in, 0, 25, 'A');
+    tmp |= translate_range(in, 26, 26 + 25, 'a');
+    tmp |= translate_range(in, 52, 61, '0');
+    tmp |= translate_exact(in, 62, '+');
+    tmp |= translate_exact(in, 63, '/');
```
tautschnig: As above: can magic numbers please be avoided?

source/encoding.c
```diff
@@ -17,6 +17,48 @@
...
+    volatile bool do_abort = true;
+    if (do_abort) {
+        abort();
+    }
```
tautschnig: So this function always aborts. Why does its implementation even exist? Shouldn't it just not be invoked on platforms that don't support it?

source/encoding.c
```diff
@@ -226,6 +268,13 @@ int aws_base64_encode(const struct aws_byte_buf *AWS_RESTRICT to_encode, struct
         return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
     }
 
+    if (aws_common_private_has_b64_simd()) {
```
tautschnig: Why is this a runtime rather than a compile-time check?

source/arch/encoding_simd_sse41.c
bdonlan: This is a base64 encoding table, effectively. These numbers are the base64 indices that correspond with 'a', '0', '+', '/', etc. `'/' - 'A'` is incorrect as these ranges are not contiguous. Extracting constants will just obscure this mapping - which is easier to compare with the RFC, `translate_exact(*in, '/', BASE64_FORWARD_SLASH_INDEX + 1)` or `translate_exact(*in, '/', 63 + 1)`?

source/arch/encoding_simd_sse41.c
bdonlan: Again, this is an encoding table, where else would the magic numbers go?

source/encoding.c
bdonlan: Because we may be executing on a CPU with different capabilities than the one that built the code.

source/arch/encoding_simd_sse41.c
bdonlan: We're already in intel-specific-code-land, and this is mostly a reminder to the reader that dword is defined as 32-bits on x86 (it may not be obvious for people who don't do a lot of windows or asm coding).

source/encoding.c
bdonlan: Correct. This is avoiding putting preprocessor blocks inside of the aws_base64_encode/decode functions, and instead consolidating them in one place at the top of the file.

source/arch/encoding_simd_sse41.c
tautschnig: Well, I have no way to compare to an RFC if I don't know about the RFC? I'm totally fine with keeping those numbers if you add a link to the RFC as a comment above the code. That would remove all "magic" from the numbers.

source/encoding.c
tautschnig: But then you'll have to re-arrange your code, as both the run-time detection as the actual code using the extensions are compiled with the same options (they are in the same file). From https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/X86-Built-in-Functions.html: "If you specify command-line switches such as -msse, the compiler could use the extended instruction sets even if the built-ins are not used explicitly in the program. For this reason, applications that perform run-time CPU detection must compile separate files for each supported architecture, using the appropriate flags. In particular, the file containing the CPU detection code should be compiled without these options."

source/arch/encoding_simd_sse41.c
tautschnig: Add comments. Either you made them up, or you got them from somewhere. If the former, say so. If the latter, add a link to your source.

source/arch/encoding_simd_sse41.c
tautschnig: Ok, but then please do so upon the first use of "dword." Don't educate people five lines later.

source/encoding.c
tautschnig: Ok. How about using a pragma instead of this somewhat strange construct? Something like:
```
#ifdef _MSC_VER
#pragma warning(disable:<WHATEVER THE WARNING NUMBER>)
#endif
```
This will 1) be a fix specific to the misbehaving compiler and 2) not break in surprising ways when the Visual Studio developers change their way of detecting dead code.

source/arch/encoding_simd_sse41.c
```diff
@@ -0,0 +1,396 @@
...
+        inlen -= stridelen;
+    }
+}
+
+#define CPUID_UNKNOWN 2
+#define CPUID_AVAILABLE 0
+#define CPUID_UNAVAILABLE 1
+static int cpuid_state = 2;
+
+void aws_common_private_set_b64_simd_state(int state) {
```
tautschnig: This is never used, and isn't documented either (how to use it, when to use it, what are valid parameter values).

CMakeLists.txt
bdonlan: Ah, yes, this should be removed. Leftover from early development.

source/encoding.c
bdonlan: `(void)x` is a very standard way of suppressing unused variable warnings, but I'll suppress it with a pragma here.

source/encoding.c
bdonlan: Correct. Using the flags for encoding.c is a bug in CMakeLists. SIMD.cmake configures the flags for the file under source/arch/ differently.

source/encoding.c
tautschnig: `(void)x` is, but this construct using `volatile` seems less standard, but I could be wrong.

bdonlan: Actually, this might warn on non-x86 GCC as well, so I'll stick with (void)s.

bdonlan: The issue that triggered the warning was a visible call to abort() - MSVC then determined that the function containing this unconditional abort() never returned, and created an unreachable code warning in the caller. The fix is to just return `-1` (a failure indication) instead; since the function is unreachable in the first place, we don't care too much what it returns.

PR #128, opened 2018-08-22
task scheduler revamp

graebm:
Scheduler revamped so that it can be relied on not to fail.

It was difficult to build systems that used the scheduler, when there was always the possibility that a task would fail to schedule. By storing `aws_task` in a linked-list most of the time, and forcing users to manage the memory of the `aws_task`, we can guarantee a task scheduled "now" cannot fail.

Array-lists are still used for storing tasks that are scheduled to run at a particular time, so tasks scheduled for the "future" might still fail do to lack of memory. We stuck with array-lists for scheduled tasks because these must be sorted, and sorting array-lists is significantly faster than sorting linked-lists.

The clock was also removed from the scheduler because, in reality, the user of the scheduler needed to keep another copy of the clock around for determining when to schedule tasks. By making the scheduler ignorant of where time comes from, it's simpler, easier to test, and more functions can be made infallible.

The "task popping" function was removed in favor of a simpler "has tasks" query.


source/task_scheduler.c
```diff
@@ -15,173 +15,116 @@
...
 }
 
 int aws_task_scheduler_schedule_future(
     struct aws_task_scheduler *scheduler,
     struct aws_task *task,
     uint64_t time_to_run) {
 
-    struct task_container container;
+    assert(task);
+    assert(task->fn);
```
ColdenCullen: `assert(scheduler)`

source/task_scheduler.c
```diff
@@ -15,173 +15,116 @@
...
 
-int aws_task_scheduler_schedule_now(struct aws_task_scheduler *scheduler, struct aws_task *task) {
-
-    uint64_t now;
-    if (scheduler->clock(&now)) {
-        return AWS_OP_ERR;
-    }
+void aws_task_scheduler_schedule_now(struct aws_task_scheduler *scheduler, struct aws_task *task) {
+    assert(task);
+    assert(task->fn);
```
ColdenCullen: `assert(scheduler)`. Assume everyone is dumb.

include/aws/common/task_scheduler.h
```diff
@@ -17,48 +17,55 @@
...
+}
+
 struct aws_task_scheduler {
     struct aws_allocator *alloc;
-    struct aws_priority_queue queue;
-    aws_task_scheduler_clock_fn *clock;
-    uint64_t min_run_time;
+    struct aws_priority_queue timed_queue; /* Tasks scheduled to run at specific times */
+    struct aws_linked_list asap_list;      /* Tasks scheduled to run as soon as possible */
+    struct aws_linked_list running_list;   /* Tasks currently being executed */
```
ColdenCullen: This can be local to the `s_run_all` function.

include/aws/common/linked_list.h
```diff
@@ -33,6 +33,14 @@ struct aws_linked_list_node {
     struct aws_linked_list_node *prev;
 };
 
+/**
+ * Set node's next and prev pointers to NULL.
+ */
+static inline void aws_linked_list_node_reset(struct aws_linked_list_node *node) {
+    node->next = NULL;
+    node->prev = NULL;
```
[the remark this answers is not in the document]
graebm: sure

include/aws/common/task_scheduler.h
```diff
@@ -17,48 +17,55 @@
...
     void *arg;
+    uint64_t timestamp;
+    struct aws_linked_list_node node;
 };
 
+static inline void aws_task_init(struct aws_task *task, aws_task_fn *fn, void *arg) {
+    task->fn = fn;
+    task->arg = arg;
+    task->timestamp = 0;
+    aws_linked_list_node_reset(&task->node);
```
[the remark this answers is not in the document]
graebm: sure

include/aws/common/task_scheduler.h
graebm: deeeerrrrrp me

include/aws/common/task_scheduler.h
```diff
@@ -68,26 +75,23 @@ AWS_COMMON_API
...
- * even if they are the highest priority. next_run_time, if it is not null, and there are any tasks scheduled
- * will be set to the timestamp for the highest priority task. This is a useful hint for setting timeouts on
- * event loops, or thread sleeps. If no tasks are scheduled, this value will be set to 0.
- * If no tasks are ready for execution AWS_ERROR_TASK_SCHEDULER_NO_READY_TASKS error will be raised.
- *
- * If no tasks are scheduled AWS_ERROR_TASK_SCHEDULER_NO_TASKS error will be raised.
- * task is copied.
+ * Returns whether the scheduler has any scheduled tasks.
+ * next_task_time (optional) will be set to time of the next task, note that 0 will be set if tasks were
+ * added via aws_task_scheduler_schedule_now() and UINT64_MAX will be set if no tasks are scheduled at all.
```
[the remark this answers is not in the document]
graebm: I like it when out-params are always set to something. I thought UINT64_MAX was a good value to use, because it basically means "the next task is infinity away"

source/task_scheduler.c
```diff
@@ -15,173 +15,116 @@
...
 
-static inline int s_get_next_task(
-    struct aws_task_scheduler *scheduler,
-    struct aws_task *task,
-    uint64_t run_before,
-    uint64_t *next_run_time);
+static void s_run_all(struct aws_task_scheduler *scheduler, uint64_t current_time, enum aws_task_status status);
+
+int aws_task_scheduler_init(struct aws_task_scheduler *scheduler, struct aws_allocator *alloc) {
+    assert(alloc);
```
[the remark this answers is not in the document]
graebm: it would crash on the next line: `scheduler->alloc = alloc;`

my internal rule thus far has been not to check stuff like that if it's going to crash due to dereference in the 1st few lines of the function. ex: `struct thing_impl *impl = thing->impl;` I wouldn't `assert(thing)` since it's immediately accessed

source/task_scheduler.c
graebm: :heart: assert :heart:

source/task_scheduler.c
graebm: :heart: assert :heart:

PR #129, opened 2018-08-23
Cursor/Buffer/String Conversions/Equality

ColdenCullen:
Adds the following:
* `aws_byte_cursor_eq`
* `aws_byte_cursor_eq_byte_buf`
* `aws_byte_cursor_from_string`

### DIFF
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -150,6 +150,22 @@ int aws_byte_buf_append(struct aws_byte_buf *to, struct aws_byte_cursor *from);
 AWS_COMMON_API
 int aws_byte_buf_cat(struct aws_byte_buf *dest, size_t number_of_args, ...);
 
+/**
+ * Compares two aws_byte_cursor structures
+ * Returns true if a has the same length as b and their buffers have the same bytes
+ * (or both buffers are null). When both a and b are null the function returns true
+ */
+AWS_COMMON_API
+bool aws_byte_cursor_eq(const struct aws_byte_cursor *a, const struct aws_byte_cursor *b);
+
+/**
+ * Compares an aws_byte_cursor against an aws_byte_buf
+ * Returns true if a has the same length as b and their buffers have the same bytes
+ * (or both buffers are null). When both a and b are null the function returns true
+ */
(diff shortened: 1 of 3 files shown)


include/aws/common/byte_buf.h
```diff
@@ -150,6 +150,22 @@ int aws_byte_buf_append(struct aws_byte_buf *to, struct aws_byte_cursor *from);
 AWS_COMMON_API
 int aws_byte_buf_cat(struct aws_byte_buf *dest, size_t number_of_args, ...);
 
+/**
+ * Compares two aws_byte_cursor structures
+ * Returns true if a has the same length as b and their buffers have the same bytes
+ * (or both buffers are null). When both a and b are null the function returns true
+ */
+AWS_COMMON_API
+bool aws_byte_cursor_eq(const struct aws_byte_cursor *a, const struct aws_byte_cursor *b);
```
graebm: would it be worth making these cmp() operators, returning -1 or 0 or 1 so they can be used for sorting as well?

ColdenCullen: Good question? I did `_eq` because all of the other functions were `_eq`. I'm not convinced there's a good reason to compare a blob of memory though unless you're sure it's a string, and we already have `aws_string_compare`. That said, I'm open to discussion.

source/byte_buf.c
```diff
@@ -167,6 +167,40 @@ int aws_byte_buf_cat(struct aws_byte_buf *dest, size_t number_of_args, ...) {
...
+    }
+
+    return !memcmp(a->ptr, b->ptr, a->len);
+}
+
+bool aws_byte_cursor_eq_byte_buf(const struct aws_byte_cursor *a, const struct aws_byte_buf *b) {
+
+    if (!a || !b) {
+        return ((void *)a == (void *)b);
+    }
```
graebm: why not just `return aws_byte_cursor_eq(a, aws_byte_cursor_from_buf(b));` ?

include/aws/common/byte_buf.h
graebm: ok, fair enough, just wanted to be sure we'd thought it through

PR #130, opened 2018-08-24
Convert writing methods to write to aws_byte_buf

ColdenCullen:
With this PR, contracts become much more clear. `aws_byte_cursor` is used for reading from, and `aws_byte_buf` is used for writing to. Users are still free to manage their memory with `aws_byte_buf` however, they just need to create a cursor for it before reading from it. Practice shows this doesn't actually happen a whole lot.

Example usage changes can be seen here, in the [IO](https://github.com/awslabs/aws-c-io/compare/tls-s2n...tls-s2n-write-buffer?expand=1) and [MQTT](https://github.com/awslabs/aws-c-mqtt/compare/write-buffer?expand=1) repos.

### DIFF
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -502,19 +502,19 @@ AWS_STATIC_IMPL bool aws_byte_cursor_read_be64(struct aws_byte_cursor *cur, uint
  * buffer unchanged.
  */
 AWS_STATIC_IMPL bool aws_byte_buf_write(
-    struct aws_byte_buf *AWS_RESTRICT cur,
+    struct aws_byte_buf *AWS_RESTRICT buf,
     const uint8_t *AWS_RESTRICT src,
     size_t len) {
 
-    if (cur->len > (SIZE_MAX >> 1) || len > (SIZE_MAX >> 1) || cur->len + len > cur->capacity) {
+    if (buf->len > (SIZE_MAX >> 1) || len > (SIZE_MAX >> 1) || buf->len + len > buf->capacity) {
         return false;
     }
 
-    memcpy(cur->buffer + cur->len, src, len);
-        cur->len += len;
+    memcpy(buf->buffer + buf->len, src, len);
+    buf->len += len;
 
-        return true;
-    }
+    return true;
+}
 
 /**
  * Copies all bytes from buffer to buffer.
@@ -524,9 +524,9 @@ AWS_STATIC_IMPL bool aws_byte_buf_write(
  * buffer unchanged.
  */
 AWS_STATIC_IMPL bool aws_byte_buf_write_from_whole_buffer(
-    struct aws_byte_buf *AWS_RESTRICT cur,
+    struct aws_byte_buf *AWS_RESTRICT buf,
     struct aws_byte_buf src) {
-    return aws_byte_buf_write(cur, src.buffer, src.len);
+    return aws_byte_buf_write(buf, src.buffer, src.len);
 }
 
 /**
@@ -537,9 +537,9 @@ AWS_STATIC_IMPL bool aws_byte_buf_write_from_whole_buffer(
  * buffer unchanged.
  */
 AWS_STATIC_IMPL bool aws_byte_buf_write_from_whole_cursor(
-    struct aws_byte_buf *AWS_RESTRICT cur,
+    struct aws_byte_buf *AWS_RESTRICT buf,
     struct aws_byte_cursor src) {
-    return aws_byte_buf_write(cur, src.ptr, src.len);
+    return aws_byte_buf_write(buf, src.ptr, src.len);
 }
 
 /**
@@ -551,8 +551,8 @@ AWS_STATIC_IMPL bool aws_byte_buf_write_from_whole_cursor(
  * If there is insufficient space in the cursor, returns false, leaving the
  cursor unchanged.
  */
-AWS_STATIC_IMPL bool aws_byte_buf_write_u8(struct aws_byte_buf *AWS_RESTRICT cur, uint8_t c) {
-    return aws_byte_buf_write(cur, &c, 1);
+AWS_STATIC_IMPL bool aws_byte_buf_write_u8(struct aws_byte_buf *AWS_RESTRICT buf, uint8_t c) {
+    return aws_byte_buf_write(buf, &c, 1);
 }
 
 /**
@@ -562,9 +562,9 @@ AWS_STATIC_IMPL bool aws_byte_buf_write_u8(struct aws_byte_buf *AWS_RESTRICT cur
  * If there is insufficient space in the cursor, returns false, leaving the
  * cursor unchanged.
  */
-AWS_STATIC_IMPL bool aws_byte_buf_write_be16(struct aws_byte_buf *cur, uint16_t x) {
+AWS_STATIC_IMPL bool aws_byte_buf_write_be16(struct aws_byte_buf *buf, uint16_t x) {
     x = aws_hton16(x);
-    return aws_byte_buf_write(cur, (uint8_t *)&x, 2);
+    return aws_byte_buf_write(buf, (uint8_t *)&x, 2);
 }
 
 /**
@@ -574,9 +574,9 @@ AWS_STATIC_IMPL bool aws_byte_buf_write_be16(struct aws_byte_buf *cur, uint16_t 
  * If there is insufficient space in the cursor, returns false, leaving the
  * cursor unchanged.
  */
-AWS_STATIC_IMPL bool aws_byte_buf_write_be32(struct aws_byte_buf *cur, uint32_t x) {
+AWS_STATIC_IMPL bool aws_byte_buf_write_be32(struct aws_byte_buf *buf, uint32_t x) {
     x = aws_hton32(x);
-    return aws_byte_buf_write(cur, (uint8_t *)&x, 4);
+    return aws_byte_buf_write(buf, (uint8_t *)&x, 4);
 }
 
 /**
@@ -586,9 +586,9 @@ AWS_STATIC_IMPL bool aws_byte_buf_write_be32(struct aws_byte_buf *cur, uint32_t 
  * If there is insufficient space in the cursor, returns false, leaving the
  * cursor unchanged.
  */
-AWS_STATIC_IMPL bool aws_byte_buf_write_be64(struct aws_byte_buf *cur, uint64_t x) {
+AWS_STATIC_IMPL bool aws_byte_buf_write_be64(struct aws_byte_buf *buf, uint64_t x) {
     x = aws_hton64(x);
-    return aws_byte_buf_write(cur, (uint8_t *)&x, 8);
+    return aws_byte_buf_write(buf, (uint8_t *)&x, 8);
 }
 
 #endif /* AWS_COMMON_BYTE_BUF_H */
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -26,6 +26,8 @@
  * to constant memory or memory that should otherwise not be freed by this
  * struct, set allocator to NULL and free function will be a no-op.
  *
+ * This structure used to define the output for all functions that write to a buffer.
+ *
  * Note that this structure allocates memory at the buffer pointer only. The
  * struct itself does not get dynamically allocated and must be either
  * maintained or copied to avoid losing access to the memory.
@@ -40,6 +42,8 @@ struct aws_byte_buf {
 
 /**
  * Represents a movable pointer within a larger binary string or buffer.
+ *
+ * This structure is used to define buffers for reading.
  */
 struct aws_byte_cursor {
     /* do not reorder this, this struct lines up nicely with windows buffer structures--saving us allocations */
@@ -61,7 +65,10 @@ int aws_byte_buf_init(struct aws_allocator *allocator, struct aws_byte_buf *buf,
  * Returns AWS_OP_SUCCESS in case of success or AWS_OP_ERR when memory can't be allocated.
  */
 AWS_COMMON_API
-int aws_byte_buf_init_copy(struct aws_allocator *allocator, struct aws_byte_buf *dest, const struct aws_byte_buf *src);
+int aws_byte_buf_init_copy(
+    struct aws_allocator *allocator,
+    struct aws_byte_buf *dest,
+    const struct aws_byte_cursor *src);
 
 AWS_COMMON_API
 void aws_byte_buf_clean_up(struct aws_byte_buf *buf);
@@ -101,7 +108,7 @@ bool aws_byte_buf_eq(const struct aws_byte_buf *a, const struct aws_byte_buf *b)
  * long enough to use the results.
  */
 AWS_COMMON_API
-bool aws_byte_buf_next_split(const struct aws_byte_buf *input_str, char split_on, struct aws_byte_cursor *substr);
+bool aws_byte_cursor_next_split(const struct aws_byte_cursor *input_str, char split_on, struct aws_byte_cursor *substr);
 
 /**
  * No copies, no buffer allocations. Fills in output with a list of
@@ -124,7 +131,10 @@ bool aws_byte_buf_next_split(const struct aws_byte_buf *input_str, char split_on
  * long enough to use the results.
  */
 AWS_COMMON_API
-int aws_byte_buf_split_on_char(const struct aws_byte_buf *input_str, char split_on, struct aws_array_list *output);
+int aws_byte_cursor_split_on_char(
+    const struct aws_byte_cursor *AWS_RESTRICT input_str,
+    char split_on,
+    struct aws_array_list *output);
 
 /**
  * No copies, no buffer allocations. Fills in output with a list of aws_byte_cursor instances where buffer is
@@ -144,8 +154,8 @@ int aws_byte_buf_split_on_char(const struct aws_byte_buf *input_str, char split_
  * It is the user's responsibility to make sure the input buffer stays in memory long enough to use the results.
  */
 AWS_COMMON_API
-int aws_byte_buf_split_on_char_n(
-    const struct aws_byte_buf *input_str,
+int aws_byte_cursor_split_on_char_n(
+    const struct aws_byte_cursor *AWS_RESTRICT input_str,
     char split_on,
     struct aws_array_list *output,
     size_t n);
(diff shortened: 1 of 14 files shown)


include/aws/common/byte_buf.h
```diff
@@ -444,14 +454,19 @@ static inline bool aws_byte_cursor_read_be64(struct aws_byte_cursor *cur, uint64
  * If there is insufficient space in the cursor, returns false, leaving the
  * cursor unchanged.
  */
-static inline bool aws_byte_cursor_write(
-    struct aws_byte_cursor *AWS_RESTRICT cur,
+static inline bool aws_byte_buf_write(
+    struct aws_byte_buf *AWS_RESTRICT cur,
```
graebm: I see `aws_byte_buf_write()` and `aws_byte_buf_write_from_whole_buffer()`, but there is no `aws_byte_buf_write_from_whole_cursor()`?

source/byte_buf.c
```diff
@@ -94,12 +97,12 @@ int aws_byte_buf_init_copy(struct aws_allocator *allocator, struct aws_byte_buf
...
+    memcpy(dest->buffer, src->ptr, src->len);
     return AWS_OP_SUCCESS;
 }
 
-int aws_byte_buf_split_on_char_n(
-    struct aws_byte_buf *input_str,
+int aws_byte_cursor_split_on_char_n(
+    const struct aws_byte_cursor *AWS_RESTRICT input_str,
     char split_on,
     struct aws_array_list *output,
```
bdonlan: Why no restrict?

source/byte_buf.c
```diff
@@ -143,8 +146,11 @@ int aws_byte_buf_split_on_char_n(
     return AWS_OP_SUCCESS;
 }
 
-int aws_byte_buf_split_on_char(struct aws_byte_buf *input_str, char split_on, struct aws_array_list *output) {
-    return aws_byte_buf_split_on_char_n(input_str, split_on, output, 0);
+int aws_byte_cursor_split_on_char(
+    const struct aws_byte_cursor *AWS_RESTRICT input_str,
+    char split_on,
+    struct aws_array_list *output) {
```
bdonlan: Why no restrict?

source/byte_buf.c
```diff
@@ -201,7 +207,7 @@ bool aws_byte_cursor_eq_byte_buf(const struct aws_byte_cursor *a, const struct a
     return !memcmp(a->ptr, b->buffer, a->len);
 }
 
-int aws_byte_buf_append(struct aws_byte_buf *to, struct aws_byte_cursor *from) {
+int aws_byte_buf_append(struct aws_byte_buf *to, const struct aws_byte_cursor *AWS_RESTRICT from) {
```
bdonlan: Looks like this should work with `to == from`, so drop the restrict.

include/aws/common/byte_buf.h
```diff
@@ -139,7 +149,7 @@ int aws_byte_buf_split_on_char_n(
  * returned. dest->len will contain the amount of data actually copied to dest.
  */
 AWS_COMMON_API
-int aws_byte_buf_append(struct aws_byte_buf *to, struct aws_byte_cursor *from);
+int aws_byte_buf_append(struct aws_byte_buf *to, const struct aws_byte_cursor *AWS_RESTRICT from);
```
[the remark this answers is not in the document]
ColdenCullen: Should be safe to self append, I'll remove restrict.

include/aws/common/byte_buf.h
```diff
@@ -182,11 +192,11 @@ static inline struct aws_byte_buf aws_byte_buf_from_c_str(const char *c_str) {
     return buf;
 }
 
-static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t *bytes, size_t len) {
+static inline struct aws_byte_buf aws_byte_buf_from_array(const void *bytes, size_t len, size_t capacity) {
```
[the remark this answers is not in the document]
ColdenCullen: That's fair, I guess the only time you would want capacity to be different is on a fresh array where capacity is 0.

include/aws/common/byte_buf.h
```diff
@@ -438,85 +455,103 @@ static inline bool aws_byte_cursor_read_be64(struct aws_byte_cursor *cur, uint64
...
- * On success, returns true and updates the cursor pointer/length
- accordingly.<<<<<< str-split
+ * On success, returns true and updates the buffer /length accordingly.
+ * If there is insufficient space in the buffer, returns false, leaving the
+ * buffer unchanged.
+ */
+static inline bool aws_byte_buf_write_from_whole_cursor(
+    struct aws_byte_buf *AWS_RESTRICT cur,
+    const struct aws_byte_cursor *AWS_RESTRICT src) {
+    return aws_byte_buf_write(cur, src->ptr, src->len);
```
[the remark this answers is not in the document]
ColdenCullen: I don't think so, since it would empty the cursor out. I'd opt to not clear potentially useful information in favor of letting the user do that if they want to. Especially since this would work pretty well:
```c
aws_byte_buf_write_from_whole_cursor(out_buf,
        aws_byte_cursor_advance(my_data_cursor, num_bytes_to_write));
```

source/byte_buf.c
```diff
@@ -73,15 +73,18 @@ bool aws_byte_buf_eq(const struct aws_byte_buf *a, const struct aws_byte_buf *b)
     return !memcmp(a->buffer, b->buffer, a->len);
 }
 
-int aws_byte_buf_init_copy(struct aws_allocator *allocator, struct aws_byte_buf *dest, const struct aws_byte_buf *src) {
+int aws_byte_buf_init_copy(
```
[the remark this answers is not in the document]
ColdenCullen: Sure! I'll do both.

source/byte_buf.c
ColdenCullen: Because I... forgot.

include/aws/common/byte_buf.h
```diff
@@ -61,7 +65,10 @@ int aws_byte_buf_init(struct aws_allocator *allocator, struct aws_byte_buf *buf,
  * Returns AWS_OP_SUCCESS in case of success or AWS_OP_ERR when memory can't be allocated.
  */
 AWS_COMMON_API
-int aws_byte_buf_init_copy(struct aws_allocator *allocator, struct aws_byte_buf *dest, const struct aws_byte_buf *src);
```
bdonlan: If we're writing methods which consume from a cursor but _don't_ advance the cursor, we should document this explicitly.

include/aws/common/byte_buf.h
```diff
@@ -108,7 +115,10 @@ bool aws_byte_buf_eq(const struct aws_byte_buf *a, const struct aws_byte_buf *b)
  * long enough to use the results.
  */
 AWS_COMMON_API
-int aws_byte_buf_split_on_char(struct aws_byte_buf *input_str, char split_on, struct aws_array_list *output);
+int aws_byte_cursor_split_on_char(
```
bdonlan: Where is the cursor left at the end? If e.g. the output array_list has a limited size I could see it being useful to have the cursor be at the point where it left off.

include/aws/common/byte_buf.h
```diff
@@ -128,18 +138,18 @@ int aws_byte_buf_split_on_char(struct aws_byte_buf *input_str, char split_on, st
  * It is the user's responsibility to make sure the input buffer stays in memory long enough to use the results.
  */
 AWS_COMMON_API
-int aws_byte_buf_split_on_char_n(
-    struct aws_byte_buf *input_str,
+int aws_byte_cursor_split_on_char_n(
```
bdonlan: Same question - where is the cursor at the end, particularly if we hit `n`?

include/aws/common/byte_buf.h
```diff
@@ -128,18 +138,18 @@ int aws_byte_buf_split_on_char(struct aws_byte_buf *input_str, char split_on, st
...
+    const struct aws_byte_cursor *AWS_RESTRICT input_str,
     char split_on,
-    struct aws_array_list *output,
+    struct aws_array_list *AWS_RESTRICT output,
     size_t n);
 
 /**
  * Copies from to to. If to is too small, AWS_ERROR_DEST_COPY_TOO_SMALL will be
  * returned. dest->len will contain the amount of data actually copied to dest.
```
bdonlan: Document that `to == from` is permitted.

include/aws/common/byte_buf.h
```diff
@@ -191,20 +201,45 @@ static inline struct aws_byte_buf aws_byte_buf_from_array(const uint8_t *bytes,
     return buf;
 }
 
+static inline struct aws_byte_buf aws_byte_buf_from_empty_array(const void *bytes, size_t capacity) {
```
bdonlan: Would it be more useful to have a variant with both length and capacity args?

include/aws/common/byte_buf.h
```diff
@@ -438,85 +473,103 @@ static inline bool aws_byte_cursor_read_be64(struct aws_byte_cursor *cur, uint64
...
- * Copies one byte to cursor.
+ * Copies all bytes from buffer to buffer.
  *
- * On success, returns true and updates the cursor pointer/length
- accordingly.<<<<<< str-split
+ * On success, returns true and updates the buffer /length accordingly.
+ * If there is insufficient space in the buffer, returns false, leaving the
+ * buffer unchanged.
+ */
+static inline bool aws_byte_buf_write_from_whole_cursor(
```
bdonlan: Ergonomic suggestion: If we're not going to modify the cursor, it may be better to pass it by value instead of as a pointer. Not only does this provide a clear indication that we're not going to modify the cursor, but it also makes it easier to do things like:

    aws_byte_buf_write_from_whole_cursor(&some_buf, aws_byte_cursor_from_c_str(some_c_string));

instead of:

    struct aws_byte_cursor tmp = aws_byte_cursor_from_c_str(some_c_string);
    aws_byte_buf_write_from_whole_cursor(&some_buf, &tmp);

include/aws/common/byte_buf.h
bdonlan: It's further down in the file.

include/aws/common/byte_buf.h
```diff
@@ -438,85 +473,103 @@ static inline bool aws_byte_cursor_read_be64(struct aws_byte_cursor *cur, uint64
...
+static inline bool aws_byte_buf_write(
+    struct aws_byte_buf *AWS_RESTRICT cur,
     const uint8_t *AWS_RESTRICT src,
     size_t len) {
-    struct aws_byte_cursor slice = aws_byte_cursor_advance_nospec(cur, len);
+
+    struct aws_byte_cursor write_cursor;
+    write_cursor.ptr = cur->buffer + cur->len;
+    write_cursor.len = cur->capacity - cur->len;
+    struct aws_byte_cursor slice = aws_byte_cursor_advance(&write_cursor, len);
```
bdonlan: Why are we doing this dance with cursors here?

include/aws/common/encoding.h
```diff
@@ -39,7 +39,7 @@ int aws_hex_compute_encoded_len(size_t to_encode_len, size_t *encoded_length);
  * output.
  */
 AWS_COMMON_API
-int aws_hex_encode(const struct aws_byte_buf *AWS_RESTRICT to_encode, struct aws_byte_buf *AWS_RESTRICT output);
```
bdonlan: I'll note that we are inconsistent wrt whether inputs or outputs go first between the core byte_buf/byte_cursor APIs and the encoding APIs. While we're repaving the earth, can we fix this as well?

source/byte_buf.c
```diff
@@ -143,8 +146,11 @@ int aws_byte_buf_split_on_char_n(
     return AWS_OP_SUCCESS;
 }
 
-int aws_byte_buf_split_on_char(struct aws_byte_buf *input_str, char split_on, struct aws_array_list *output) {
-    return aws_byte_buf_split_on_char_n(input_str, split_on, output, 0);
+int aws_byte_cursor_split_on_char(
```
bdonlan: Also has inconsistent inputs and output parameter ordering.

include/aws/common/byte_buf.h
ColdenCullen: Like you said, I think accepting it by value gets the point across. I'll do that.

include/aws/common/byte_buf.h
ColdenCullen: Good thinking, I'll make it do that.

ColdenCullen: Just kidding, that only applies to the N permutation. This version just keeps appending to the list for ever (assumes dynamic list).

include/aws/common/byte_buf.h
ColdenCullen: Will do

include/aws/common/byte_buf.h
ColdenCullen: I don't think so, so far all use cases I've seen have followed one of the two patterns. I've yet to see a len that isn't 0 or capacity.

include/aws/common/byte_buf.h
ColdenCullen: Originally I had this using the `nospec` version and didn't want to rewrite it for buffers, but I should be able to factor it out now.

include/aws/common/encoding.h
ColdenCullen: Yup!

bdonlan: Hmm. In cases where we're doing a conversion, there's no clear dividing line that would say that the input is the thing we're performing the action on, vs the output buffer. Remember that the C standard library uses the convention that output buffers always come first; why should we not follow that here?

ColdenCullen:
I've just been reminded how annoying these APIs are without this change, so I'm going to try to ram this through.

@bdonlan @david-koenig if it would make your migration path easier, I could make a new `byte_buf_deprecated.h` or something with all the old byte_cursor_write functions in it, would that be helpful? That way all your code would continue to work.

bdonlan:
I think it'll be simpler for us to pin a revision before this lands instead. I'll read the PR a bit later today.

include/aws/common/byte_buf.h
```diff
@@ -108,7 +115,10 @@ bool aws_byte_buf_eq(const struct aws_byte_buf *a, const struct aws_byte_buf *b)
  * long enough to use the results.
  */
 AWS_COMMON_API
-int aws_byte_buf_split_on_char(struct aws_byte_buf *input_str, char split_on, struct aws_array_list *output);
+int aws_byte_cursor_split_on_char(
+    struct aws_byte_cursor input_str,
+    char split_on,
+    struct aws_array_list *AWS_RESTRICT output);
```
bdonlan: Why output after input?

include/aws/common/byte_buf.h
```diff
@@ -123,23 +133,28 @@ int aws_byte_buf_split_on_char(struct aws_byte_buf *input_str, char split_on, st
...
 AWS_COMMON_API
-int aws_byte_buf_split_on_char_n(
-    struct aws_byte_buf *input_str,
+int aws_byte_cursor_split_on_char_n(
+    struct aws_byte_cursor input_str,
     char split_on,
-    struct aws_array_list *output,
-    size_t n);
+    size_t n,
+    struct aws_array_list *AWS_RESTRICT output);
```
bdonlan: Same.

include/aws/common/byte_buf.h
```diff
@@ -439,85 +479,100 @@ AWS_STATIC_IMPL bool aws_byte_cursor_read_be64(struct aws_byte_cursor *cur, uint
...
- * If there is insufficient space in the cursor, returns false, leaving the
- * cursor unchanged.
+ * On success, returns true and updates the buffer length accordingly.
+ * If there is insufficient space in the buffer, returns false, leaving the
+ * buffer unchanged.
  */
-AWS_STATIC_IMPL bool aws_byte_cursor_write(
-    struct aws_byte_cursor *AWS_RESTRICT cur,
+AWS_STATIC_IMPL bool aws_byte_buf_write(
+    struct aws_byte_buf *AWS_RESTRICT cur,
```
bdonlan: Parameter name shouldn't be `cur` (this is an issue on multiple functions here)

include/aws/common/encoding.h
bdonlan: Did we come to a consensus on this?

include/aws/common/encoding.h
```diff
@@ -79,14 +79,14 @@ int aws_base64_compute_decoded_len(const char *input, size_t len, size_t *decode
...
-int aws_base64_decode(const struct aws_byte_buf *AWS_RESTRICT to_decode, struct aws_byte_buf *AWS_RESTRICT output);
+int aws_base64_decode(const struct aws_byte_cursor *AWS_RESTRICT to_decode, struct aws_byte_buf *AWS_RESTRICT output);
 
 AWS_EXTERN_C_END
 
 /* Add a 64 bit unsigned integer to the buffer, ensuring network - byte order
  * Assumes the buffer size is at least 8 bytes.
  */
-AWS_STATIC_IMPL void aws_write_u64(uint8_t *buffer, uint64_t value) {
+AWS_STATIC_IMPL void aws_write_u64(uint64_t value, uint8_t *buffer) {
```
bdonlan: Why is the output buffer last?

include/aws/common/string.h
```diff
@@ -125,16 +125,16 @@ AWS_EXTERN_C_END
...
- * accordingly. If there is insufficient space in the cursor, returns
- * false, leaving the cursor unchanged.
+ * On success, returns true and updates the buf pointer/length
+ * accordingly. If there is insufficient space in the buf, returns
+ * false, leaving the buf unchanged.
  */
-AWS_STATIC_IMPL bool aws_byte_cursor_write_from_whole_string(
-    struct aws_byte_cursor *AWS_RESTRICT cur,
+AWS_STATIC_IMPL bool aws_byte_buf_write_from_whole_string(
+    struct aws_byte_buf *AWS_RESTRICT cur,
```
bdonlan: parameter name does not match doc comment

include/aws/common/encoding.h
ColdenCullen: I'm with Henso on this one, I like having the out param last. Plus, in the case of the encoding functions, it's easier to think of the input as the `this` param than the output. I.E. I think this:
```c
arrayOfThings.encode(outbuf);
```
is clearer than:
```c
outbuf.encode(arrayOfThings);
```

ColdenCullen:
Ok I fixed the CBMC checks! The issue is a semicolon was missing, not sure how that ever worked.

If we can come to a consensus on the out param first vs last thing, I think this should be g2g.

tautschnig:
> Ok I fixed the CBMC checks! The issue is a semicolon was missing, not sure how that ever worked.

That semicolon got lost in the commit "Reordered a bunch of parameters, more @bdonlan feedbacks" in this PR.

ColdenCullen:
> > Ok I fixed the CBMC checks! The issue is a semicolon was missing, not sure how that ever worked.
> 
> That semicolon got lost in the commit "Reordered a bunch of parameters, more @bdonlan feedbacks" in this PR.

Aaaah that would do it.

PR #131, opened 2018-08-29
Memory pool

RandyGaul:
Memory pool implementation. Use case is to implement a logger that utilizes a message pool. Source pulled from old logger branch and updated here in this PR: https://github.com/awslabs/aws-c-common/tree/logger


include/aws/common/memory_pool.h
```diff
@@ -0,0 +1,68 @@
...
+ * Release memory at `to_release`. Releases the memory to the pool or via `alloc->mem_release` depending on if
+ * `to_release` was allocated from the internal arena, or from a "one-shot" allocation (happens when the arena is
+ * completely full).
+ */
+AWS_COMMON_API void aws_memory_pool_release(struct aws_memory_pool *pool, void *to_release);
+
+/**
+ * Returns the internal memory arena and its size in the pointers `arena` and `arena_size` respectively.
+ */
+AWS_COMMON_API void aws_memory_pool_get_arena(struct aws_memory_pool *pool, void **arena, size_t *arena_size);
```
graebm: What's the use-case for this as a public function? Maybe remove it?

ColdenCullen:
Where you able to maintain API compatibility with [the pool in c-io](https://github.com/awslabs/aws-c-io/blob/master/include/aws/io/message_pool.h)? That would save some time.

source/memory_pool.c
```diff
@@ -0,0 +1,116 @@
...
+    } else {
+        aws_mem_release(pool->alloc, pool);
+    }
+}
+
+void *aws_memory_pool_try_acquire(struct aws_memory_pool *pool) {
+    if (pool->free_list) {
+        void *mem = pool->free_list;
+        pool->free_list = *((void **)pool->free_list);
+        return mem;
```
ColdenCullen: Can we `memset(mem, 0, pool->element_size)` before returning this? I just hit a bug in my MQTT code because I was assuming that the memory the memory pool returns is 0'd.

JonathanHenson:
we should pull the memory pool out of aws-c-io, to use this.

PR #133, opened 2018-08-29
AWS_ZERO_STRUCT and AWS_ZERO_ARRAY to do while pattern

RandyGaul:
Noticed a couple macros didn't have do-while pattern and it freaked me out a little bit. Really should swap these to prevent the off-chance the macro expansion messes with scoping rules in an odd manner.


include/aws/common/common.h
```diff
@@ -336,8 +336,8 @@ static inline void aws_secure_zero(void *pBuf, size_t bufsize) {
 #endif     // #else not windows
 }
 
-#define AWS_ZERO_STRUCT(object) memset(&object, 0, sizeof(object));
-#define AWS_ZERO_ARRAY(array) memset((void *)array, 0, sizeof(array));
+#define AWS_ZERO_STRUCT(object) do { memset(&(object), 0, sizeof(object)); } while (0)
+#define AWS_ZERO_ARRAY(array) do { memset((void *)array, 0, sizeof(array)); } while (0)
```
graebm: what's the advantage of doing this on these particular macros? They're already 1-liners?

I could see advantage to removing the semicolon from the original macro though, forcing users to put in one of their own.

RandyGaul: A) The macro can be modified freely to include more than one statement; B) it's a safe rule to have all function-like macros wrapped into do-while loops, and brings some consistency; C) people are less likely to write these kinds of macros with an ending semicolon if in general things are wrapped up in the do-while, case in point here someone wrote the macros with semicolons baked in.

PR #134, opened 2018-08-30
Add a aws_hash_table_put helper

bdonlan:
This makes the common use case of inserting a key/value pair with overwrite
easier (removes the need to declare an element local).

*Issue #, if available:*

*Description of changes:*


source/hash_table.c
```diff
@@ -496,6 +496,38 @@ int aws_hash_table_create(
...
+
+    if (!*was_created) {
+        if (p_elem->key != key) {
+            destroy_key_fn((void *)p_elem->key);
+        }
+
+        destroy_value_fn((void *)p_elem->value);
+    }
+
+    p_elem->key = key;
```
david-koenig: Very minor, but strictly this only needs to be inside the conditional started at line 517. Otherwise it's a redundant assignment.

source/hash_table.c
```diff
@@ -496,6 +496,38 @@ int aws_hash_table_create(
...
+int aws_hash_table_put(struct aws_hash_table *map, const void *key, void *value, int *was_created) {
+    struct hash_table_state *state = map->p_impl;
+    struct aws_hash_element *p_elem;
+    int was_created_fallback;
+
+    if (!was_created) {
+        was_created = &was_created_fallback;
+    }
+
+    /* We might end up resizing the table, so stash the destroy fns now */
```
david-koenig: I don't understand why these functions needed to be stashed here.

bdonlan: I guess I can move the state initialization down below the create call.

source/hash_table.c
bdonlan: True enough, but it's basically free and more clearly captures the intent.

PR #135, opened 2018-08-31
Mini Logger

RandyGaul:


bdonlan:
Please explain the revert in 8bdafcf; there are still deadlock issues remaining due to the inconsistent use of different mutexes with a single cvar (and associated state).

source/log.c
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /* Push message onto queue. */
+    aws_mutex_lock(&s_mutex);
+    aws_linked_list_push_front(&s_ctx->messages, &msg->node);
+    aws_mutex_unlock(&s_mutex);
+
+    if (s_log_thread_running) {
+        /* Notify log thread. */
+        aws_condition_variable_notify_one(&s_cv);
```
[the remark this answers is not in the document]
bdonlan: Retaking the lock isn't the right way to do this. You had the mutex on line 199, just don't unlock it until you finish notifying.

source/log.c
```diff
@@ -0,0 +1,304 @@
...
+
+struct s_log_ctx {
+    struct aws_allocator *alloc;
+    void *msg_pool;
+    size_t message_size;
+    struct aws_linked_list messages;
+};
+
+/* Logging globals. */
+struct s_log_ctx *s_ctx;
```
bdonlan: All of these globals should probably be static.

source/log.c
```diff
@@ -0,0 +1,296 @@
...
+    }
+    s_ctx = (struct s_log_ctx *)aws_mem_acquire(alloc, sizeof(*s_ctx));
+    if (!s_ctx) {
+        return aws_raise_error(AWS_ERROR_LOG_FAILURE);
+    }
+    s_ctx->alloc = alloc;
+    s_ctx->msg_pool = NULL;
+    s_ctx->message_size = max_message_len;
+    (void)memory_pool_message_count;
+    aws_mutex_init(&s_mutex);
```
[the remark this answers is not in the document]
bdonlan: Note also that reinitializing the mutex is unsafe if there's a parallel call being made to any of the log APIs.

include/aws/common/common.h
```diff
@@ -340,6 +351,4 @@ static inline void aws_secure_zero(void *pBuf, size_t bufsize) {
 #define AWS_ZERO_ARRAY(array) memset((void *)array, 0, sizeof(array));
 #define AWS_ARRAY_SIZE(array) (sizeof(array) / sizeof(array[0]))
 
-#define AWS_ENABLE_HW_OPTIMIZATION 1
```
ColdenCullen: why u remove dis?

include/aws/common/log.h
```diff
@@ -0,0 +1,104 @@
...
+#include <stdarg.h>
+
+enum aws_log_level {
+    AWS_LOG_LEVEL_OFF = 0, /* No logs will be processed at all. */
+    AWS_LOG_LEVEL_FATAL,
+    AWS_LOG_LEVEL_ERROR,
+    AWS_LOG_LEVEL_WARN,
+    AWS_LOG_LEVEL_INFO,
+    AWS_LOG_LEVEL_DEBUG,
+    AWS_LOG_LEVEL_TRACE
```
ColdenCullen: trailing comma plz

include/aws/common/log.h
```diff
@@ -0,0 +1,104 @@
...
+};
+
+typedef void (aws_log_report_fn)(const char *tag, const char *log_message);
+
+#define AWS_LOG(level, tag, fmt) do { aws_log(level, tag, fmt); } while (0)
+#define AWS_VLOG(level, tag, fmt, ...) do { aws_log(level, tag, fmt, __VA_ARGS__); } while (0)
```
ColdenCullen: why VLOG and not LOGV?

include/aws/common/log.h
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Sets the global log level. Log levels above this level are ignored.
+ */
+AWS_COMMON_API void aws_log_system_set_level(enum aws_log_level level);
+
+/**
+ * Retrieves the global log level.
+ */
+AWS_COMMON_API enum aws_log_level aws_log_system_get_level(void);
```
ColdenCullen: remove `void`

include/aws/common/log.h
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Cleans up all logging related resources and flushes any remaining logs.
+ */
+AWS_COMMON_API void aws_log_system_clean_up();
```
ColdenCullen: Put this next to `init`

include/aws/common/log.h
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Call this function to process any previous logs captured by `aws_log` calls. Can be called in a loop, on a condition variable,
+ * or by any other means deemed necessary.
+ */
+AWS_COMMON_API int aws_log_flush();
+
+/**
+ * Spawns a thread that wakes whenever logs are recieved, and calls `aws_log_flush`.
+ */
+AWS_COMMON_API int aws_log_spawn_log_thread(struct aws_allocator *alloc);
```
ColdenCullen: If `init` already takes an allocator, why can't we reuse that here?

source/log.c
```diff
@@ -0,0 +1,304 @@
...
+ * or in the "license" file accompanying this file.This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied.See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/condition_variable.h>
+#include <aws/common/linked_list.h>
+#include <aws/common/log.h>
+//#include <aws/common/memory_pool.h>
```
ColdenCullen: remove commented out code

source/log.c
```diff
@@ -0,0 +1,304 @@
...
+    void *msg_pool;
+    size_t message_size;
+    struct aws_linked_list messages;
+};
+
+/* Logging globals. */
+struct s_log_ctx *s_ctx;
+struct aws_mutex s_mutex = AWS_MUTEX_INIT;
+aws_log_report_fn *s_report;
+enum aws_log_level s_level;
```
ColdenCullen: Why aren't these in the ctx object?

source/log.c
```diff
@@ -0,0 +1,296 @@
...
+struct aws_condition_variable s_cv = AWS_CONDITION_VARIABLE_INIT;
+struct aws_mutex s_cv_mutex = AWS_MUTEX_INIT;
+
+static inline struct s_msg *s_msg_new(void) {
+    if (!s_ctx) {
+        return NULL;
+    }
+
+    aws_mutex_lock(&s_mutex);
+    struct s_msg *msg = (struct s_msg *)aws_mem_acquire(s_ctx->alloc, sizeof(struct s_msg) + s_ctx->message_size);
```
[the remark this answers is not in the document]
JonathanHenson: we assume the thread safety of allocators. This is usually the case anyways since most allocators are stateless and they pin any state they need to the memory they allocate anyways. If they're more complicated than that, it's up to the allocator to handle it.

source/log.c
```diff
@@ -0,0 +1,296 @@
...
+        aws_mem_release(s_ctx->alloc, msg);
+        aws_mutex_unlock(&s_mutex);
+    }
+}
+
+static inline char *s_get_msg_data(struct s_msg *msg) {
+    return (char *)(msg + 1);
+}
+
+void aws_log_set_reporting_callback(aws_log_report_fn *report_callback) {
```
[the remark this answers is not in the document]
JonathanHenson: all callbacks should carry a context, or user_data pointer. At the very least, I'd need to track my fd and or file rotater.

source/log.c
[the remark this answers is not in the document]
JonathanHenson: There shouldn't be parallel calls made here, so I'm fine with that part, but it is already initialized.

source/log.c
```diff
@@ -0,0 +1,296 @@
...
+    /* Null out s_ctx and cleanup messages. */
+    aws_mutex_lock(&s_mutex);
+    while (!aws_linked_list_empty(&s_ctx->messages)) {
+        struct s_msg *msg = AWS_CONTAINER_OF(aws_linked_list_pop_front(&s_ctx->messages), struct s_msg, node);
+        aws_mem_release(s_ctx->alloc, msg);
+    }
+    struct s_log_ctx *ctx = s_ctx;
+    s_ctx = NULL;
+    aws_mutex_unlock(&s_mutex);
+
```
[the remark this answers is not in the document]
JonathanHenson: I think it's a fair tradeoff here to say there won't be and if there is it's a programmer error. 

this should be the last line in a main() function.

source/log.c
```diff
@@ -0,0 +1,296 @@
...
+    aws_log_destroy_log_thread();
+    aws_mem_release(ctx->alloc, ctx);
+}
+
+int aws_log(enum aws_log_level level, const char *tag, const char *fmt, ...) {
+    if (!(level != AWS_LOG_LEVEL_OFF && level >= s_level)) {
```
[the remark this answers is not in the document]
JonathanHenson: Simple demorgans makes this a bit easier to read.

if (level == AWS_LOG_LEVEL_OFF || level < s_level)

include/aws/common/log.h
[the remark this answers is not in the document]
bdonlan: Flush may be useful for flushing logs prior to application shutdown.

source/log.c
bdonlan: The pain of having to pass around the log context means this infrastructure probably won't actually get used if it's not global - or at least inherited-thread-local.

PR #139, opened 2018-09-07
Implement atomics wrappers

bdonlan:
*Issue #, if available:*

*Description of changes:*


bdonlan:
Probably the tests are failing because they don't have enough cores to run quickly. I'll set some timeouts.

include/aws/common/atomics_msvc.inc
```diff
@@ -0,0 +1,324 @@
...
+
+/**
+ * Reads an atomic var as an integer, using the specified ordering, and returns the result.
+ */
+AWS_STATIC_IMPL
+aws_atomic_int_impl_t aws_atomic_load_int(
+    volatile const struct aws_atomic_var *var,
+    enum aws_memory_order memory_order) {
+    aws_atomic_priv_check_order(memory_order);
+    return var->intval;
```
graebm: I went debug-stepping through MSVC 2017's x64 code, they implemented it like this: 
```
_Uint8_t _Load_seq_cst_8(volatile _Uint8_t *_Tgt)
    _Uint8_t _Value;
    _Value = *_Tgt;
    _Compiler_barrier();
    return _Value;
```
(from xatomic.h)

Is it really safe to just read the value if Microsoft is using a barrier afterwards?

include/aws/common/atomics.h
```diff
@@ -0,0 +1,247 @@
...
+ *
+ * As future work, we could test to see if the C11 atomics API on this platform behaves consistently
+ * with the other APIs and use it if it does.
+ */
+
+/**
+ * aws_atomic_int_t is defined to be an integer type that can fit into an struct aws_atomic_var.
+ * It is guaranteed to be at least 32 bits wide, but the size may vary depending on architecture.
+ */
+typedef aws_atomic_int_impl_t aws_atomic_int_t;
```
JonathanHenson: is this going to be the same size as size_t ? could we make it use those semantics? can we make it unsigned?

include/aws/common/atomics.h
```diff
@@ -0,0 +1,247 @@
...
+/**
+ * Initializes an atomic variable with a pointer value. This operation should be done before any
+ * other operations on this atomic variable, and must be done before attempting any parallel operations.
+ */
+void aws_atomic_init_ptr(volatile struct aws_atomic_var *var, void *p);
+
+/**
+ * Reads an atomic var as an integer, using the specified ordering, and returns the result.
+ */
+aws_atomic_int_t aws_atomic_load_int(volatile const struct aws_atomic_var *var, enum aws_memory_order memory_order);
```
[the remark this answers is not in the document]
bdonlan: We could but this doubles the number of wrappers to write.

include/aws/common/atomics_fallback.inc
```diff
@@ -0,0 +1,26 @@
+/*
```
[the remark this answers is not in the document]
bdonlan: Ah, oops. Will rename.

include/aws/common/atomics_gnu.inc
```diff
@@ -0,0 +1,245 @@
...
+        case aws_memory_order_acquire:
+            return __ATOMIC_ACQUIRE;
+        case aws_memory_order_release:
+            return __ATOMIC_RELEASE;
+        case aws_memory_order_acq_rel:
+            return __ATOMIC_ACQ_REL;
+        case aws_memory_order_seq_cst:
+            return __ATOMIC_SEQ_CST;
+        default: /* Unknown memory order */
+            abort();
```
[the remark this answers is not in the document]
bdonlan: Generally, I would expect the atomics wrappers to get inlined out, and this switch to be eliminated as the switch condition is likely to be a constant. In release builds, `assert(0)` would generate less code, but `abort()` is also fairly small.

include/aws/common/atomics_gnu.inc
```diff
@@ -0,0 +1,245 @@
...
+    return __atomic_fetch_xor(&var->intval, n, aws_atomic_priv_xlate_order(order));
+}
+
+/**
+ * Provides the same reordering guarantees as an atomic operation with the specified memory order, without
+ * needing to actually perform an atomic operation.
+ */
+AWS_STATIC_IMPL
+void aws_atomic_thread_fence(enum aws_memory_order order) {
+    __atomic_thread_fence(order);
```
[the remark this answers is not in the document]
bdonlan: Ah, good catch.

include/aws/common/atomics_msvc.inc
```diff
@@ -0,0 +1,324 @@
...
+#include <aws/common/common.h>
+
+#include <intrin.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#include <Winnt.h>
+
+#if !(defined(_M_IX86) || defined(_M_X64))
+#    error Atomics are not currently supported for non-x86 MSVC platforms
```
JonathanHenson: are we about to just break the arm build? Can we at least have a fallback?

bdonlan: The reason I've disabled this is because I can't test it, and having subtly broken atomics is far worse than having atomics that don't build.

include/aws/common/atomics_msvc.inc
bdonlan: https://msdn.microsoft.com/en-us/library/12a04hfd.aspx

A read of a volatile object (except on ARM, apparently) has acquire semantics. This means that no reads or writes that come after the read will be reordered prior to the read. This is stronger than the compiler barrier itself implies (as it includes hardware reordering as well). Most likely MSVC has that barrier for ARM support.

include/aws/common/atomics.h
bdonlan: Hmm. I made this signed because I expect that use cases might want to add negative numbers, but if we assume twos complement that's not essential. As for size_t, we don't know for sure what sizes will be supported, but I suppose `size_t` is pretty likely to be supported, so I could maybe hard-code that.

include/aws/common/atomics_gnu.inc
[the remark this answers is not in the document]
graebm: good point, it will probably be inlined out

include/aws/common/atomics_msvc.inc
```diff
@@ -0,0 +1,324 @@
...
+    aws_atomic_priv_check_order(order);
+
+    switch (order) {
+        case aws_memory_order_seq_cst:
+            MemoryBarrier();
+            break;
+        case aws_memory_order_release:
+        case aws_memory_order_acquire:
+        case aws_memory_order_acq_rel:
+            x = x;
```
graebm: Weirdly, MSVC's thread_fence only uses MemoryBarrier() for ARM, and for other architectures uses an atomic exchange on a dummy static.

```
inline void _Atomic_thread_fence(memory_order _Order)
	{	/* force memory visibility and inhibit compiler reordering */
 #if defined(_M_ARM) || defined(_M_ARM64)
	if (_Order != memory_order_relaxed)
		{
		_Memory_barrier();
		}

 #else
	_Compiler_barrier();
	if (_Order == memory_order_seq_cst)
		{	/* force visibility */
		static _Uint4_t _Guard;
		_Atomic_exchange_4(&_Guard, 0, memory_order_seq_cst);
		_Compiler_barrier();
		}
 #endif
	}
```

just pointing out the differences

include/aws/common/atomics_msvc.inc
bdonlan: This is correct, seq_cst only needs hardware barriers on stores on x86.

cmake/AwsCFlags.cmake
```diff
@@ -24,7 +24,9 @@ function(aws_set_common_properties target)
     cmake_parse_arguments(SET_PROPERTIES "${options}" "" "" ${ARGN})
 
     if(MSVC)
-        list(APPEND AWS_C_FLAGS /W4 /WX)
+	# /volatile:iso relaxes some implicit memory barriers that MSVC normally applies for volatile accesses
+	# Since we want to be compatible with user builds using /volatile:iso, use it for the tests.
+        list(APPEND AWS_C_FLAGS /W4 /WX /volatile:iso)
```
graebm: [trivial] tabs -> spaces
[trivial] break the volatile flag out into its own line

include/aws/common/atomics_fallback.inc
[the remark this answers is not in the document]
graebm: still using ".inc"

include/aws/common/atomics_msvc.inc
```diff
@@ -0,0 +1,368 @@
...
+        case aws_memory_order_seq_cst:
+            return;
+        default: /* Unknown memory order */
+            abort();
+    }
+#endif
+    (void)order;
+}
+
+enum aws_atomic_mode_priv { aws_atomic_priv_load, aws_atomic_priv_store, aws_atomic_priv_load_store };
```
graebm: `aws_atomic_priv_load_store` isn't used anywhere

bdonlan: I'll remove it.

cmake/AwsCFlags.cmake
bdonlan: Whoops, fixing.

PR #141, opened 2018-09-11
Don't do memcpy of header on string creation

david-koenig:
Small improvement to constructor function that removes a memcpy.

### DIFF
--- a/source/string.c
+++ b/source/string.c
@@ -14,12 +14,6 @@
  */
 #include <aws/common/string.h>
 
-#ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


source/string.c
```diff
@@ -25,12 +25,12 @@ struct aws_string *aws_string_new_from_c_str(struct aws_allocator *allocator, co
 }
```
ColdenCullen: You should be able to revert the warning disables above, too!

david-koenig: Good catch! I just removed them.

PR #148, opened 2018-09-25
Fixed builds for ancient linux distros.

JonathanHenson:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,6 +101,8 @@ endif()
 
 add_library(${CMAKE_PROJECT_NAME} ${LIBTYPE} ${COMMON_SRC})
 aws_set_common_properties(${CMAKE_PROJECT_NAME} NO_WEXTRA)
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -101,6 +101,11 @@ endif()
 
 add_library(${CMAKE_PROJECT_NAME} ${LIBTYPE} ${COMMON_SRC})
 aws_set_common_properties(${CMAKE_PROJECT_NAME} NO_WEXTRA)
+if (UNIX AND NOT APPLE)
```
bdonlan: is NOT APPLE required here? If so, why?

JonathanHenson: It's not required because it handles POSIX_C_SOURCE differently (namely with the SC_NPROCESSORS_ONLN ) flag. 

Even though per these docs:
http://man7.org/linux/man-pages/man3/sysconf.3.html, it's in this posix version,
Apple implemented it in a later release

PR #152, opened 2018-10-05
Add an API to allow for arbitrary pqueue nodes to be removed in O(log(n)) time

bdonlan:
*Issue #, if available:*

*Description of changes:*


source/priority_queue.c
```diff
@@ -91,43 +145,123 @@ void aws_priority_queue_init_static(
...
+    struct aws_priority_queue_node *backpointer) {
     int err = aws_array_list_push_back(&queue->container, item);
     if (err) {
         return err;
     }
 
+    size_t index = aws_array_list_length(&queue->container) - 1;
+
+    if (backpointer && !queue->backpointers.alloc) {
+        if (!queue->backpointers.alloc) {
```
JonathanHenson: why is this branch here when we already know this condition has to be true in order to get here?

bdonlan: oops, leftover from a code reorganization

PR #153, opened 2018-10-05
Default Allocator Leak Checker

RandyGaul:
Implemented a quick and dirty leak checker for the default allocator in aws-c-common. Useful for grabbing file/line pairs, even Apple machines (where valgrind/sanitizers can't perform leak checking).

Just throwing this out here in case anyone else thinks it might be good to add to common. Otherwise it can just live in this branch, and I'll personally refer to it as needed.

Everything is compiled away by default, except for two extra parameters passed to `aws_mem_acquire` and `aws_mem_realloc` (one for `const char *file` and one for `int line`). To enable the leak checking `AWS_DEFAULT_ALLOCATOR_LEAKCHECK` can be defined.

There's a function called `aws_mem_print_leaks_from_default_allocator`, which prints to `stderr` like so:

```
LEAKED 280 bytes from file "/Users/randgaul/Desktop/work/aws-c-common/source/common.c" at line 269 from address 0x6120000004e8.
LEAKED 280 bytes from file "/Users/randgaul/Desktop/work/aws-c-common/source/common.c" at line 269 from address 0x612000000368.
LEAKED 120 bytes from file "/Users/randgaul/Desktop/work/aws-c-common/source/common.c" at line 269 from address 0x60e000000228.
LEAKED 1096 bytes from file "/Users/randgaul/Desktop/work/aws-c-common/source/common.c" at line 269 from address 0x6190000023a8.
LEAKED 664 bytes from file "/Users/randgaul/Desktop/work/aws-c-common/source/common.c" at line 269 from address 0x6170000000a8.
WARNING: Memory leaks detected (see above).
```

And if no leaks are there, prints somethiing like: `SUCCESS: No memory leaks detected.`

I considered implementing leak checking in a way such that even user-provided allocators could use the leak printing function -- but this turned out to not really be worth mangling the allocator API. Instead, just providing this function for the default allocator was good enough for my own uses.

### DIFF
--- a/include/aws/common/common.h
+++ b/include/aws/common/common.h
@@ -138,10 +138,10 @@ AWS_STATIC_ASSERT(sizeof(char) == 1);
 
 /* Allocator structure. An instance of this will be passed around for anything needing memory allocation */
 struct aws_allocator {
-    void *(*mem_acquire)(struct aws_allocator *allocator, size_t size);
+    void *(*mem_acquire)(struct aws_allocator *allocator, size_t size, const char *file, int line);
     void (*mem_release)(struct aws_allocator *allocator, void *ptr);
     /* Optional method; if not supported, this pointer must be NULL */
-    void *(*mem_realloc)(struct aws_allocator *allocator, void *oldptr, size_t oldsize, size_t newsize);
+    void *(*mem_realloc)(struct aws_allocator *allocator, void *oldptr, size_t oldsize, size_t newsize, const char *file, int line);
     void *impl;
 };
 
@@ -154,6 +154,15 @@ typedef const struct __CFAllocator *CFAllocatorRef;
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+AWS_COMMON_API
+void *aws_default_malloc(struct aws_allocator *allocator, size_t size, const char *file, int line);
+
+AWS_COMMON_API
+void aws_default_free(struct aws_allocator *allocator, void *ptr);
+
+AWS_COMMON_API
+void *aws_default_realloc(struct aws_allocator *allocator, void *ptr, size_t oldsize, size_t newsize, const char *file, int line);
 
 AWS_COMMON_API
 struct aws_allocator *aws_default_allocator(void);
@@ -178,7 +187,7 @@ void aws_wrapped_cf_allocator_destroy(CFAllocatorRef allocator);
  * Returns at least `size` of memory ready for usage or returns NULL on failure.
  */
 AWS_COMMON_API
-void *aws_mem_acquire(struct aws_allocator *allocator, size_t size);
+void *aws_mem_acquire_implementation(struct aws_allocator *allocator, size_t size, const char *file, int line);
 
 /**
  * Allocates many chunks of bytes into a single block. Expects to be called with alternating void ** (dest), size_t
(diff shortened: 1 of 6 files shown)


include/aws/common/common.h
```diff
@@ -138,10 +138,10 @@ AWS_STATIC_ASSERT(sizeof(char) == 1);
 
 /* Allocator structure. An instance of this will be passed around for anything needing memory allocation */
 struct aws_allocator {
-    void *(*mem_acquire)(struct aws_allocator *allocator, size_t size);
+    void *(*mem_acquire)(struct aws_allocator *allocator, size_t size, const char *file, int line);
```
justinboswell: My suggestion here would be to have a structure which encapsulates file/line, and potentially future things like name/category/flags or any other distinguishing info we may come up with. You can construct the structure and pass it by value via macro. That way the API doesn't need to change should we need to collect more info, we could just add an additional macro or something to provide it.

RandyGaul: Hey good idea :)

include/aws/common/common.h
```diff
@@ -216,6 +225,13 @@ int aws_mem_realloc(struct aws_allocator *allocator, void **ptr, size_t oldsize,
  * that we can leave unchanged on failure.
  */
 
+#define aws_mem_acquire(alloc, size) aws_mem_acquire_implementation(alloc, size, __FILE__, __LINE__)
+#define aws_mem_realloc(alloc, ptr, oldsize, newsize) aws_mem_realloc_implementation(alloc, ptr, oldsize, newsize, __FILE__, __LINE__)
+
+#ifdef AWS_DEFAULT_ALLOCATOR_LEAKCHECK
+AWS_COMMON_API void aws_mem_print_leaks_from_default_allocator(void);
```
bdonlan: Should this just be a no-op function if the macro is not defined?

source/common.c
```diff
@@ -28,34 +33,114 @@
...
+
+    aws_mutex_lock(&s_mutex);
+    struct aws_allocation *head = s_head();
+    allocation->prev = head;
+    allocation->next = head->next;
+    head->next->prev = allocation;
+    head->next = allocation;
+    aws_mutex_unlock(&s_mutex);
+
+    mem = (void *)(allocation + 1);
```
bdonlan: The alignment of this memory may be incorrect. For example, on x86_64, your struct will be `5 * 8` bytes, with 8 byte alignment, but there are certain SIMD operations which require 16-byte or 32-byte alignment.

You'll either need to add some platform-specific minimum alignment values, or use something like a hash table instead.

source/common.c
```diff
@@ -28,34 +33,114 @@
...
 #endif
 
-static void *s_default_malloc(struct aws_allocator *allocator, size_t size) {
+#ifdef AWS_DEFAULT_ALLOCATOR_LEAKCHECK
+struct aws_allocation {
+    const char *file;
+    int line;
+    size_t size;
+
+    struct aws_allocation *next;
```
bdonlan: We have an intrusive linked-list already, shouldn't we be using those helper functions?

source/common.c
```diff
@@ -150,6 +235,28 @@ int aws_mem_realloc(struct aws_allocator *allocator, void **ptr, size_t oldsize,
...
+    while (next != head) {
+        fprintf(stderr, "LEAKED %zu bytes from file \"%s\" at line %d from address %p.\n", next->size, next->file, next->line, (void*)(next + 1));
+        next = next->next;
+        leaks = 1;
+    }
+
+    if (leaks) {
+        fprintf(stderr, "WARNING: Memory leaks detected (see above).\n");
+    } else {
+        fprintf(stderr, "SUCCESS: No memory leaks detected.\n");
```
bdonlan: Perhaps we should return a bool or something at the end

PR #155, opened 2018-10-08
Make most array_list methods inlinable

bdonlan:
Most of the array list methods have very simple logic and are likely to be used
in performance-critical segments (eg, as part of the implementation of a
priority queue). This change makes them inlinable, to reduce such overhead.
It also adds restrict annotations to the array list parameter, to further reduce
spills due to false aliasing.

*Issue #, if available:*

*Description of changes:*


include/aws/common/array_list.inl
```diff
@@ -0,0 +1,230 @@
...
+/* This is implicitly included, but helps with editor highlighting */
+#include <aws/common/array_list.h>
+/*
+ * Do not add system headers here; add them to array_list.h. This file is included under extern "C" guards,
+ * which might break system headers.
+ */
+
+
+AWS_STATIC_IMPL
+int aws_array_list_init_dynamic(
```
bdonlan: Note that, while I put some of the resize/allocation routines out-of-line, making the initializers inline means the compiler can have visibility into the `list->item_size` element in other inlined methods (particularly given the `AWS_RESTRICT` annotation). This may allow it to make further optimizations, such as eliminating multiplications in favor of cheaper scaled-displacement addressing modes.

include/aws/common/array_list.h
```diff
@@ -17,6 +17,11 @@
  */
 #include <aws/common/common.h>
 
+#include <assert.h>
+#include <stdlib.h>
+
+#define ARRAY_LIST_DEBUG_FILL 0xDD
```
justinboswell: Shouldn't this be prefixed with AWS_?

bdonlan: indeed.

PR #157, opened 2018-10-09
Add badges to README.md, fix LGTM Alert

ColdenCullen:
probe_cap is INT_MAX, and probe_idx is int, so probe_idx can't possibly be > probe_cap.


source/hash_table.c
```diff
@@ -327,7 +327,6 @@ static int s_find_entry1(
     /* If we find a deleted entry, we record that index and return it as our probe index (i.e. we'll keep searching to
```
bdonlan: While we're in here, I think the probe index needs to be `size_t` as well.

justinboswell: +1 to this.

PR #160, opened 2018-10-10
Fix CMake packaging & installation script

marcomagdy:
*Description of changes:*

This allows upstream projects to consume this library more easily and idiomatically.

The changes needed are mainly:
- Include a namespace for the library (surprisingly this is not just for aesthetics)
- Split the config from the targets file. This matters when this library has dependencies. In this particular case, it doesn't, but for consistency with the other packages, we should keep it that way.


CMakeLists.txt
```diff
@@ -140,17 +140,32 @@ foreach(HEADER_SRCPATH IN ITEMS ${AWS_COMMON_HEADERS} ${AWS_COMMON_OS_HEADERS} $
...
+  @ONLY)
+
+install(EXPORT "${CMAKE_PROJECT_NAME}-targets"
+    DESTINATION "${LIBRARY_DIRECTORY}/${CMAKE_PROJECT_NAME}/cmake/"
+    NAMESPACE AWS::)
+
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}-config.cmake"
+    DESTINATION "${LIBRARY_DIRECTORY}/${CMAKE_PROJECT_NAME}/cmake/")
+
+list(APPEND EXPORT_MODULES
```
bdonlan: How much of this needs to be installed? Consumers of this library shouldn't need to use the SIMD-flag-setting logic, for example.

marcomagdy: This part is from the original script.
@ColdenCullen can speak for it I guess.
My understanding is that those are helpful CMake scripts that can be utilized by other projects, and since aws-c-common is the lowest in the dependency tree, it made sense to have it be part of its installation.

PR #163, opened 2018-10-12
Enable clang-tidy on Travis

ColdenCullen:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,17 @@ os:
 compiler:
   - clang
 
+install:
+  - brew update
+  - brew install llvm
+  - export PATH="/usr/local/opt/llvm/bin:$PATH"
+
 #execute build
 script:
   - mkdir build
   - cd build
-  - cmake -DENABLE_SANITIZERS=ON ..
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -9,10 +9,17 @@ os:
...
+
 #execute build
 script:
   - mkdir build
   - cd build
-  - cmake -DENABLE_SANITIZERS=ON ..
+  - cmake -DENABLE_SANITIZERS=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
   - make && make test
+  - cd ..
+  - clang-tidy -p=build **/*.c
```
bdonlan: Can we split the clang-tidy run out to a separate job to make it clear at a glance what failed?

ColdenCullen: Travis doesn't really make it easy to split out jobs like that, but on the bright side you don't need to log in or anything to view travis results, so if you click the failure link and scroll to the bottom it should be pretty easy to figure out. Example: https://travis-ci.org/awslabs/aws-c-common/builds/440823707

PR #164, opened 2018-10-12
Add an aws_hash_table_eq helper

bdonlan:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/aws/common/hash_table.h
+++ b/include/aws/common/hash_table.h
@@ -319,6 +319,18 @@ int aws_hash_table_foreach(
     void *context);
 
 /**
+ * Compares two hash tables for equality. Both hash tables must have equivalent
+ * key comparators; values will be compared using the comparator passed into this
+ * function. The key hash function does not need to be equivalent between the
+ * two hash tables.
+ */
+AWS_COMMON_API
+bool aws_hash_table_eq(
+    const struct aws_hash_table *a,
+    const struct aws_hash_table *b,
+    bool (*value_eq)(const void *a, const void *b));
+
+/**
  * Removes every element from the hash map. destroy_fn will be called for
  * each element.
  */
--- a/source/hash_table.c
+++ b/source/hash_table.c
@@ -646,6 +646,37 @@ int aws_hash_table_foreach(
     return AWS_OP_SUCCESS;
 }
 
+bool aws_hash_table_eq(
+    const struct aws_hash_table *a,
+    const struct aws_hash_table *b,
+    bool (*value_eq)(const void *a, const void *b)) {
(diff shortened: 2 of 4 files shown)


source/hash_table.c
```diff
@@ -646,6 +646,37 @@ int aws_hash_table_foreach(
     return AWS_OP_SUCCESS;
 }
 
+bool aws_hash_table_eq(
+    bool (*value_eq)(const void *a, const void *b),
+    const struct aws_hash_table *a,
+    const struct aws_hash_table *b) {
+    if (aws_hash_table_get_entry_count(a) != aws_hash_table_get_entry_count(b)) {
```
ttjsu-aws: Do we want to add a check for the case when hash_table's are NULL?

bdonlan: Generally, we assume that arguments should be non-null except for destructors and for methods that document themselves as accepting null. I'm not sure that null == null is a condition we want to explicitly support here.

include/aws/common/hash_table.h
```diff
@@ -318,6 +318,18 @@ int aws_hash_table_foreach(
...
 
+/**
+ * Compares two hash tables for equality. Both hash tables must have equivalent
+ * key comparators; values will be compared using the comparator passed into this
+ * function. The key hash function does not need to be equivalent between the
+ * two hash tables.
+ */
+AWS_COMMON_API
+bool aws_hash_table_eq(
+    bool (*value_eq)(const void *a, const void *b),
```
ColdenCullen: Just curious, why is this the first parameter?

bdonlan: I'm happy with it being either first or last. Which do you think is more consistent?

source/hash_table.c
```diff
@@ -646,6 +646,37 @@ int aws_hash_table_foreach(
...
+        return false;
+    }
+
```
justinboswell: If, as the docs say, the key comparators must match, shouldn't that be verified here by comparing equals_fn on both?

include/aws/common/hash_table.h
ColdenCullen: My vote is last, looking at things like C++'s [`std::equal`](https://en.cppreference.com/w/cpp/algorithm/equal), the predicate is always last.

source/hash_table.c
bdonlan: We don't require that they are _equal_, just that they are _equivalent_. For example, we might pass in two FFI-generated bindings to the same lambda function in an interpreted language.

source/hash_table.c
ttjsu-aws: I see, thanks for stating the assumption.

PR #167, opened 2018-10-16
Added AWS_MIN/AWS_MAX

justinboswell:

### DIFF
--- a/include/aws/common/common.h
+++ b/include/aws/common/common.h
@@ -128,6 +128,9 @@ AWS_STATIC_ASSERT(sizeof(intptr_t) == sizeof(void *));
 AWS_STATIC_ASSERT(sizeof(char) == 1);
 #endif
 
+#define AWS_MIN(a, b) ((a) < (b) ? (a) : (b))
+#define AWS_MAX(a, b) ((a) > (b) ? (a) : (b))
+
 #ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


include/aws/common/common.h
```diff
@@ -128,6 +128,9 @@ AWS_STATIC_ASSERT(sizeof(intptr_t) == sizeof(void *));
 AWS_STATIC_ASSERT(sizeof(char) == 1);
 #endif
 
+#define AWS_MIN(a, b) ((a) < (b) ? (a) : (b))
+#define AWS_MAX(a, b) ((a) > (b) ? (a) : (b))
```
tautschnig: This may lead results not expected by the user if `a` or `b` have side effects or are `volatile` or otherwise shared variables. Other libraries would use code along the lines of `#define AWS_MIN(a, b) ({__typeof__(a) _a = a; __typeof__(b) _b = b; _a < _b ? _a : _b; })`

tautschnig:
I believe `clang-tidy` might even warn when using such code. The commit message doesn't provide any justification so let me ask the question: why is it desirable to have such macros introduced?

include/aws/common/common.h
justinboswell: That'll work for gcc and clang, lemme see what msvc wants for that.

justinboswell:
Actually, nevermind, discussed with team offline.

PR #169, opened 2018-10-18
Warn about implicit conversions that result in precision loss

marcomagdy:
*Description of changes:*

MSVC warns about those anyway with /W4. Having them pop up in POSIX
environments saves us finding out about the warning from CI.

### DIFF
--- a/source/hash_table.c
+++ b/source/hash_table.c
@@ -56,7 +56,7 @@ struct hash_table_state {
     size_t size, entry_count;
     size_t max_load;
     /* We AND a hash value with mask to get the slot index */
-    uint64_t mask;
+    size_t mask;
     double max_load_factor;
     /* actually variable length */
     struct hash_table_entry slots[1];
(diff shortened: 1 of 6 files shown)


source/hash_table.c
```diff
@@ -212,7 +212,7 @@ static int s_update_template_size(struct hash_table_state *template, size_t expe
     }
 
     template->size = size;
-    template->max_load = (size_t)(template->max_load_factor * template->size);
+    template->max_load = (size_t)(template->max_load_factor * (double)template->size);
```
marcomagdy: @bdonlan this change is making me a little uncomfortable.
I don't know how big the template->size can be, but if it gets big enough (more than 63 bits), I think this will be a bug when converting to a signed floating point.

I'm debating whether we should multiply the max_load_factor by 100 and cast that to size_t instead (and obviously divide template->size by 100 to even it out)
Thoughts?

PR #170, opened 2018-10-18
Add support for byte_cursors as hash_table keys

ColdenCullen:
This is useful when the value has a byte_cursor on it that is the key.


ColdenCullen:
This could for sure be a footgun. The use case that I'm working towards is trie, where each node has a byte cursor to the part of the MQTT topic it's responsible for, and the whole string is owned by each leaf node.

PR #172, opened 2018-10-19
Fix 'Overriding /W3 with /W4' warning on MSVC

ColdenCullen:


cmake/AwsCFlags.cmake
```diff
@@ -24,6 +24,12 @@ function(aws_set_common_properties target)
     cmake_parse_arguments(SET_PROPERTIES "${options}" "" "" ${ARGN})
 
     if(MSVC)
+        # Remove other /W flags
+        if(CMAKE_C_FLAGS MATCHES "/W3")
```
justinboswell: Do you want to search for /W{1-4} instead, just in case someone's gone mad with power?

ColdenCullen: Oh sure

PR #175, opened 2018-10-24
Include CMakeParseArguments module

marcomagdy:
The function cmake_parse_arguments() is defined natively in the newer
CMake versions and does not require an explicit include(). However,
CMake v3.4 and lower require explicitly including the module to use that
function.

The minimum required CMake version in aws-c-common is currently set to
v3.0


marcomagdy:
See https://cmake.org/cmake/help/v3.5/module/CMakeParseArguments.html

PR #178, opened 2018-10-29
Nuke Windows.h

ColdenCullen:
Windows.h is now included by exactly 0 headers.


ColdenCullen:
Looks like cppcheck is failing? But I'm not sure why, maybe it has a bad preproc?

include/aws/common/common.h
```diff
@@ -29,14 +29,11 @@
 #    define AWS_STATIC_IMPL static inline
 #endif
 
-#define AWS_STATIC_ASSERT0(cond, msg) typedef char static_assertion_##msg[(!!(cond)) * 2 - 1]
-#define AWS_STATIC_ASSERT1(cond, line) AWS_STATIC_ASSERT0(cond, static_assertion_at_line_##line)
+#define AWS_CONCAT(A, B) A ## B
+#define AWS_STATIC_ASSERT0(cond, msg) typedef char AWS_CONCAT(static_assertion_, msg)[(!!(cond)) * 2 - 1]
```
marcomagdy: This `AWS_CONCAT` is not needed.

include/aws/common/atomics_msvc.inl
```diff
@@ -65,8 +65,8 @@
  * this use case.
  */
 
-#pragma warning(push)
-#pragma warning(disable : 4201) /* nameless union */
+#include <Windows.h>
```
justinboswell: I think we would do well to just forward declare the WINAPI functions required, instead of pulling in Windows.h here, naked, and foisting it on users. They are very stable, and should be consistent across windows API revs.

include/aws/common/atomics.inl
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+#define AWS_ATOMICS_INLINE
+
+#include <aws/common/atomics.h>
+
+/* Helpers for converting aws_atomic_var into its parts. */
+#define AWS_ATOMIC_VAR_PTRVAL(var) (var->value)
+#define AWS_ATOMIC_VAR_INTVAL(var) (*(aws_atomic_impl_int_t *)var)
+
```
justinboswell: I'm not sure this is better than the union for readability or maintenance, what was the impetus for the change?

include/aws/common/windows/rw_lock.inl
```diff
@@ -18,9 +18,14 @@
 
 AWS_EXTERN_C_BEGIN
 
+#include <Windows.h>
```
justinboswell: Same as with atomics, forward declare and get rid of this include

tests/atomics_test.c
```diff
@@ -13,14 +13,21 @@
...
 #include <aws/common/common.h>
 #include <aws/common/condition_variable.h>
 #include <aws/common/mutex.h>
 #include <aws/common/thread.h>
 
 #include <aws/testing/aws_test_harness.h>
 
+#ifdef _WIN32
+#    include <malloc.h>
+#    define alloca _alloca
```
justinboswell: Beginning to wonder if we should just have a block of these in common.h, aws_alloca, aws_strncpy, etc. Gets around all the _s ridiculousness.

include/aws/common/byte_order.h
```diff
@@ -18,7 +18,9 @@
 
 #include <aws/common/common.h>
 
-#ifndef _WIN32
+#ifdef _WIN32
+#    include <stdlib.h>
+#else
```
graebm: ifdef on `_MSC_VER`, that's what the functions using _byteswap are doing

source/common.c
```diff
@@ -18,6 +18,10 @@
 #include <stdarg.h>
 #include <stdlib.h>
 
+#ifdef _WIN32
+#    include <Windows.h>
+#endif
```
graebm: the code that uses SecureZeroMemory branches on _MSC_VER

graebm:
This is great, btw

include/aws/common/atomics_msvc.inl
ColdenCullen: Turns out we don't need to! If we just prefix each function with `_`, it uses the compiler intrinsic and we don't need to include Windows.h OR forward declare. Because all Windows.h is:
```c
#define InterlockedExchangePointer _InterlockedExchangePointer
/* etc. */
```

include/aws/common/atomics.inl
ColdenCullen: We don't know how big the int type is here.

include/aws/common/windows/rw_lock.inl
ColdenCullen: Moved to .c file

source/common.c
ColdenCullen: Marco is right. _byteswap should be _MSC_VER, this should be _WIN32.

PR #182, opened 2018-11-02
Added -FPIC to non-windows compile targets

justinboswell:
Since all this code ends up in shared libs eventually, this avoids linker errors. -fPIC was being applied by CI scripts previously.


cmake/AwsCFlags.cmake
```diff
@@ -56,6 +56,9 @@ function(aws_set_common_properties target)
         # Without -Wno-sign-conversion, this becomes too noisy.
         list(APPEND AWS_C_FLAGS -Wconversion -Wno-sign-conversion)
 
+	# Always enable position independent code, since this code will always end up in a shared lib
+	list(APPEND AWS_C_FLAGS -fPIC)
```
ColdenCullen: wHiTeSpAcE

PR #195, opened 2018-11-20
Enable -fsanitize=integer by default

ColdenCullen:
This should help us catch integer value change and overflow.

### DIFF
--- a/cmake/AwsSanitizers.cmake
+++ b/cmake/AwsSanitizers.cmake
@@ -47,7 +47,7 @@ endfunction()
 
 # This function enables sanitizers on the given target
 # Options:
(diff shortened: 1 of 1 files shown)


tautschnig:
What is the performance cost of adding sanitizers? (If I understand correctly that this is enabled in production, and not only for tests.)

ColdenCullen:
> What is the performance cost of adding sanitizers?
Very low. ["The checks have small runtime cost and no impact on address space layout or ABI"](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html).

> (If I understand correctly that this is enabled in production, and not only for tests.)
This is false, they are only enabled when `-DENABLE_SANITIZERS=ON` is passed, which our team only does in CI.

PR #201, opened 2018-11-27
Fix build on ARM

bdonlan:
*Issue #, if available:*

*Description of changes:*

Tested on an a1.xlarge instance :)

### DIFF
--- a/include/aws/common/math.h
+++ b/include/aws/common/math.h
@@ -21,14 +21,55 @@
 
 #include <stdlib.h>
 
-#if defined(AWS_HAVE_GCC_OVERFLOW_MATH_EXTENSIONS)
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if defined(AWS_HAVE_GCC_OVERFLOW_MATH_EXTENSIONS) && (defined(__clang__) || !defined(__cplusplus))
+/*
+ * GCC and clang have these super convenient overflow checking builtins...
+ * but (in the case of GCC) they're only available when building C source.
+ * We'll fall back to one of the other inlinable variants (or a non-inlined version)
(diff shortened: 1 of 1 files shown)


include/aws/common/math.fallback.inl
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/exports.h>
+#include <stdint.h>
+#include <stdlib.h>
```
ColdenCullen: Because all of these inl files are included by math.h, which already includes common.h, and stdint.h and stdlib.h, you could probably just remove all of these includes.

This will also help in the case we're on a hella old GCC install without stdint.h, in which case we define our own int types in common.h

bdonlan: This is mostly to make my editor shut up and stop redlining all the uint64_ts.

bdonlan: I'll replace with common.h though

include/aws/common/math.h
```diff
@@ -25,179 +16,61 @@
...
-
-    return out == 0;
-#    else
-#        error "Not supported."
-#    endif
-}
+#include <stdlib.h>
+
+#ifdef __cplusplus
+extern "C" {
```
ColdenCullen: [optional] you can replace this whole block with `AWS_EXTERN_C_BEGIN` and the close with `AWS_EXTERN_C_END`

PR #203, opened 2018-11-28
fix logic bug in byte swap functions

david-koenig:
which only would have broken on big endian windows machine,
if such a thing exists.


david-koenig:
The one failure in CI was not related to this code change: conditional_wait_timeout test. Merging in latest from master which should redo the CI anyway.

PR #205, opened 2018-11-28
Date time implementation

JonathanHenson:


JonathanHenson:
This pull request **introduces 1 alert** when merging 73c3fb9542208497c2cf6d2ffb23e386fbf6da0b into 75acca41eb3d4204d6cb433d95472299a5c7bb6b - [view on LGTM.com](https://lgtm.com/projects/g/awslabs/aws-c-common/rev/pr-84ac6e1f23108bb134e453756a04f057aaccdc63)

**new alerts:**

* 1 for Comparison result is always the same

---

*Comment posted by [LGTM.com](https://lgtm.com)*

source/date_time.c
```diff
@@ -0,0 +1,826 @@
...
+    int state = 0;
+    bool error = false;
+
+    while (state <= finalState && !error && index < len) {
+        char c = date_str->buffer[index];
+
+        switch (state) {
+            case 0:
+                if (c == ',') {
+                    int weekNumber =
```
justinboswell: standards: week_number/week_num

source/date_time.c
```diff
@@ -0,0 +1,826 @@
...
+                            return -1;
+                    }
+                default:
+                    return -1;
+            }
+        case 'T':
+        case 't':
+            next = time_string[++index];
+            switch (next) {
+                case 'H':
```
bdonlan: Despite the comment above, this nested switch may require up to nine comparisons:

* 4 comparisons for the outer switch (as a binary search of 10 elements)
* 3 comparisons for the middle switch (as a binary search of 4 elements)
* Two comparisons for the innermost switch

It could also be implemented as three switch tables, or some combination of switchtables and binary searches, at the expense of L1, but exactly three comparisons is an unlikely compilation result.

Anyway - writing this kind of code by hand is a bit of a waste of effort, and quite hard to debug. I would recommend one of:

* using a simple strcmp sequence
* doing a hash table lookup
* using a lexer generator like re2c
* using a perfect hash generator like gperf, if you really want the maximum possible performance

source/date_time.c
```diff
@@ -0,0 +1,826 @@
...
+#include <aws/common/time.h>
+
+#include <ctype.h>
+#include <math.h>
+
+static const char *RFC822_DATE_FORMAT_STR_MINUS_Z = "%a, %d %b %Y %H:%M:%S GMT";
+static const char *RFC822_DATE_FORMAT_STR_WITH_Z = "%a, %d %b %Y %H:%M:%S %Z";
+static const char *ISO_8601_LONG_DATE_FORMAT_STR = "%Y-%m-%dT%H:%M:%SZ";
+
+static int get_week_day_number_from_str(const char *time_string, size_t startIndex, size_t stopIndex) {
```
bdonlan: Do we really need to parse the day of the week? Isn't it redundant?

source/date_time.c
```diff
@@ -0,0 +1,826 @@
...
+    if (stopIndex - startIndex < 3) {
+        return -1;
+    }
+
+    size_t index = startIndex;
+
+    char c = time_string[index];
+    char next = 0;
+
+    /* it's ugly but this should compile down to EXACTLY 3 comparisons and no memory allocations */
```
bdonlan: Consider using a lexer generator, like [re2c](http://re2c.org/).

You could also use a single switch table by doing:

    switch ((int)(uint8_t)tolower(time_string[0]) | ((int)(uint8_t)tolower(time_string[1]) << 8) | ((int)(uint8_t)tolower(time_string[2]) << 16)

source/date_time.c
```diff
@@ -0,0 +1,826 @@
...
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+    }
+
+    memcpy(output_buf->buffer, formatted_string, len);
+    output_buf->len = len;
+
+    return AWS_OP_SUCCESS;
+}
+
+int aws_date_time_to_utc_time_str(struct aws_date_time *dt, enum aws_date_format fmt, struct aws_byte_buf *output_buf) {
```
justinboswell: Don't we usually put output/destination buffers as the first argument?

source/date_time.c
```diff
@@ -0,0 +1,826 @@
...
+    if (dt->utc_assumed) {
+        dt->timestamp = aws_mk_gm_time(&parsed_time);
+    } else {
+        dt->timestamp = mktime(&parsed_time);
+    }
+
+    return AWS_OP_SUCCESS;
+}
+
+int aws_date_time_to_local_time_str(
```
justinboswell: Don't we usually put output/destination buffers as the first argument?

source/date_time.c
```diff
@@ -0,0 +1,826 @@
...
+        return -1;
+    }
+
+    size_t index = startIndex;
+
+    char c = time_string[index];
+    char next = 0;
+
+    /* it's ugly but this should compile down to EXACTLY 3 comparisons and no memory allocations */
+    switch (c) {
```
bdonlan: See above suggestions - don't write triply-nested switches by hand.

If you don't want to use code generators, and still want a single switch, consider:

```
#define STR_TRIPLET_TO_INDEX(str) \
  (   ((uint32_t)(uint8_t)tolower((str)[0]) << 0) | \
      ((uint32_t)(uint8_t)tolower((str)[1]) << 8) | \
      ((uint32_t)(uint8_t)tolower((str)[2]) << 16) )

// ...
if (!time_string[0] || !time_string[1] || !time_string[2]) return -1;

switch(STR_TRIPLET_TO_INDEX(str)) {
  case STR_TRIPLET_TO_INDEX("jan"): return 0;
  case STR_TRIPLET_TO_INDEX("feb"): return 1;
// ...
  default: return -1;
}
```

Note that I'm not sure if STR_TRIPLET_TO_INDEX will work in the case labels, but you should be able to precompute it.

source/date_time.c
```diff
@@ -0,0 +1,826 @@
...
+
+uint64_t aws_date_time_as_nanos(struct aws_date_time *dt) {
+    return (uint64_t)dt->timestamp * AWS_TIMESTAMP_NANOS;
+}
+
+uint64_t aws_date_time_as_millis(struct aws_date_time *dt) {
+    return (uint64_t)dt->timestamp * AWS_TIMESTAMP_MILLIS;
+}
+
+struct tm s_get_time_struct(struct aws_date_time *dt, bool local_time) {
```
justinboswell: This will probably optimize, but to a memset() of the return value, so why not pass the struct tm in as an out param?

source/date_time.c
```diff
@@ -0,0 +1,826 @@
...
+static int s_parse_iso_8601(const struct aws_byte_buf *date_str, struct tm *parsed_time) {
+    size_t index = 0;
+    size_t state_start_index = 0;
+    const int final_state = 7;
+    int state = 0;
+    bool error = false;
+
+    while (state <= final_state && !error && index < date_str->len) { /* lgtm [cpp/constant-comparison] */
+        char c = date_str->buffer[index];
+        switch (state) {
```
bdonlan: Please use a lexer generator. Validating these state machines by hand is not trivial.

source/date_time.c
```diff
@@ -0,0 +1,826 @@
...
+    }
+
+    struct tm time;
+    AWS_ZERO_STRUCT(time);
+    aws_local_time(dt->timestamp, &time);
+    char formatted_string[100];
+    AWS_ZERO_ARRAY(formatted_string);
+
+    if (fmt == AWS_DATE_FORMAT_RFC822) {
+        strftime(formatted_string, sizeof(formatted_string), RFC822_DATE_FORMAT_STR_WITH_Z, &time);
```
bdonlan: Check the return value of strftime.

source/date_time.c
```diff
@@ -0,0 +1,826 @@
...
+        aws_local_time(dt->timestamp, &time);
+    } else {
+        aws_gm_time(dt->timestamp, &time);
+    }
+
+    return time;
+}
+
+uint16_t aws_date_time_year(struct aws_date_time *dt, bool local_time) {
+    struct tm time = s_get_time_struct(dt, local_time);
```
bdonlan: So if I want to read out the components of the aws_date_time struct, I have to convert from the timestamp to the gmtime/localtime structure eight times?

include/aws/common/date_time.h
```diff
@@ -0,0 +1,118 @@
...
+    AWS_DATE_DAY_OF_WEEK_MONDAY,
+    AWS_DATE_DAY_OF_WEEK_TUESDAY,
+    AWS_DATE_DAY_OF_WEEK_WEDNESDAY,
+    AWS_DATE_DAY_OF_WEEK_THURSDAY,
+    AWS_DATE_DAY_OF_WEEK_FRIDAY,
+    AWS_DATE_DAY_OF_WEEK_SATURDAY,
+};
+
+struct aws_date_time {
+    time_t timestamp;
```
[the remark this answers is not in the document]
JonathanHenson: time_t is size and if I stored nanos, I'd have to constantly be converting back and forth before doing anything with it.

tests/date_time_test.c
```diff
@@ -0,0 +1,326 @@
...
+    (void)allocator;
+    (void)ctx;
+
+    struct aws_date_time date_time;
+    const char *date_str = "Wed, 02 Oct 2002 08:05:09 GMT";
+    struct aws_byte_buf date_buf = aws_byte_buf_from_c_str(date_str);
+
+    ASSERT_SUCCESS(aws_date_time_init_from_str(&date_time, &date_buf, AWS_DATE_FORMAT_RFC822));
+    ASSERT_INT_EQUALS(AWS_DATE_DAY_OF_WEEK_WEDNESDAY, aws_date_time_day_of_week(&date_time, false));
+    ASSERT_UINT_EQUALS(02, aws_date_time_month_day(&date_time, false));
```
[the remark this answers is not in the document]
JonathanHenson: that was just a typing mistake, will fix.

source/date_time.c
JonathanHenson: will fix

source/date_time.c
JonathanHenson: output is usually the last argument.

source/date_time.c
JonathanHenson: last argument, which it is here.

source/date_time.c
JonathanHenson: I don't want to pull in a dependency for this,
your second suggestion looks promising though.

source/date_time.c
JonathanHenson: will try it out.

source/date_time.c
JonathanHenson: see above.... I'm not pulling in a dependency for this.

source/date_time.c
JonathanHenson: I suppose we could stuff the tm into the aws_date_time struct.

include/aws/common/date_time.h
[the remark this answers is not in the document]
bdonlan: time_t can be 32 bits in some configurations. We should explicitly use a 64-bit type.

source/date_time.c
bdonlan: For matching a fixed set of strings, strcmp is way simpler:

```
char day_buf[4] = {0};
strncpy(day_buf, time_string, sizeof(day_buf) - 1);
for (size_t i = 0; i < 3; i++) day_buf[i] = tolower(day_buf[i]);

if (!strcmp(time_string, "mon")) return 1;
if (!strcmp(time_string, "tue")) return 2;
// ...
```

include/aws/common/date_time.h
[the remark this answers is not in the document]
JonathanHenson: how will that help, if the underlying libc calls only accept time_t? I'd still have to cast it down and the error would still surface. We had to do this for android, I guess we could add overflow errors as possible outputs. Any better suggestions?

source/date_time.c
bdonlan: `mktime`, `timegm`, and `timelocal` ignore the original value of `tm_wday` and `tm_yday`, and set them to the correct values on return. So you don't necessarily need to actually parse this part.

source/date_time.c
JonathanHenson: Yeah, I think it's often simpler inside a single branch of a state machine, just not for the whole thing. I'll see what I can do on it.

source/date_time.c
bdonlan: Then at least write it in a more readable way. With a few token-extraction helpers you can write simple straight-line code to do the parse, something like this:

```
char tmpbuf[16];

if (!extract_token(tmpbuf, sizeof(tmpbuf), &input_ptr) {
  return -1;
}
day_of_week = parse_day_of_week(tmpbuf);

if (*input_ptr != ',') return -1;
input_ptr++;

while (*input_ptr == ' ') input_ptr++;

if (!extract_int(&day_of_month, &input_ptr)) {
  return -1;
}
// ...
```

source/date_time.c
JonathanHenson: good call, will remove then. I do get nervous though, because I've been bitten by platforms like windows not fully implementing the spec for these functions (hence the reason we aren't using the built in date parsers).

source/date_time.c
JonathanHenson: will do.

source/date_time.c
JonathanHenson: hmm. for this to work we'd need to store two tm, one for utc and one for localtime... not that big a deal though I guess.

include/aws/common/date_time.h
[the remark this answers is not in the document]
bdonlan: Remember that our nanosecond timestamps are much larger than time_t millisecond timestamps. The current nanosecond timestamp already does not fit in a 32-bit field.

source/date_time.c
bdonlan: Another option for this would be something like:

```
uint32_t tmp = 0;
strncpy(&tmp, str, 3);

// This AND forces the string to lowercase (assuming ASCII)
switch (aws_ntohl(tmp) & ~0x20202020) {
  // ...
}
```

This might be a bit faster, as it avoids having all those individual tolower calls. If you can guarantee that str is at least 4 bytes long then a memcpy (and mask of the last byte) would be even better.

source/date_time.c
JonathanHenson: removed

source/date_time.c
JonathanHenson: I wrote it, and I hate it. It's not clear, and it doesn't have the advantage that an FSM has, which is it's strict by default. I'm going to keep these. It's a well established pattern and is what a regex or something would compile down to anyways.

JonathanHenson: https://paste.amazon.com/show/henso/1543545571

source/date_time.c
justinboswell: That seems like a better move, it would do away with some of the other precision issues and structure copying going on.

PR #206, opened 2018-11-30
Add a string copy function

david-koenig:

### DIFF
--- a/source/string.c
+++ b/source/string.c
@@ -19,18 +19,29 @@ struct aws_string *aws_string_new_from_c_str(struct aws_allocator *allocator, co
 }
 
 struct aws_string *aws_string_new_from_array(struct aws_allocator *allocator, const uint8_t *bytes, size_t len) {
-    struct aws_string *hdr = aws_mem_acquire(allocator, sizeof(struct aws_string) + len + 1);
-    if (!hdr) {
+    struct aws_string *str = aws_mem_acquire(allocator, sizeof(struct aws_string) + len + 1);
+    if (!str) {
         return NULL;
     }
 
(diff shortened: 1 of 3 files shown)


source/string.c
```diff
@@ -19,18 +19,29 @@ struct aws_string *aws_string_new_from_c_str(struct aws_allocator *allocator, co
...
+    *(struct aws_allocator **)(&str->allocator) = allocator;
+    *(size_t *)(&str->len) = len;
+    memcpy((void *)str->bytes, bytes, len);
+    *(uint8_t *)&str->bytes[len] = '\0';
 
-    return hdr;
+    return str;
+}
+
+struct aws_string *aws_string_copy(struct aws_allocator *allocator, const struct aws_string *str) {
```
bdonlan: Better to call this _clone or _dup? _copy (without an _init or _new) implies to me that it's copying to an existing destination

source/string.c
```diff
@@ -19,18 +19,29 @@ struct aws_string *aws_string_new_from_c_str(struct aws_allocator *allocator, co
...
+    size_t size = sizeof(struct aws_string) + str->len + 1;
```
bdonlan: Delegate to aws_string_new_from_array.

source/string.c
ColdenCullen: I like _dup, it matches what C's `strdup`.

david-koenig: We settled on ``aws_string_new_from_string`` based on conversation with the team.

source/string.c
david-koenig: Done.

PR #208, opened 2018-12-06
optimization: don't allocate a new string on static strings

david-koenig:
Detects if a string being duplicated is static and if so just continues to use the static string as the duplicate. Works because aws_string_destroy is a no-op on static strings.

### DIFF
--- a/source/string.c
+++ b/source/string.c
@@ -34,7 +34,10 @@ struct aws_string *aws_string_new_from_array(struct aws_allocator *allocator, co
 }
 
 struct aws_string *aws_string_new_from_string(struct aws_allocator *allocator, const struct aws_string *str) {
-    return aws_string_new_from_array(allocator, str->bytes, str->len);
+    if (str->allocator) return aws_string_new_from_array(allocator, str->bytes, str->len);
+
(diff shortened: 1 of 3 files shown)


source/string.c
```diff
@@ -34,7 +34,10 @@ struct aws_string *aws_string_new_from_array(struct aws_allocator *allocator, co
 }
 
 struct aws_string *aws_string_new_from_string(struct aws_allocator *allocator, const struct aws_string *str) {
-    return aws_string_new_from_array(allocator, str->bytes, str->len);
+    if (str->allocator) return aws_string_new_from_array(allocator, str->bytes, str->len);
```
bdonlan: Hmm... Would this be surprising behavior for a caller? We should clearly document it at a minimum.

ColdenCullen:
This seems dangerous. What if someone wants to copy a string so that they may modify the new one?

david-koenig:
Ended up putting this optimization into the package where I'm using it, rather than in aws-c-common. Still not certain that it wouldn't be better in aws-c-common, but I don't feel like fighting that battle, so I'm closing this one out.

PR #209, opened 2018-12-11
UUID, device_random, and call_once

JonathanHenson:


source/uuid.c
```diff
@@ -0,0 +1,110 @@
...
+#endif
+
+int aws_uuid_init(struct aws_uuid *uuid) {
+    struct aws_byte_buf buf = {
+        .buffer = uuid->uuid_data,
+        .len = 0,
+        .capacity = sizeof(uuid->uuid_data),
+        .allocator = NULL,
+    };
+    return aws_device_random_buffer(&buf);
```
justinboswell: using urandom isn't actually going to be globally unique without a machine id (MAC addr or similar, MACs are lies in the cloud) + timestamp + monotonic counter. Do we care?

See: https://github.com/aws/lumberyard/blob/master/dev/Code/Framework/AzCore/AzCore/Component/Entity.cpp#L1420 for a possibly imperfect, but so far never collided implementation :)

tests/device_random_test.c
```diff
@@ -0,0 +1,107 @@
...
+static int s_device_rand_u64_fn(struct aws_allocator *allocator, void *ctx) {
+    (void)allocator;
+    (void)ctx;
+
+    uint64_t last_value = 0;
+
+    for (size_t i = 0; i < 100000; ++i) {
+        uint64_t next_value = 0;
+        ASSERT_SUCCESS(aws_device_random_u64(&next_value));
+        ASSERT_FALSE(next_value == last_value);
```
justinboswell: It is perfectly valid for the same number to come twice in sequence, even if astronomically unlikely. Also, you'd want to do ASSERT_NOT_EQUAL() here so we can see the values.

A better test might be to make a set and see if it is like 95% unique or something like that.

include/aws/common/uuid.h
```diff
@@ -0,0 +1,38 @@
...
+};
+
+/* 36 bytes for the UUID plus one more for the null terminator. */
+#define AWS_UUID_STR_LEN 37
+
+AWS_EXTERN_C_BEGIN
+
+AWS_COMMON_API int aws_uuid_init(struct aws_uuid *uuid);
+AWS_COMMON_API int aws_uuid_init_from_str(struct aws_uuid *uuid, struct aws_byte_cursor *uuid_str);
+AWS_COMMON_API int aws_uuid_to_str(struct aws_uuid *uuid, struct aws_byte_buf *output);
```
justinboswell: How about aws_uuid_equals(struct aws_uuid *a, struct aws_uuid *b) or similar? Would make it much easier to use in hashes and such

include/aws/common/thread.h
```diff
@@ -15,11 +15,12 @@
  * express or implied. See the License for the specific language governing
  * permissions and limitations under the License.
  */
-
 #include <aws/common/common.h>
 
 #ifndef _WIN32
 #    include <pthread.h>
+#else
+#    include <windows.h>
```
ColdenCullen: NONONONONONO

It should be pretty easy to F12 on INIT_ONCE and see what it actually is. It's probably a void *.

include/aws/common/thread.h
```diff
@@ -32,6 +33,15 @@ struct aws_thread_options {
     size_t stack_size;
 };
 
+#ifdef _WIN32
+typedef INIT_ONCE aws_thread_once;
+#        define AWS_THREAD_ONCE_INIT                                                                                   \
+            { NULL }
```
ColdenCullen: `INIT_ONCE_STATIC_INIT`

include/aws/common/uuid.h
```diff
@@ -0,0 +1,38 @@
...
+    uint8_t uuid_data[16];
+};
+
+/* 36 bytes for the UUID plus one more for the null terminator. */
+#define AWS_UUID_STR_LEN 37
+
+AWS_EXTERN_C_BEGIN
+
+AWS_COMMON_API int aws_uuid_init(struct aws_uuid *uuid);
+AWS_COMMON_API int aws_uuid_init_from_str(struct aws_uuid *uuid, struct aws_byte_cursor *uuid_str);
```
ColdenCullen: `uuid_str` can be `const`.

source/uuid.c
JonathanHenson: The probability of a clash from urandom should be lower than a clash from cloned Mac addresses with a matching time stamp.

include/aws/common/uuid.h
ColdenCullen: `uuid` can be const.

JonathanHenson: I like this idea

source/uuid.c
```diff
@@ -0,0 +1,110 @@
...
+#include <aws/common/uuid.h>
+
+#include <aws/common/byte_buf.h>
+#include <aws/common/device_random.h>
+
+#include <inttypes.h>
+#include <stdio.h>
+
+#define UUID_FORMAT                                                                                                    \
+    "%02" SCNx8 "%02" SCNx8 "%02" SCNx8 "%02" SCNx8 "-%02" SCNx8 "%02" SCNx8 "-%02" SCNx8 "%02" SCNx8 "-%02" SCNx8     \
```
ColdenCullen: Maybe macro out "%02" too?

source/uuid.c
```diff
@@ -0,0 +1,110 @@
...
+#    pragma warning(disable : 4996)
+#endif
+
+int aws_uuid_init(struct aws_uuid *uuid) {
+    struct aws_byte_buf buf = {
+        .buffer = uuid->uuid_data,
+        .len = 0,
+        .capacity = sizeof(uuid->uuid_data),
+        .allocator = NULL,
+    };
```
ColdenCullen: This could be: `aws_byte_buf_from_empty_array(uuid->uuid_data, sizeof(uuid->uuid_data));`

source/device_random.c
```diff
@@ -0,0 +1,68 @@
...
+#endif
+
+int aws_device_random_u64(uint64_t *output)
+{
+    struct aws_byte_buf buf = {
+        .buffer = (uint8_t *)output,
+        .len = 0,
+        .capacity = sizeof(uint64_t),
+        .allocator = NULL,
+    };
```
ColdenCullen: These could all be: `aws_byte_buf_from_empty_array((uint8_t*)output, sizeof(*output));`

tests/device_random_test.c
JonathanHenson: There is no ASSERT_NOT_EQUAL at the moment, I don't have a set structure to use ( I suppose we could just use the hash table but that's a lot of work when really this is just a sanity check to make sure we're actually generating random values). Whatever test we do will be probablistic anyways so I'd prefer to keep it as is for the moment.

include/aws/common/uuid.h
JonathanHenson: done

include/aws/common/thread.h
justinboswell: 👍

tests/device_random_test.c
justinboswell: Ok, fair enough

JonathanHenson:
Overriding colden at the moment since I’ve addressed his feedback and he’s indisposed at the moment.

PR #212, opened 2018-12-13
Implement Semaphore, C++ style rw_lock

ColdenCullen:


include/aws/common/rw_lock.h
```diff
@@ -37,7 +43,11 @@ struct aws_rw_lock {
         { .lock_handle = NULL }
 #else
 #    define AWS_RW_LOCK_INIT                                                                                           \
-        { .lock_handle = PTHREAD_RWLOCK_INITIALIZER }
+        {                                                                                                              \
+            .readers = AWS_ATOMIC_INIT_INT(0), .holdouts = AWS_ATOMIC_INIT_INT(0),                                     \
+            .reader_sem = AWS_SEMAPHORE_INIT(0, INT), .writer_sem = AWS_SEMAPHORE_INIT(0, 1),                          \
```
bdonlan: What is the `INT` constant?

source/posix/rw_lock.c
```diff
@@ -12,48 +12,103 @@
  * express or implied. See the License for the specific language governing
```
bdonlan: _Please_ document the protocol being implemented for this rwlock. It's non-obvious what is going on here.

source/posix/rw_lock.c
```diff
@@ -12,48 +12,103 @@
...
 }
 
 int aws_rw_lock_rlock(struct aws_rw_lock *lock) {
 
-    return aws_private_convert_and_raise_error_code(pthread_rwlock_rdlock(&lock->lock_handle));
+    const reader_count_t num_readers = aws_atomic_fetch_add(&lock->readers, 1) + 1;
+    if (num_readers < 0) {
+        aws_semaphore_acquire(&lock->reader_sem);
+    }
+    assert(aws_atomic_load_int(&lock->readers) > 0);
```
bdonlan: Problematic race condition:

* R: aws_rw_lock_rlock
* R: lock->readers += 1; num_readers = 1;
* W: aws_rw_lock_wlock
* W: lock->readers -= s_max_readers; lock->readers = 1 - INT64_MAX
* R: assert(lock->readers > 0) // assert fail

source/posix/rw_lock.c
```diff
@@ -12,48 +12,103 @@
...
 int aws_rw_lock_wunlock(struct aws_rw_lock *lock) {
 
-    return aws_private_convert_and_raise_error_code(pthread_rwlock_unlock(&lock->lock_handle));
+    assert(aws_atomic_load_int(&lock->holdouts) == 0);
+
+    const reader_count_t current = aws_atomic_fetch_add(&lock->readers, s_max_readers) + s_max_readers;
+    assert(current >= 0);
+
+    for (reader_count_t i = 0; i < current; ++i) {
+        aws_semaphore_release_one(&lock->reader_sem);
```
bdonlan: Why not introduce a release_multiple API to avoid reacquiring the internal semaphore mutex multiple times?

source/posix/rw_lock.c
ColdenCullen: Oh you're right. These asserts were useful in testing but I think I should remove them before more complex cases cause problems.

include/aws/common/rw_lock.h
ColdenCullen: It would be a typo :smile:

source/posix/rw_lock.c
ColdenCullen: That's a good idea, and shouldn't be too hard.

source/posix/rw_lock.c
bdonlan: This highlights the need for documentation; if the asserts are wrong, it suggests that your mental model when writing this code was wrong, and so a deeper analysis is needed to ensure that this is correct.

Have you built a TLA+ model for this rwlock protocol?

bdonlan: Also, do you have a test that hammers this with concurrent operations?

ColdenCullen: The rw_lock tests should hit this pretty hard.

As for the asserts, I simply added them to validate my understandings of how this code works in a single-threaded context. The implementation is identical to the one in the [C++ SDK](https://github.com/aws/aws-sdk-cpp/blob/master/aws-cpp-sdk-core/source/utils/threading/ReaderWriterLock.cpp).

bdonlan: The behavior of the code in a single-threaded context is not all that interesting for multithreaded code - and the implementation is far from identical.

Again, you're writing a nontrivial synchronization protocol here, what have we done to document the assumptions and invariants of this protocol, and what have we done to verify correctness?

ColdenCullen: Disagree, if you inline all C++ operators that are obviously not available here, the code should be functionally the same. If you spot any discrepancies, please call them out.

Please see docs added since your review.

include/aws/common/rw_lock.h
```diff
@@ -21,29 +21,40 @@
...
 #ifdef _WIN32
 #    define AWS_RW_LOCK_INIT                                                                                           \
         { .lock_handle = NULL }
 #else
+/* -1 ^ (1ull << (sizeof(size_t) * 8 - 1)): SSIZE_MAX but without guaranteeing SSIZE_MAX exists. */
 #    define AWS_RW_LOCK_INIT                                                                                           \
-        { .lock_handle = PTHREAD_RWLOCK_INITIALIZER }
+        {                                                                                                              \
+            .readers = AWS_ATOMIC_INIT_INT(0), .holdouts = AWS_ATOMIC_INIT_INT(0),                                     \
+            .reader_sem = AWS_SEMAPHORE_INIT(0, -1 ^ (1ull << (sizeof(size_t) * 8 - 1))),                              \
```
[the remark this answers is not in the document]
ColdenCullen: This isn't true, because I'm always casting the value to a signed type, and we need to be able to subtract max_readers without underflowing the value. The max readers must NOT be the unsigned max, it must be the signed max or else we would violate this requirement. You can verify this by changing it to SIZE_MAX locally, where you'll notice the tests deadlock.

I can just branch on SIZE_MAX == UINT32/64_MAX though, that's a good idea.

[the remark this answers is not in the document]
bdonlan: Do we ever read the semaphore's max count? I agree that changing `s_max_readers` to SIZE_MAX would break, but I don't think the semaphore limit should impact anything.

source/posix/rw_lock.c
```diff
@@ -12,48 +12,108 @@
...
 
 int aws_rw_lock_wlock(struct aws_rw_lock *lock) {
 
-    return aws_private_convert_and_raise_error_code(pthread_rwlock_wrlock(&lock->lock_handle));
-}
+    aws_mutex_lock(&lock->writer_lock);
 
-int aws_rw_lock_try_rlock(struct aws_rw_lock *lock) {
+    const reader_count_t num_readers = aws_atomic_fetch_sub(&lock->readers, s_max_readers);
+    if (num_readers > 0) {
```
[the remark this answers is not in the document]
ColdenCullen: C++ actually checks > 0 as well (see [line 52](https://github.com/aws/aws-sdk-cpp/blob/master/aws-cpp-sdk-core/source/utils/threading/ReaderWriterLock.cpp#L52)), `if(const auto current = m_readers.fetch_sub(MaxReaders))` is only takes the true branch if current (`num_readers` in my parlance) is > 0.

include/aws/common/rw_lock.h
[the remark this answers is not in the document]
ColdenCullen: They do need to be the same, because otherwise rlock would let more readers in than max readers, so when wlock subtracts max readers, the number would still be positive, breaking all sorts of logic.

source/posix/rw_lock.c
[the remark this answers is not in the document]
bdonlan: Looks like a != 0 check to me?

source/semaphore.c
```diff
@@ -74,7 +74,10 @@ void aws_semaphore_release(struct aws_semaphore *semaphore, size_t num_resources
     if (semaphore->count > semaphore->max_count) {
         semaphore->count = semaphore->max_count;
     }
-    aws_condition_variable_notify_one(&semaphore->sync_point);
+
+    for (size_t i = 0; i < num_resources; ++i) {
+        aws_condition_variable_notify_one(&semaphore->sync_point);
```
bdonlan: I wonder if just using a mutex instead of a semaphore to wake up readers would be better? It would avoid the need to loop over notify_one.

bdonlan: This would be a different protocol at this point. A sketch:

```
struct rwlock {
  // If negative, a write lock is pending; readers = (remaining readers) - max_readers
  // If non-negative, no write lock is pending; readers = number of read locks
  atomic readers = 0;
  mutex mut_writer;
  // True if a writer is active; this blocks other writers from entering while 
  // a write lock is pending
  bool writer_active;
  // signalled when a write lock is released to allow readers to proceed
  cvar release_readers;
  // signalled when the last reader leaves the lock while a write is pending
  cvar reader_to_writer;
  // signalled once when a write lock is released to allow one pending writer to proceed
  cvar writer_to_writer;
};

read_lock:
  loop {
    compare_exchange: if readers >= 0 { readers++; return; }
    lock (mut_writer) {
      while (readers < 0) { await release_readers; }
    }
 }

read_unlock:
    new_val = readers--;
    if (new_val == -max_readers) {
      // this lock is critical to avoid a race where the writer is just before the sleep when we update readers
      lock (mut_writer) { cvar_awake(reader_to_writer); }
    }

write_lock:
  lock(mut_writer);
  while (writer_active) { await(writer_to_writer); }
  writer_active = true;
  atomic: readers -= max_readers;
  while (readers > -max_readers) {
    await(reader_to_writer);
  }

write_unlock:
  writer_active = false;
  readers += max_readers;
  cvar_awake_all(release_readers);
  cvar_awake(writer_to_writer);
  unlock(mut_writer);  

```

ColdenCullen: Okay so maybe I just don't understand the pseudo-code syntax, but here are my thoughts so far:
1. If there's an open reader, and a new reader and a new writer come along, mut_writer is locked by write_lock and read_lock, as far as I can tell it's up to the scheduler whether the next holder of the lock will be a writer or the readers.
2. If readers are open and someone calls write_lock, acquiring mut_writer, how would read_unlock acquire mut_writer to notify the cvar?
3. I don't at all understand the read_lock function, is that supposed to be like:
```c
while (true) {
    num_readers = aws_atomic_load_int(&lock->readers);
    if (num_readers >= 0) {
        if (aws_atomic_compare_exchange_int(&lock->readers, (size_t *)&num_readers, num_readers + 1)) {
            return success;
        }
    } else {
        lock(mut_writer);
        cvar_wait_pred(release_readers, [&]() { aws_atomic_load_int(&lock->readers) >= 0; });
    }
}
```
Or am I missing something?

I'm sure there's a lot more going on here, tomorrow I'll take a crack at implementing this and see how it goes.

ColdenCullen:
😢

Issue #213, opened 2018-12-17
Packaging Challenges


psalvaggio:
Hello, it looks like the AWS C++ SDK depends on this library, so I'm trying to package this library so I can continue to package that library. Currently, this library has a couple challenges to packaging.

1. When building shared libraries, only a .so file is output (normally there's symlinks with the API/ABI versions). CMake let's you do this with `set_target_properties()` with `VERSION` and `SOVERSION`. Normally the `.so` link goes in the development package, while the other two go into the runtime package.
2. On Fedora, the libraries are going to /usr/lib instead of /usr/lib64. This can be solved by using the GNUInstallDirs package in CMake.
3. This file: /usr/include/aws/testing/aws_test_harness.h is getting installed when perhaps it doesn't need to be?
4. The CMake config install is a little funky. Currently it is installing the files listed below. Would it be possible to dump everything into a `$LIBDIR/cmake/aws-c-common` directory so there's no risk of conflicts with other packages? Are all of these files required or just the two config files?
```
-- Installing: /builddir/build/BUILDROOT/aws-c-common-0.1-1.x86_64/usr/lib/aws-c-common/cmake/aws-c-common-config.cmake
-- Installing: /builddir/build/BUILDROOT/aws-c-common-0.1-1.x86_64/usr/lib/aws-c-common/cmake/aws-c-common-config-release.cmake
-- Installing: /builddir/build/BUILDROOT/aws-c-common-0.1-1.x86_64/usr/lib/cmake/AwsCFlags.cmake
-- Installing: /builddir/build/BUILDROOT/aws-c-common-0.1-1.x86_64/usr/lib/cmake/AwsTestHarness.cmake
-- Installing: /builddir/build/BUILDROOT/aws-c-common-0.1-1.x86_64/usr/lib/cmake/LibFuzzer.cmake
-- Installing: /builddir/build/BUILDROOT/aws-c-common-0.1-1.x86_64/usr/lib/cmake/Sanitizers.cmake
```

I'm really only familiar with packaging RPMs, but I could probably help set up CPack for RPMs if needed.

Thanks,
Phil

justinboswell:
Hi Phil,

1. I've got a fix for this, I believe. I will be testing it and it'll be in this PR: #214.
2. I'm experimenting with GNUInstallDirs right now
3. The test harness is a dependency of all of our C libraries which consume aws-c-common, so it should be installed
[...]

PR #214, opened 2018-12-17
CMake packaging cleanup/fixes

justinboswell:
*Issue #, if available: #213 


cmake/AwsCFlags.cmake
```diff
@@ -125,6 +125,8 @@ function(aws_set_common_properties target)
         if (NOT MSVC)
             list(APPEND AWS_C_FLAGS "-fvisibility=hidden")
         endif()
+
+        set_target_properties(${target} PROPERTIES VERSION 1.0.0 SOVERSION 0)
```
ColdenCullen: Why are we forcing everyone who uses our CFlags to be at version 1.0.0? Won't we want to release at different paces?

justinboswell: I was making sure there's at least something set here. If we don't do this, there will not be a so.version output unless the consuming project hand specifies VERSION on their target.

ColdenCullen: I think we punt this. I'd rather it didn't work and someone have to manually set their version than just go "ah but it works, I don't have to set this."

PR #216, opened 2018-12-17
Initial attempt at CPack for RPM generation

psalvaggio:
*Issue #, if available:*

*Description of changes:*
I added some CPack configuration to allow for packaging of RPMs (on RedHat-based OS's, I used Fedora 29). In order to make a package:
```
mkdir build && cd build
cmake -DBUILD_SHARED_LIBS=ON ..
make -j$(nproc) package
```


justinboswell:
Once you can merge CMakeLists.txt, I think we can review. There shouldn't be much different in there from master after my merge. The CPack stuff seems reasonable, so we can probably at least take that as a starting point.

CMakeLists.txt
```diff
@@ -169,32 +169,46 @@ foreach(HEADER_SRCPATH IN ITEMS ${AWS_COMMON_HEADERS} ${AWS_COMMON_OS_HEADERS} $
 		message(ERROR "Couldn't find source root for header ${HEADER_SRCPATH}")
 	endif()
 
-    install(FILES ${HEADER_SRCPATH} DESTINATION ${HEADER_DSTDIR})
+    install(FILES ${HEADER_SRCPATH}
+            DESTINATION ${HEADER_DSTDIR}
+            COMPONENT Development)
 endforeach()
 
-set(LIBRARY_DIRECTORY ${CMAKE_INSTALL_PREFIX}/lib)
```
psalvaggio: The `CMAKE_INSTALL_PREFIX` is implicit when specifying `DESTINATION` in `install()`, so it's normally better to leave it off for packaging, as specifying it will cause stuff to get installed to `/usr/local/` instead of `/usr/`.

CMakeLists.txt
```diff
@@ -169,32 +169,46 @@ foreach(HEADER_SRCPATH IN ITEMS ${AWS_COMMON_HEADERS} ${AWS_COMMON_OS_HEADERS} $
...
+set(LIBRARY_DIRECTORY lib)
 # Set the default lib installation path on GNU systems with GNUInstallDirs
 if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT AND UNIX AND NOT APPLE)
```
psalvaggio: Along the same lines, I think `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` would be unneeded, as you would still want the libraries to go to `lib64` (or equivalent) under a local install prefix.

psalvaggio:
Another issue that I'm coming across is that the imported targets seem to rely on a static library, even if the library is being built with BUILD_SHARED_LIBS=ON. I can look into a solution for this in this PR or separately.

CMakeLists.txt
justinboswell: Good point.

CMakeLists.txt
```diff
@@ -204,7 +221,9 @@ list(APPEND EXPORT_MODULES
     "cmake/AwsSIMD.cmake"
     )
 
-install(FILES ${EXPORT_MODULES} DESTINATION "lib/cmake/")
+install(FILES ${EXPORT_MODULES}
+        DESTINATION "${LIBRARY_DIRECTORY}/cmake/${PROJECT_NAME}"
```
justinboswell: This should be CMAKE_PROJECT_NAME, no?

psalvaggio: Had to look this up:

https://stackoverflow.com/questions/38938315/difference-between-cmake-project-name-and-project-name

In this situation, there's only one project command so the two would be equivalent. I don't really think there would be much reason to add another `project()` command, so I'm happy to use whichever one.

justinboswell: Aye, ok. I guess let's just error on the side of being consistent. We use CMAKE_PROJECT_NAME everywhere, let's keep doing it :)

justinboswell:
> Another issue that I'm coming across is that the imported targets seem to rely on a static library, even if the library is being built with BUILD_SHARED_LIBS=ON. I can look into a solution for this in this PR or separately.

I'm looking into this now as well.

psalvaggio:
So I tried a couple things out and it seems to work if I added
```
# Set the default build type to Release
if (NOT CMAKE_BUILD_TYPE)
  set(CMAKE_BUILD_TYPE "Release")
endif()
```
and changed the install of the targets to read like:
```
if (BUILD_SHARED_LIBS)
  install(TARGETS ${CMAKE_PROJECT_NAME}
      EXPORT ${CMAKE_PROJECT_NAME}-targets
      LIBRARY
        DESTINATION ${LIBRARY_DIRECTORY}
        NAMELINK_SKIP
        COMPONENT Runtime
      RUNTIME
        DESTINATION ${LIBRARY_DIRECTORY}
        COMPONENT Runtime)
  install(TARGETS ${CMAKE_PROJECT_NAME}
      EXPORT ${CMAKE_PROJECT_NAME}-targets
      LIBRARY
        DESTINATION ${LIBRARY_DIRECTORY}
        NAMELINK_ONLY
      COMPONENT Development)
else()
  install(TARGETS ${CMAKE_PROJECT_NAME}
      EXPORT ${CMAKE_PROJECT_NAME}-targets
      ARCHIVE DESTINATION ${LIBRARY_DIRECTORY}
      COMPONENT Development)
endif()
```
the release mode changed the filename from `noconfig` to `release` on the installed targets file (don't think that is related at all). I think getting the `ARCHIVE DESTINATION` out of the install when `BUILD_SHARED_LIBS` is on may have some effect. Does seem to work for me after those changes.

psalvaggio:
It looks like "Release" mode breaks a bit of the CI, I'll try reverting that part of the change, as I don't think that is related to the issue I was seeing.

justinboswell:
I'm repeating the test that failed in Travis thousands of times, no failure here. I'm going to re-run that build.

psalvaggio:
Cool, odd failure. I'm gonna make PRs for the two other repos that the C++ SDK depends on (checksums and c-event-stream) so they can get a similar treatment.

PR #220, opened 2018-12-24
Add extern C wrappers around aws_secure_zero

david-koenig:
Extern C wrappers were missing around aws_secure_zero, preventing proper
linkage from C++ code. This fixes the bug.


david-koenig:
@graebm Totally fair. I will switch it to have only one extern C clause, and after that passes CI, I will merge it.

PR #222, opened 2018-12-28
Versioning is in the way

JonathanHenson:

### DIFF
--- a/cmake/AwsSharedLibSetup.cmake
+++ b/cmake/AwsSharedLibSetup.cmake
@@ -11,6 +11,8 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
+option(VERSION_LIBS "Turns on versioning of shared libs, defaults to ON, you'll want to turn this off if you're building bindings and copying shared libs around." ON)
+
 set(LIBRARY_DIRECTORY lib)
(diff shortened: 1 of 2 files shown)


cmake/AwsSharedLibSetup.cmake
```diff
@@ -19,6 +21,12 @@ if (UNIX AND NOT APPLE)
 endif()
 
 function(aws_prepare_shared_lib_exports target)
+    if (VERSION_LIBS)
+        # Our ABI is not yet stable
+        set_target_properties(${target} PROPERTIES VERSION 1.0.0)
+        set_target_properties(${target} PROPERTIES SOVERSION 0unstable)
```
graebm: `option(VERSION_LIBS ...)` belongs in the `AwsSharedLibSetup.cmake`, but these lines should probably still be in aws-c-common's `CMakeLists.txt`. That way, each library can set its own version# (when VERSION_LIBS is ON)

JonathanHenson: Having to remember to put it in each lib defeats the point

ColdenCullen: Yeah but do we really want to force all of our libs to always use the same version number? I'm ok with putting SOVERSION here, but I do think the VERSION should be lib specific.

PR #223, opened 2019-01-03
Specialize hashlittle2, add aws_hash_iter_delete suppression

ColdenCullen:
* Only suppress errors uint overflow and global buffer overflow from `hashlittle2` (there's comments in the function about why it does these things)
* Suppress signed int overflow in `aws_hash_iter_delete`, there's also a comment about how it does this on purpose.

### DIFF
--- a/sanitizer-blacklist.txt
+++ b/sanitizer-blacklist.txt
@@ -3,5 +3,7 @@
 fun:hashlittle2
 
 [unsigned-integer-overflow]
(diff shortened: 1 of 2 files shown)


sanitizer-blacklist.txt
```diff
@@ -1,2 +1,5 @@
 # Disable hashlittle2, which purposefully reads past string ends
+[address|unsigned-integer-overflow]
 fun:hashlittle2
+[signed-integer-overflow]
```
graebm: put a comment on this one too

PR #224, opened 2019-01-03
Better type-safety and organization of callback functions for hash tables

david-koenig:
-make aws_string_destroy, aws_string_destroy_secure, and aws_string_eq type safe
-move aws_string_eq to string.h where it belongs
-change name of aws_equals_fn to aws_hash_element_eq_fn
-make aws_c_string_eq type safe
-remove aws_byte_cursor_ptr_eq
-use (aws_hash_element_eq_fn)aws_byte_cursor_eq instead of removed function
-make all calls to aws_hash_table_init, aws_hash_table_eq, and aws_lru_cache_init
 cast the pointers of type-safe functions to the prototypes taking void
-remove out of date casts from tests
-update comments in hash_table.h

### DIFF
--- a/source/hash_table.c
+++ b/source/hash_table.c
@@ -47,10 +47,10 @@ struct hash_table_entry {
 };
 
 struct hash_table_state {
-    aws_hash_fn hash_fn;
-    aws_hash_element_eq_fn equals_fn;
-    aws_hash_element_destroy_fn destroy_key_fn;
-    aws_hash_element_destroy_fn destroy_value_fn;
+    aws_hash_fn *hash_fn;
+    aws_hash_element_eq_fn *equals_fn;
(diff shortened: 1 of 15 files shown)


source/hash_table.c
```diff
@@ -833,8 +833,16 @@ uint64_t aws_hash_ptr(const void *item) {
     return ((uint64_t)b << 32) | c;
 }
 
-bool aws_c_string_eq(const char *a, const char *b) {
-    return !strcmp(a, b);
+bool aws_hash_callback_c_str_eq(const void *a, const void *b) {
```
JonathanHenson: You don’t need to cast to and from void*

david-koenig: Fixed

PR #226, opened 2019-01-05
Fix find_package on stupid systems that do the lib64 pathing thing.

JonathanHenson:
This is ssooooooo dumb, but dems tha rules.


singku:
Don't know why, but It turns out CPP SDK works correctly finding c-common in lib64 without this patch.

JonathanHenson:
It’s likely only a problem for C compilers, and it depends on cmake version. But event stream will need it regardless even if it’s fine for cpp sdk

PR #234, opened 2019-01-17
Fix Crossplatform Build for Android

singku:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/source/posix/time.c
+++ b/source/posix/time.c
@@ -54,17 +54,17 @@
  */
 #    include <time64.h>
 
-static const time_t s_time_max = ~(1L << (sizeof(time_t) * CHAR_BIT - 1));
-static const time_t s_time_min = (1L << (sizeof(time_t)) * CHAR_BIT - 1));
+static const time_t s_time_max = ~(1L << ((sizeof(time_t) * __CHAR_BIT__ - 1)));
+static const time_t s_time_min = (1L << ((sizeof(time_t)) * __CHAR_BIT__ - 1));
 
 /* 32-bit Android has only timegm64() and not timegm(). */
 time_t aws_timegm(struct tm *const t) {
 
     time64_t result = timegm64(t);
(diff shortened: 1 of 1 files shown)


source/posix/time.c
```diff
@@ -54,17 +54,17 @@
  */
 #    include <time64.h>
 
-static const time_t s_time_max = ~(1L << (sizeof(time_t) * CHAR_BIT - 1));
-static const time_t s_time_min = (1L << (sizeof(time_t)) * CHAR_BIT - 1));
+static const time_t s_time_max = ~(1L << ((sizeof(time_t) * __CHAR_BIT__ - 1)));
```
singku: Without proper header, `CHAR_BIT` is undefined. `__CHAR_BIT__` is defined.

singku: operator `<<` and `-` have different priority, compiler will give warning about `<<` has lower priority than `-`, even though this is exactly what we want.

source/posix/time.c
```diff
@@ -54,17 +54,17 @@
...
+static const time_t s_time_min = (1L << ((sizeof(time_t)) * __CHAR_BIT__ - 1));
 
 /* 32-bit Android has only timegm64() and not timegm(). */
 time_t aws_timegm(struct tm *const t) {
 
     time64_t result = timegm64(t);
-    if (result < time_min || result < time_max) {
+    if (result < s_time_min || result > s_time_max) {
```
singku: Typo here, should be `s_time_min` and `s_time_max` defined above.
Also should be `result > s_time_max`

source/posix/time.c
```diff
@@ -54,17 +54,17 @@
...
         return -1;
     }
-    return result;
+    return (time_t)result;
```
singku: Compiler warning of lost precision, better use explicit type conversion.

PR #240, opened 2019-01-25
Added CI job for x64 old linux.

JonathanHenson:

### DIFF
--- a/codebuild/ancient-linux-x64.yml
+++ b/codebuild/ancient-linux-x64.yml
@@ -1,0 +1,20 @@
+version: 0.2
+#this build spec assumes the ubuntu 14.04 trusty image
+phases:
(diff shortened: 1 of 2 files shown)


codebuild/ancient-linux-x64.yml
```diff
@@ -0,0 +1,24 @@
...
+
+  pre_build:
+    commands:
+      - export CC=gcc
+  build:
+    commands:
+      - echo Build started on `date`
+      - mkdir build
+      - cd build
+      - cmake ../
```
justinboswell: No generator? Does that work?

JonathanHenson: yeah, make is the default.

It works everywhere except for windows.

PR #243, opened 2019-02-01
First draft of logging system and format macros

bretambrose:
Will add comments/documentation once feedback settles and people are satisfied with the general framework/approach.

#### Properties
* Simple printf-style formatting
* Standard log levels; includes both dynamic filtering (via log level on the logger) and compile-time filtering (via AWS_STATIC_LOG_LEVEL)
* Current tests verify both static and dynamic filtering - for each level, verify that logging at levels that should pass do pass, and that logging at levels that should not pass, don't pass.
* (Log) Subject is a TBD placeholder for a second kind of filtering/partitioning concept.  When that is fleshed out it will have its own set of macros that in turn resolve down to LOGF_RAW.  Feedback/ideas/requirements around Subject are welcomed too.
* Next step (beyond feedback/documentation) would be implementations of basic loggers (stdout, synchronized file)


include/aws/common/logging.h
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/common.h>
+
+#define AWS_LOG_LEVEL_NONE 0
```
justinboswell: What's the reason for having these redundantly declared as macros and an enum? The enum could just have these values in it and cover everything?

tests/logging/logging_general_test.c
```diff
@@ -0,0 +1,47 @@
...
+static void s_log_all_levels_v2(enum aws_log_level level) {
+    LOGF_FATAL("%d", (int)AWS_LL_FATAL)
+    LOGF_ERROR("%d", (int)AWS_LL_ERROR)
+    LOGF_WARN("%d", (int)AWS_LL_WARN)
+    LOGF_INFO("%d", (int)AWS_LL_INFO)
+    LOGF_DEBUG("%d", (int)AWS_LL_DEBUG)
+    LOGF_TRACE("%d", (int)AWS_LL_TRACE)
+}
+
+TEST_LEVEL_FILTER(AWS_LL_TRACE, "123456", s_log_all_levels_v1)
```
justinboswell: This is so well abstracted that it is not clear at all what this is actually testing without a lot of digging. Could you add some comments to explain what is being tested here?

tests/logging/test_logger.c
```diff
@@ -0,0 +1,117 @@
...
+#define TEST_LOGGER_MAX_LOG_LINE_SIZE 256
+
+int s_test_logger_log_fn(struct aws_logger *logger, enum aws_log_level log_level, aws_log_subject_t subject, const char *format, ...) {
+    va_list format_args;
+    va_start(format_args, format);
+
+    static char buffer[TEST_LOGGER_MAX_LOG_LINE_SIZE];
+
+#ifdef WIN32
+    int written = vsnprintf_s(buffer, MAX_LOG_LINE_SIZE, _TRUNCATE, format, format_args);
```
justinboswell: Shouldn't this be TEST_LOGGER_MAX_LOG_LINE_SIZE?

tests/logging/logging_filter_debug_static_test.c
```diff
@@ -0,0 +1,22 @@
...
+ * permissions and limitations under the License.
+ */
+
+#define AWS_STATIC_LOG_LEVEL 5
+
+#include "logging_test_utilities.h"
+
+DECLARE_LOG_ALL_LEVELS_FUNCTION(s_log_all_levels_debug_cutoff)
+
+TEST_LEVEL_FILTER(AWS_LL_TRACE, "12345", s_log_all_levels_debug_cutoff)
```
justinboswell: It would be good for each of these files to comment on what is being tested

include/aws/common/logging.h
bretambrose: In order to do the preprocessor-level filter comparison we must have the value #defined.  The enum isn't strictly necessary, but I kind of prefer it as the user-facing type.  I don't think it adds any type safety though in C?

tests/logging/logging_general_test.c
bretambrose: Definitely

tests/logging/test_logger.c
bretambrose: Yes it should.

tests/logging/logging_filter_debug_static_test.c
bretambrose: I agree.  Waiting a bit before documentation in case there are large-scale changes that invalidate comments.

PR #248, opened 2019-02-15
Simulated codebuild environment for travis-ci

justinboswell:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,5 @@ install:
   - export PATH=$(brew --prefix llvm)/bin:$PATH
 
 script:
+  - export CODEBUILD_SRC_DIR=$(pwd)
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,4 +12,5 @@ install:
   - export PATH=$(brew --prefix llvm)/bin:$PATH
 
 script:
+  - export CODEBUILD_SRC_DIR=$(pwd)
```
graebm: alternatively, common-posix.sh could do this instead of using CODEBUILD_SRC_DIR:
```
SRC_DIR=`pwd`
```

justinboswell: No it can't, because everything has to be based on the codebuild env vars, or it's not guaranteed to work. The working directory has shifted on us.

graebm: 🥇

PR #249, opened 2019-02-19
Prevent integer overflow in array_list_init functions

danielsn:
*Description of changes:*
Add error checking to the ```array_list_init*``` functions to avoid potential integer overflow.

### DIFF
--- a/include/aws/common/array_list.h
+++ b/include/aws/common/array_list.h
@@ -16,6 +16,7 @@
  * permissions and limitations under the License.
  */
 #include <aws/common/common.h>
+#include <aws/common/math.h>
 
 #include <assert.h>
 #include <stdlib.h>
@@ -61,7 +62,7 @@ int aws_array_list_init_dynamic(
  * by this API. Once this list is full, new items will be rejected.
  */
 AWS_STATIC_IMPL
-void aws_array_list_init_static(
+int aws_array_list_init_static(
     struct aws_array_list *AWS_RESTRICT list,
     void *raw_array,
     size_t item_count,
--- a/include/aws/common/array_list.h
+++ b/include/aws/common/array_list.h
@@ -62,7 +62,7 @@ int aws_array_list_init_dynamic(
  * by this API. Once this list is full, new items will be rejected.
  */
 AWS_STATIC_IMPL
-int aws_array_list_init_static(
+void aws_array_list_init_static(
     struct aws_array_list *AWS_RESTRICT list,
     void *raw_array,
     size_t item_count,
--- a/include/aws/common/array_list.inl
+++ b/include/aws/common/array_list.inl
@@ -51,7 +51,7 @@ int aws_array_list_init_dynamic(
 }
 
 AWS_STATIC_IMPL
-int aws_array_list_init_static(
+void aws_array_list_init_static(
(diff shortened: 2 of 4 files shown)


include/aws/common/array_list.inl
```diff
@@ -58,8 +61,9 @@ void aws_array_list_init_static(
     assert(item_size);
 
     list->alloc = NULL;
-
-    list->current_size = item_count * item_size;
+    if(!aws_mul_u64_checked(item_count, item_size, &list->current_size)) {
+        return AWS_OP_ERR;
```
ColdenCullen: This function has a return type of void. Did you compile this code locally?

danielsn: Sorry - I forgot to mark this "review only"

danielsn: Fixed.

include/aws/common/array_list.h
```diff
@@ -61,7 +62,7 @@ int aws_array_list_init_dynamic(
  * by this API. Once this list is full, new items will be rejected.
  */
 AWS_STATIC_IMPL
-void aws_array_list_init_static(
+int aws_array_list_init_static(
```
ColdenCullen: This is a big change, static inits are never expected to fail. If this is what we decide to do, we should go through every call to this function and add the error check.

bdonlan: For the static init it's probably reasonable to skip the overflow check. After all, the caller is promising to already have this memory allocated; the overflow check should have happened at that point (and usually static is used for arrays of statically defined size).

ColdenCullen: In that case should we remove the error return and instead assert that the size is reasonable? If the size is incorrect, this would imply programmer error, making an assert the correct way to raise the issue.

danielsn: there is an assert in https://github.com/awslabs/aws-c-common/pull/249/files#diff-a1ef357fc1b5c512b3842ba2d64ffac7R61 
Would it make sense to make this an assert as well, and return to the void function contract?

ColdenCullen: I meant the total size, the one we're checking for overflow. I would rather see that become an assert and the function return void, especially if the caller isn't expected to handle the error.

include/aws/common/array_list.inl
```diff
@@ -61,13 +61,11 @@ int aws_array_list_init_static(
     assert(item_size);
 
     list->alloc = NULL;
-    if(!aws_mul_size_checked(item_count, item_size, &list->current_size)) {
-        return AWS_OP_ERR;
-    }
+    int no_overflow = aws_mul_size_checked(item_count, item_size, &list->current_size);
+    assert(no_overflow);
```
bretambrose: Address windows compilation issue by switching to AWS_FATAL_ASSERT maybe.

danielsn:
Switched to the ```aws_fatal_assert()``` macro.  It was actually missing the ```stdio.h``` required for ```fprintf```, so I had to add that too.

bretambrose:
Didn't see Colden's comments and merged prematurely.  Can you create a new PR with the final fix?  Update seems small enough to warrant avoiding reverting and reapplying.

danielsn:
https://github.com/awslabs/aws-c-common/pull/251

PR #254, opened 2019-02-22
create a new macro AWS_RETURN_ERR_IF() and use it when possible

danielsn:
*Description of changes:*

1. Following s2n, replace the common pattern
``` 
if(cond) {
    return AWS_OP_ERR;
}
``` 
with a new macro
``` AWS_RETURN_ERR_IF(cond);```

2. Use the new macro whenever possible

### CODE REVIEW

File: source/priority_queue.c [hunk 1]
```diff
@@ -218,10 +218,9 @@ int aws_priority_queue_push_ref(
 }
 
 static int s_remove_node(struct aws_priority_queue *queue, void *item, size_t item_index) {
-    if (aws_array_list_get_at(&queue->container, item, item_index)) {
-        /* shouldn't happen, but if it does we've already raised an error... */
-        return AWS_OP_ERR;
-    }
+
+    /* shouldn't happen, but if it does we've already raised an error... */
```
tautschnig: That comment may now need rewording.

source/byte_buf.c
```diff
@@ -89,9 +87,7 @@ int aws_byte_buf_init_copy_from_cursor(
     }
 
     dest->buffer = (uint8_t *)aws_mem_acquire(allocator, sizeof(uint8_t) * src.len);
-    if (dest->buffer == NULL) {
-        return AWS_OP_ERR;
-    }
+    AWS_RETURN_ERR_IF(dest->buffer == NULL);
```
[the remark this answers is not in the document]
danielsn: I'm OK either way.  My thought was to do the a direct translation that is obviously correct, which can then be improved in a second PR.  But I'm happy to change if it that would make better code.

graebm:
I feel like:
1/3 of the time I'm doing `if (err) return AWS_OP_ERR;`
1/3 of the time I'm doing `if (err) return aws_raise_error(AWS_ERROR_X);`
1/3 of the time I'm doing `if (err) goto cleanup;`

Error handling is finicky and and I'm not sure if having a macro to wrap just 1 of the common cases makes coding mistakes more or less likely. Like, will people lean towards using this macro and forgot to raise new errors where necessary?

I don't feel super strongly, just playing devil's advocate.

JonathanHenson:
I’m on vacation and have only followed this briefly but I had the same initial reaction as Mike. We at least need two macros, one that raises and error, and one that assumes one has already been raised.

JonathanHenson:
Err three. And additional one that goes to a label, but I fear that one would be more cognitive burden on future developers

danielsn:
I added a macro for the RAISE_ERR case, and used it when appropriate.

A couple questions:
1. There are several files which use the pattern 
```
if (something_went_right) return AWS_OP_SUCCESS;
return aws_raise_error(ERRCODE);
```
the s2n way, which I like, is to detect and raise errors early, then the straightline path leads to success.  Should I rewrite those cases to follow this pattern.
2. There are fairly few cases of ```goto```.  Would it make sense to have a macro for them?
```
./include/aws/common/private/lookup3.c:	  if (z==MAXPAIR) goto done;
./source/common.c:            goto cleanup;
./source/posix/thread.c:            goto cleanup;
./source/posix/thread.c:                goto cleanup;
./source/posix/thread.c:        goto cleanup;
./source/posix/thread.c:        goto cleanup;
./source/priority_queue.c:            goto backpointer_update_failed;
./source/priority_queue.c:            goto backpointer_update_failed;
./source/priority_queue.c:            goto backpointer_update_failed;
```

include/aws/common/error.h
```diff
@@ -126,6 +126,20 @@ aws_error_handler_fn *aws_set_thread_local_error_handler_fn(aws_error_handler_fn
...
+#define AWS_RETURN_ERR_IF(cond)                                                                                        \
+    do {                                                                                                               \
+        if (cond) {                                                                                                    \
+            return AWS_OP_ERR;                                                                                         \
+        }                                                                                                              \
+    } while (0)
+
+#define AWS_RAISE_ERR_IF(cond, err)                                                                                    \
+    do {                                                                                                               \
+        if (cond) {                                                                                                    \
```
danielsn: Since we expect errors to be unlikely, would it make sense to wrap this condition in ```AWS_UNLIKELY```?

PR #255, opened 2019-02-22
Update Makefiles for CBMC proofs

feliperodri:
*Description of changes:*

Update all Makefiles to run each harness locally. Thus, we move all proofs to their correspondent harness directories and update all makefiles and yaml files.


.cbmc-batch/jobs/Makefile.common
```diff
@@ -0,0 +1,211 @@
...
+################################################################
+# Define locations and binaries
+ABSTRACTIONS ?= ""
+BASEDIR ?= $(abspath ../../../..)
+BATCH ?= cbmc-batch
+
+CBMCFLAGS += \
+	--bounds-check \
+	--pointer-check \
+	--function $(ENTRY) \
```
danielsn: You should really also set --unwinding-assertions \

feliperodri: I've added `--unwinding-assertions` since it doesn't break any proofs, but we should add the rest of CBMC flags on a separate PR.

PR #256, opened 2019-02-26
fix incorrect cast in aws_mul_size_checked()

danielsn:
*Description of changes:*
```aws_mul_size_checked()``` incorrectly downcasts in the case of 64 bit numbers.  This PR fixes it.

### DIFF
--- a/include/aws/common/math.h
+++ b/include/aws/common/math.h
@@ -91,10 +91,10 @@ AWS_STATIC_IMPL size_t aws_mul_size_saturating(size_t a, size_t b) {
 }
 
 AWS_STATIC_IMPL int aws_mul_size_checked(size_t a, size_t b, size_t *r) {
-    if ((uint64_t)SIZE_MAX == ~(uint32_t)0) {
-        return (int)aws_mul_u32_checked((uint32_t)a, (uint32_t)b, (uint32_t *)r);
+    if (sizeof(size_t) == sizeof(uint32_t)) {
+        return aws_mul_u32_checked(a, b, (uint32_t *)r);
     }
-    return (int)aws_mul_u64_checked((uint32_t)a, (uint32_t)b, (uint64_t *)r);
+    return aws_mul_u64_checked(a, b, (uint64_t *)r);
 }
 
 #if _MSC_VER
--- a/include/aws/common/math.h
+++ b/include/aws/common/math.h
@@ -78,16 +78,13 @@ AWS_COMMON_API int aws_mul_u32_checked(uint32_t a, uint32_t b, uint32_t *r);
 #endif                              /* _MSC_VER */
 
 AWS_STATIC_IMPL size_t aws_mul_size_saturating(size_t a, size_t b) {
-    /* static assert: SIZE_MAX == (~(uint32_t)0) || (~(uint64_t)0)*/
-    char assert_sizet_is_32_or_64_bit
-        [(((uint64_t)SIZE_MAX == ~(uint32_t)0) || ((uint64_t)SIZE_MAX == ~(uint64_t)0)) ? 1 : -1] = {0};
-    /* suppress unused variable warning */
-    (void)assert_sizet_is_32_or_64_bit;
-
-    if ((uint64_t)SIZE_MAX == ~(uint32_t)0) {
-        return (size_t)aws_mul_u32_saturating((uint32_t)a, (uint32_t)b);
-    }
+#if SIZE_MAX == UINT32_MAX
+    return (size_t)aws_mul_u32_saturating(a, b);
+#elif SIZE_MAX == UINT64_MAX
     return (size_t)aws_mul_u64_saturating(a, b);
+#else
+#    error "Target not supported"
+#endif
 }
 
 AWS_STATIC_IMPL int aws_mul_size_checked(size_t a, size_t b, size_t *r) {
--- a/include/aws/common/math.h
+++ b/include/aws/common/math.h
@@ -91,10 +91,13 @@ AWS_STATIC_IMPL size_t aws_mul_size_saturating(size_t a, size_t b) {
 }
 
 AWS_STATIC_IMPL int aws_mul_size_checked(size_t a, size_t b, size_t *r) {
-    if (sizeof(size_t) == sizeof(uint32_t)) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by bdonlan** [comment]
(of course, the test will need to no-op out on 32-bit systems)

**Comment by danielsn** [comment]
Willdo

File: include/aws/common/math.h [hunk 1]
```diff
@@ -91,10 +91,10 @@ AWS_STATIC_IMPL size_t aws_mul_size_saturating(size_t a, size_t b) {
 }
 
 AWS_STATIC_IMPL int aws_mul_size_checked(size_t a, size_t b, size_t *r) {
-    if ((uint64_t)SIZE_MAX == ~(uint32_t)0) {
-        return (int)aws_mul_u32_checked((uint32_t)a, (uint32_t)b, (uint32_t *)r);
+    if (sizeof(size_t) == sizeof(uint32_t)) {
```
ColdenCullen: There should be a way to do this at compile time, right? Can you do:
```c
#if SIZE_MAX == UINT32_MAX
    return aws_mul_u32_checked(a, b, r);
#elif SIZE_MAX == UINT64_MAX
    return aws_mul_u64_checked(a, b, r);
#else
#   error "Target not supported"
#endif
```

danielsn: Yes.  That looks good.

danielsn: > (of course, the test will need to no-op out on 32-bit systems)

What's the best way to make a test that only runs on some systems? Is there a flag in the testing framework?

danielsn: Also, should I apply that format to the ```aws_mul_size_saturating()``` function, while I'm here? Or save that for a separate PR?

ColdenCullen: 1. Just `#ifdef` the body of the test, so on 32 bits the only code that is compiled is `return AWS_OP_SUCCESS`
2. Yeah I'd say go for it, if we say "let's do it later" it'll never happen 😛

include/aws/common/math.h
```diff
@@ -91,10 +91,13 @@ AWS_STATIC_IMPL size_t aws_mul_size_saturating(size_t a, size_t b) {
 }
 
 AWS_STATIC_IMPL int aws_mul_size_checked(size_t a, size_t b, size_t *r) {
-    if ((uint64_t)SIZE_MAX == ~(uint32_t)0) {
-        return (int)aws_mul_u32_checked((uint32_t)a, (uint32_t)b, (uint32_t *)r);
-    }
-    return (int)aws_mul_u64_checked((uint32_t)a, (uint32_t)b, (uint64_t *)r);
+#if SIZE_MAX == UINT32_MAX
+    return aws_mul_u32_checked(a, b, (uint32_t *)r);
```
ColdenCullen: These casts should no longer be required? If we can safely remove them, having 1 fewer cast could help the compiler find issues like this in the future.

include/aws/common/math.h
danielsn: Weirdly, it seems to require the cast: if I remove it, I get
```
make[2]: *** [CMakeFiles/aws-c-common.dir/source/byte_buf.c.o] Error 1
In file included from /Users/dsn/ws/CCommon-CBMC/aws-c-common/source/date_time.c:17:
In file included from /Users/dsn/ws/CCommon-CBMC/aws-c-common/include/aws/common/array_list.h:19:
/Users/dsn/ws/CCommon-CBMC/aws-c-common/include/aws/common/math.h:97:38: error: incompatible pointer types passing 'size_t *' (aka 'unsigned long *') to parameter of type 'uint64_t *' (aka 'unsigned long long *') [-Werror,-Wincompatible-pointer-types]
    return aws_mul_u64_checked(a, b, r);
                                     ^
/Users/dsn/ws/CCommon-CBMC/aws-c-common/include/aws/common/math.gcc_overflow.inl:39:75: note: passing argument to parameter 'r' here
AWS_STATIC_IMPL int aws_mul_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
                                                                          ^
1 error generated.
make[2]: *** [CMakeFiles/aws-c-common.dir/source/date_time.c.o] Error 1
```

include/aws/common/math.h
danielsn: Weirdly, it seems to require the cast: if I remove it, I get
```
make[2]: *** [CMakeFiles/aws-c-common.dir/source/byte_buf.c.o] Error 1
In file included from /Users/dsn/ws/CCommon-CBMC/aws-c-common/source/date_time.c:17:
In file included from /Users/dsn/ws/CCommon-CBMC/aws-c-common/include/aws/common/array_list.h:19:
/Users/dsn/ws/CCommon-CBMC/aws-c-common/include/aws/common/math.h:97:38: error: incompatible pointer types passing 'size_t *' (aka 'unsigned long *') to parameter of type 'uint64_t *' (aka 'unsigned long long *') [-Werror,-Wincompatible-pointer-types]
    return aws_mul_u64_checked(a, b, r);
                                     ^
/Users/dsn/ws/CCommon-CBMC/aws-c-common/include/aws/common/math.gcc_overflow.inl:39:75: note: passing argument to parameter 'r' here
AWS_STATIC_IMPL int aws_mul_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
                                                                          ^
1 error generated.
make[2]: *** [CMakeFiles/aws-c-common.dir/source/date_time.c.o] Error 1
```

include/aws/common/math.h
ColdenCullen: If it's trying to tell you that a `size_t` is 32 bits, then something's wrong with the branching.

danielsn: I did some digging.  It seems that on MacOS, ```size_t``` is a 64 bit ```long unsigned int``` while ```uint64_t``` is a 64bit ```unsigned long long```.  The compiler complains when you try to turn a pointer to one into a pointer to the other without a cast, even tho they are the same size.

tautschnig:
The verification failure of aws_array_list_init_static is spurious - it requires a fix in CBMC (https://github.com/diffblue/cbmc/pull/4283) to be merged. This is showing up only now as we didn't previously trigger any use of fprintf/vfprintf on this code.

tautschnig:
The CBMC fix has been merged, all CI jobs succeed.

PR #257, opened 2019-02-27
Add support for checked addition in math.h

danielsn:
*Description of changes:*
1. Add support for checked addition in math.h 
1. Add appropriate tests.


include/aws/common/math.fallback.inl
```diff
@@ -66,3 +66,51 @@ AWS_STATIC_IMPL int aws_mul_u32_checked(uint32_t a, uint32_t b, uint32_t *r) {
     }
     return 1;
 }
+
+/**
+ * Adds a + b.  If the result overflows returns 2^64 - 1.
+ */
+AWS_STATIC_IMPL uint64_t aws_add_u64_saturating(uint64_t a, uint64_t b) {
+  uint64_t x = a + b;
```
danielsn: Another way to do this is to pre-check -> this requires one extra arithmetic operation but avoids ever having overflow behaviour
```
 if(UINT64_MAX - b < a) {
    return UINT64_MAX;
  }
  return a + b;
```

include/aws/common/math.fallback.inl
```diff
@@ -66,3 +66,51 @@ AWS_STATIC_IMPL int aws_mul_u32_checked(uint32_t a, uint32_t b, uint32_t *r) {
...
+    return UINT64_MAX;
+  }
+  return x;
+}
+
+/**
+ * Adds a + b and returns the result in *r. If the result overflows,
+ * returns 0, else returns 1.
+ */
+AWS_STATIC_IMPL int aws_add_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
```
graebm: we usually return AWS_OP_SUCCESS(0) for success and AWS_OP_ERR(-1) for failure

this is returning 1 for success and 0 for failure

danielsn: This is the same behaviour as ```aws_mul_u64_checked()``` https://github.com/awslabs/aws-c-common/blob/master/include/aws/common/math.fallback.inl#L33

I agree that its odd.  Would be make more sense to
1. fix the ```mul``` functions to use the standard convention
1. follow the same (nonstandard) pattern as ```mul```
1. have two different patterns?

danielsn: In an ideal world, the best thing would be option 1.  But do we know if we have consumers who would be broken by that?

graebm: TLDR; I'll ask other teams that I know are using this library. It's not GA yet, so in theory we could change this.

I have all the code belonging to the `awslabs/aws-sdk-common-runtime-team` on my drive and searched it for uses of the `aws_mul_X_checked()` functions. Only `aws-c-common` uses this function, once in `hash_table.c` and once in `array_list.inl`.

The [GCC builtin](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html) is `bool __builtin_add_overflow(A,B,C)`. It returns true(1) if it overflows, but the function is named "overflow" so that makes sense. Ours is named "checked" (feels like the opposite of "overflow") and returns an int instead of a bool, so I would expect AWS_OP_ERR if it overflowed.

graebm: Per the discussion in the chatroom, it's safe to change the signature of these functions. The only uses are here in aws-c-common.

danielsn: Change implemented as requested.

PR #266, opened 2019-03-11
Simple predicate-based trimming for byte cursors

bretambrose:
Adds left and right trim functionality to byte cursor API


tests/cursor_test.c
```diff
@@ -237,3 +237,129 @@ static int s_byte_cursor_limit_tests_fn(struct aws_allocator *allocator, void *c
...
+static const char *s_all_whitespace = " \t\r\n ";
+static const char *s_left_whitespace = "\t \r" TEST_STRING;
+static const char *s_right_whitespace = TEST_STRING " \r \t \n";
+static const char *s_both_whitespace = "  \t \r\n " TEST_STRING " \r \t \n";
+static const char *expected_non_empty_result = TEST_STRING;
+
+static bool s_is_whitespace(uint8_t value) {
+    char value_as_char = (char)value;
+
+    switch (value_as_char) {
```
JonathanHenson: isspace() ?

bretambrose: The signatures are incompatible and I feel like the trim function predicates should match the type used internally by byte_cursor/string etc... (even though the cast from uint8_t -> int is safe as far as I can tell).  I can see the usefulness of having the bytecursor predicates compatible with clib character predicates though.  I'm conflicted.

PR #267, opened 2019-03-11
Make a private header file for the hash_table implementation

danielsn:
*Description of changes:*
Declare the hash_table implementation functions in a private header to make them easier to verify/test.

### DIFF
--- a/include/aws/common/common.h
+++ b/include/aws/common/common.h
@@ -22,6 +22,10 @@
 #include <stdio.h>
 #include <string.h>
 
+#ifndef AWS_STATIC
+#    define AWS_STATIC static
+#endif
+
 #ifndef AWS_STATIC_IMPL
 /*
  * In order to allow us to export our inlinable methods in a DLL/.so, we have a designated .c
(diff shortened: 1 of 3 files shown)


include/aws/common/private/hash_table.h
```diff
@@ -0,0 +1,103 @@
...
+
+AWS_STATIC int s_expand_table(struct aws_hash_table *map);
+
+/* Clears an entry. Does _not_ invoke destructor callbacks.
+ * Returns the last slot touched (note that if we wrap, we'll report an index
+ * lower than the original entry's index)
+ */
+AWS_STATIC size_t s_remove_entry(struct hash_table_state *state, struct hash_table_entry *entry);
+
+AWS_STATIC_IMPL void s_get_next_element(struct aws_hash_iter *iter, size_t start_slot);
```
ColdenCullen: `AWS_STATIC_IMPL` isn't doing what you think it's doing, it exists so that codegen.c can export functions that can also be used statically. This function should not be exported, so it should not be `AWS_STATIC_IMPL`, just `static inline`.

include/aws/common/common.h
```diff
@@ -22,6 +22,10 @@
 #include <stdio.h>
 #include <string.h>
 
+#ifndef AWS_STATIC
+#    define AWS_STATIC static
+#endif
```
ColdenCullen: Why do we need a macro for this?

PR #274, opened 2019-03-13
Added AL2012 CI build

justinboswell:


codebuild/linux-al2012-gcc44-x64.yml
```diff
@@ -0,0 +1,6 @@
+version: 0.2
+#this buildspec assumes the aws-sdk-common-runtime al2012:latest image
+phases:
+  build:
+    commands:
+      - ./codebuild/common-posix.sh -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DENABLE_SANITIZERS=OFF -DPERFORM_HEADER_CHECK=OFF
```
graebm: Do I even **want** to know why sanitizers and header-check are turned off?

justinboswell: Yes, you do. It's because gcc-4.4.6 has no idea what sanitizers are, and because header check requires a C++ compiler, and g++ is not installed.

PR #279, opened 2019-03-15
Implemented most of posix getlong_opt()

JonathanHenson:
Some of the naming is inconsistent with our coding standards, but they're that way to match the api programmers are used to with getopt_long()


source/command_line_parser.c
```diff
@@ -0,0 +1,121 @@
...
+
+int aws_cli_getopt_long(
+    int argc,
+    char *const argv[],
+    const char *optstring,
+    const struct aws_cli_option *longopts,
+    int *longindex) {
+    aws_cli_optarg = NULL;
+
+    if (aws_cli_optind == argc) {
```
justinboswell: Any reason not to have this check >= argc? Should render it impossible to overrun argv provided they match

JonathanHenson: Sure

Issue #282, opened 2019-03-20
Integrate remaining aws_string proofs once necessary CBMC enhancements done


danielsn:
- [ ] aws_string_new_from_c_str
- [ ] aws_string_new_from_array
- [ ] aws_string_new_from_string
- [ ] aws_array_list_comparator_string

ColdenCullen:
What's the status of this? Can it be closed?

tautschnig:
@danielsn ?

PR #289, opened 2019-03-26
Remaining string proofs

danielsn:
*Issue #, if available:*
https://github.com/awslabs/aws-c-common/issues/282

*Description of changes:*
The required CBMC changes were made, merge in the remaining proofs.

### DIFF
--- a/.cbmc-batch/jobs/aws_array_list_comparator_string/aws_array_list_comparator_string_harness.c
+++ b/.cbmc-batch/jobs/aws_array_list_comparator_string/aws_array_list_comparator_string_harness.c
@@ -19,7 +19,7 @@
 #include <proof_helpers/utils.h>
 #include <stddef.h>
 
-const size_t MAX_STRING_LEN = 16;
+//MAX_STRING_LEN is defined in the makefile
 
 /**
  * Coverage: 1.00 (56 lines out of 56 statically-reachable lines in 10 functions reached)
--- a/.cbmc-batch/jobs/aws_array_list_comparator_string/aws_array_list_comparator_string_harness.c
+++ b/.cbmc-batch/jobs/aws_array_list_comparator_string/aws_array_list_comparator_string_harness.c
@@ -19,7 +19,7 @@
 #include <proof_helpers/utils.h>
 #include <stddef.h>
 
-//MAX_STRING_LEN is defined in the makefile
+// MAX_STRING_LEN is defined in the makefile
 
 /**
  * Coverage: 1.00 (56 lines out of 56 statically-reachable lines in 10 functions reached)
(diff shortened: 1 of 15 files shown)


.cbmc-batch/jobs/aws_array_list_comparator_string/aws_array_list_comparator_string_harness.c
```diff
@@ -0,0 +1,36 @@
...
+ * Runtime: real	0m4.682s
+ */
+void aws_array_list_comparator_string_harness() {
+    struct aws_string *str_a = make_arbitrary_aws_string_nondet_len_with_max(can_fail_allocator(), MAX_STRING_LEN);
+    struct aws_string *str_b =
+        nondet_bool() ? str_a : make_arbitrary_aws_string_nondet_len_with_max(can_fail_allocator(), MAX_STRING_LEN);
+    int rval = aws_array_list_comparator_string(&str_a, &str_b);
+    if (!rval) {
+        assert_bytes_match(str_a->bytes, str_b->bytes, str_a->len);
+    }
```
feliperodri: Can you make sure that `str_a->bytes` does not match `str_b->bytes` if `!rval == false`?

.cbmc-batch/jobs/aws_array_list_comparator_string/aws_array_list_comparator_string_harness.c
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+#include <aws/common/string.h>
+#include <proof_helpers/make_common_data_structures.h>
+#include <proof_helpers/proof_allocators.h>
+#include <proof_helpers/utils.h>
+#include <stddef.h>
+
+const size_t MAX_STRING_LEN = 16;
```
karkhaz: You're defining identical values in each of the harnesses.

I'm not going to block this PR, but please consider defining this once in a tiny makefile, and including that makefile from each of the harness-specific makefiles. Then write a comment above the define explaining why you chose the value 16. That way we have a way to change the value from one place if necessary, while retaining the ability to override it on a per-harness basis.

.cbmc-batch/jobs/aws_array_list_comparator_string/aws_array_list_comparator_string_harness.c
danielsn: That would require a universal quantifier, rather than an existential qualifier, and would be inefficient.  If we can find a nice way to do it, we should add that in a future PR.

.cbmc-batch/jobs/aws_array_list_comparator_string/aws_array_list_comparator_string_harness.c
danielsn: Done

PR #290, opened 2019-03-26
case-insensitive string functions

graebm:
Needed by aws-c-http.

Fun Fact: Did some local benchmarking. Looking up values in a static array is a 1.4X speedup vs checking if a letter is between 'A':'Z' and lowercasing it.

### DIFF
--- a/source/byte_buf.c
+++ b/source/byte_buf.c
@@ -260,6 +260,24 @@ bool aws_byte_cursor_eq_case_insensitive(const struct aws_byte_cursor *a, const 
     return true;
 }
 
+uint64_t aws_hash_byte_cursor_ptr_case_insensitive(const void *item) {
+    /* FNV-1a: https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function */
+    const uint64_t fnv_offset_basis = 0xcbf29ce484222325ULL;
+    const uint64_t fnv_prime = 0x100000001b3ULL;
+
+    const struct aws_byte_cursor *cursor = item;
+    const uint8_t *i = cursor->ptr;
+    const uint8_t *end = cursor->ptr + cursor->len;
+
+    uint64_t hash = fnv_offset_basis;
+    while (i != end) {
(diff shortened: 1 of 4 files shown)


source/byte_buf.c
```diff
@@ -227,6 +227,57 @@ bool aws_byte_cursor_eq(const struct aws_byte_cursor *a, const struct aws_byte_c
...
+    'x', 'y', 'z', 91,  92,  93,  94,  95,  96,  'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm',
+    'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', 123, 124, 125, 126, 127, 128, 129, 130, 131,
+    132, 133, 134, 135, 136, 137, 138, 139, 140, 141, 142, 143, 144, 145, 146, 147, 148, 149, 150, 151, 152, 153,
+    154, 155, 156, 157, 158, 159, 160, 161, 162, 163, 164, 165, 166, 167, 168, 169, 170, 171, 172, 173, 174, 175,
+    176, 177, 178, 179, 180, 181, 182, 183, 184, 185, 186, 187, 188, 189, 190, 191, 192, 193, 194, 195, 196, 197,
+    198, 199, 200, 201, 202, 203, 204, 205, 206, 207, 208, 209, 210, 211, 212, 213, 214, 215, 216, 217, 218, 219,
+    220, 221, 222, 223, 224, 225, 226, 227, 228, 229, 230, 231, 232, 233, 234, 235, 236, 237, 238, 239, 240, 241,
+    242, 243, 244, 245, 246, 247, 248, 249, 250, 251, 252, 253, 254, 255};
+
+bool aws_byte_cursor_eq_case_insensitive(const struct aws_byte_cursor *a, const struct aws_byte_cursor *b) {
```
justinboswell: Could you not just make this into a function that compares 2 ranges (a, a_len, b, b_len) so we could use the same function for  byte_buf, byte_cursor, and string?

PR #292, opened 2019-03-27
Compare ALL the buffer types ALL the ways!

graebm:
We have 4 different widely used buffer types: `aws_byte_cursor`, `aws_byte_buf`, `const char *`, & `aws_string`

- Add **new** functions for comparing against `const char *` in a single pass (no need for `strlen()`)
- Add case-insensitive comparison functions for **every** buffer type against **every other** buffer type.
- Added base implementations (ex: `aws_array_eq()`) used by each type-specific permutation.
- renamed from `_case_insensitive` to `_ignore_case` because it's less letters.

This change is AWESOME. I'm using nearly all of these permutations now in aws-c-http.

NITPICKS:
These algorithms bend over backwards to accept NULL. I wouldn't have personally chosen that path if the original algorithms didn't already handle it.


include/aws/common/byte_buf.h
```diff
@@ -233,36 +301,73 @@ AWS_COMMON_API
 int aws_byte_buf_cat(struct aws_byte_buf *dest, size_t number_of_args, ...);
 
 /**
- * Compares two aws_byte_cursor structures
- * Returns true if a has the same length as b and their buffers have the same bytes
- * (or both buffers are null). When both a and b are null the function returns true
+ * Compare two aws_byte_cursor structures.
+ * Return whether their contents are equivalent.
+ * NULL inputs are considered equal if both are NULL, or both lengths are 0.
```
justinboswell: "NULL or empty (length = 0) inputs are considered equal if both are null or empty" perhaps? As it reads right now, it implies that null inputs have lengths.

include/aws/common/string.h
```diff
@@ -68,27 +68,91 @@ AWS_STATIC_IMPL const uint8_t *aws_string_bytes(const struct aws_string *str) {
...
+
+    return aws_array_eq(a_ptr, a_len, b_ptr, b_len);
+}
+
+/**
+ * Returns true if bytes of string and buffer are equivalent, using a case-insensitive comparison.
+ */
+AWS_STATIC_IMPL
+bool aws_string_eq_byte_buf_ignore_case(const struct aws_string *str, const struct aws_byte_buf *buf) {
+    const uint8_t *a_ptr = str ? str->bytes : NULL;
```
justinboswell: aws_string_bytes()?

source/byte_buf.c
```diff
@@ -242,32 +259,102 @@ static const uint8_t s_tolower_table[256] = {
...
 }
 
-uint64_t aws_hash_byte_cursor_ptr_case_insensitive(const void *item) {
+bool aws_array_eq(const void *array_a, size_t len_a, const void *array_b, size_t len_b) {
+    if (len_a != len_b) {
+        return false;
+    }
+
+    /* If either is null: true if both are null or the length is 0 */
+    if (!array_a || !array_b) {
```
justinboswell: if (array_a == array_b) is always equivalent, null or not.

justinboswell: In fact, you could do this (perhaps more simply) as:
```c
if (array_a == array_b && len_a == len_b)
```

source/byte_buf.c
```diff
@@ -242,32 +259,102 @@ static const uint8_t s_tolower_table[256] = {
...
+        if (s_tolower_table[array_bytes[i]] != s_tolower_table[s]) {
+            return false;
+        }
+    }
+
+    return str_bytes[array_len] == '\0';
+}
+
+bool aws_array_eq_c_str(const void *array, size_t array_len, const char *c_str) {
+    const uint8_t *array_bytes = array;
```
justinboswell: Why doesn't this strlen() and delegate to ```aws_array_eq()```?

source/byte_buf.c
```diff
@@ -242,32 +259,102 @@ static const uint8_t s_tolower_table[256] = {
...
+    /* If either is null: true if both are null or the length is 0 */
+    if (!array_a || !array_b) {
+        return (array_a == array_b) || (len_a == 0);
+    }
+
+    return !memcmp(array_a, array_b, len_a);
+}
+
+bool aws_array_eq_c_str_ignore_case(const void *array, size_t array_len, const char *c_str) {
+    const uint8_t *array_bytes = array;
```
justinboswell: Why doesn't this strlen() and delegate to ```aws_array_eq_ignore_case```?

tests/byte_buf_test.c
```diff
@@ -320,42 +320,84 @@ static int s_test_buffer_printf(struct aws_allocator *allocator, void *ctx) {
...
+        uint8_t c[] = {'a', 'b', 'c', 'd', '9'}; /* different */
+
+        /* Simple */
+        ASSERT_TRUE(aws_array_eq_ignore_case(a, 5, b, 5));
+        ASSERT_FALSE(aws_array_eq_ignore_case(a, 5, c, 5));
+        ASSERT_FALSE(aws_array_eq_ignore_case(a, 5, b, 3));
+
+        /* Comparisons against self */
+        ASSERT_TRUE(aws_array_eq_ignore_case(a, 5, a, 5));
+        ASSERT_FALSE(aws_array_eq_ignore_case(a, 5, a, 4));
```
justinboswell: Ooh, this is an interesting case. It would be good to add an additional assert for 2 different arrays, with different lengths, just to verify that path.

tautschnig:
The failing CBMC jobs are due to failing PutMetricData operations. I have re-submitted the request to clean this up.

source/byte_buf.c
graebm: Calling strlen() means traversing the c-str an additional time.

This function is very copy/pasty similar to the other one, but checks for null-terminators in the c_str while doing the comparison, so there's only one traversal.

I could benchmark the two implementations to see if there's measurable difference.

Thoughts?

source/byte_buf.c
graebm: The way it's written, it takes this branch if either is NULL, so there's no chance of dereferencing a nullptr later

the way you wrote it, if one was NULL, it would bypass this check and dereference the nullptr when it entered the for-loop later

source/byte_buf.c
justinboswell: Oh, yeah, of course. That makes sense, missed that one :)

source/byte_buf.c
justinboswell: 🙅‍♂️

include/aws/common/string.h
```diff
@@ -68,27 +68,91 @@ AWS_STATIC_IMPL const uint8_t *aws_string_bytes(const struct aws_string *str) {
  * Returns true if bytes of string are the same, false otherwise.
  */
 AWS_STATIC_IMPL bool aws_string_eq(const struct aws_string *a, const struct aws_string *b) {
-    return a->len == b->len && !memcmp(a->bytes, b->bytes, a->len);
+    const uint8_t *a_ptr = a ? a->bytes : NULL;
```
bretambrose: It seems like a handful of macros could reduce these to one-liners

#define STRING_TO_ARRAY_PARAMS(str) ((str) ? (str->bytes) : NULL), ((str) ? (str->len) : 0)
 
etc...

source/byte_buf.c
graebm: Did some benchmarks and it is worth avoiding the strlen.
- 2X speedup when avoiding it in `aws_array_eq_cstr()` (uses memcmp under the hood)
- 1.5X speedup when avoiding it in `aws_array_eq_c_str_ignore_case()` (uses my crazy table under the hood)

include/aws/common/byte_buf.h
graebm: NULL is no longer accepted when comparing aws_byte_cursor, aws_byte_buf, aws_string, c_str.

It **is** still accepted in the underlaying aws_array_eq functions if the length is zero and I commented that like:

"NULL may be passed as the array pointer if its length is declared to be 0."

include/aws/common/string.h
graebm: I got rid of these lines in the latest versions

PR #297, opened 2019-04-02
Byte buf utility functions - append that grows, append with internal tolower, reserve

bretambrose:
Added a byte_buf append function that attempts to grow when size limit is exceeded.


source/byte_buf.c
```diff
@@ -380,6 +380,75 @@ int aws_byte_buf_append(struct aws_byte_buf *to, const struct aws_byte_cursor *f
...
+         * Then do a final check before allocating to verify that the computed
+         * new capacity is actually containable within the platform's size_t type.
+         *
+         * If size_t ever exceeds 64 bits, this won't be valid
+         */
+
+        /*
+         * NewCapacity = Max(OldCapacity * 1.5, OldCapacity + MissingCapacity)
+         */
+        uint64_t missing_capacity = (uint64_t)(from->len - (to->capacity - to->len));
```
graebm: aws_array_list_ensure_capacity() doubles capacity, we might as well do the same

source/byte_buf.c
```diff
@@ -380,6 +380,75 @@ int aws_byte_buf_append(struct aws_byte_buf *to, const struct aws_byte_cursor *f
...
 }
 
+int aws_byte_buf_append_dynamic(struct aws_byte_buf *to, const struct aws_byte_cursor *from) {
+    assert(from->ptr);
+    assert(to->buffer);
+
+    if (to->capacity - to->len < from->len) {
+        /*
+         * We don't have size_t math so use 64 bits for overflow checking.
+         * Then do a final check before allocating to verify that the computed
```
graebm: yes we do: aws_add_size_checked()

source/byte_buf.c
```diff
@@ -380,6 +380,75 @@ int aws_byte_buf_append(struct aws_byte_buf *to, const struct aws_byte_cursor *f
...
+         */
+        uint64_t missing_capacity = (uint64_t)(from->len - (to->capacity - to->len));
+        uint64_t new_capacity = 0;
+        if (aws_add_u64_checked(to->capacity, missing_capacity, &new_capacity)) {
+            return AWS_OP_ERR;
+        }
+
+        uint64_t growth_capacity = 0;
+        if (aws_add_u64_checked(to->capacity, to->capacity / 2, &growth_capacity)) {
+            return AWS_OP_ERR;
```
graebm: if growth_capacity overflows, we could still try using the one based on missing_capacity

source/byte_buf.c
```diff
@@ -380,6 +380,75 @@ int aws_byte_buf_append(struct aws_byte_buf *to, const struct aws_byte_cursor *f
...
+            new_capacity = growth_capacity;
+        }
+
+        if (new_capacity > SIZE_MAX) {
+            return aws_raise_error(AWS_ERROR_OVERFLOW_DETECTED);
+        }
+
+        uint8_t *new_buffer = aws_mem_acquire(to->allocator, (size_t)new_capacity);
+        if (new_buffer == NULL) {
+            return AWS_OP_ERR;
```
graebm: if we can't acquire enough for growth_capacity, and the necessary size was less, we could try that instead.

source/byte_buf.c
```diff
@@ -380,6 +384,118 @@ int aws_byte_buf_append(struct aws_byte_buf *to, const struct aws_byte_cursor *f
...
+
+        /*
+         * Attempt to resize - we intentionally do not use reserve() in order to preserve
+         * the (unlikely) use case of from and to being the same buffer range.
+         */
+
+        /*
+         * Try the max, but if that fails and the required is smaller, try it in fallback
+         */
+        uint8_t *new_buffer = aws_mem_acquire(to->allocator, new_capacity);
```
graebm: You could use `aws_mem_realloc()`. It already does a lot of this copy-from-old-to-new-buffer stuff for you.
Also, if it could theoretically be faster in the case that the existing allocation can simply grow in place

... ughh
but that wouldn't work if `from` used memory from `to`. You'd need to detect that cause and do pointer math to correct the from pointer

source/byte_buf.c
```diff
@@ -380,6 +384,118 @@ int aws_byte_buf_append(struct aws_byte_buf *to, const struct aws_byte_cursor *f
...
+        if (new_buffer == NULL) {
+            if (new_capacity > required_capacity) {
+                new_capacity = required_capacity;
```
graebm: need to return ERR if that 2nd if fails

source/byte_buf.c
```diff
@@ -380,6 +384,118 @@ int aws_byte_buf_append(struct aws_byte_buf *to, const struct aws_byte_cursor *f
...
+
+    return AWS_OP_SUCCESS;
+}
+
+int aws_byte_buf_reserve(struct aws_byte_buf *buffer, size_t requested_capacity) {
+    if (requested_capacity <= buffer->capacity) {
+        return AWS_OP_SUCCESS;
+    }
+
+    uint8_t *new_buffer = aws_mem_acquire(buffer->allocator, requested_capacity);
```
graebm: reserve() is an excellent candidate to use realloc() under the hood

include/aws/common/byte_buf.h
```diff
@@ -361,6 +395,13 @@ uint64_t aws_hash_array_ignore_case(const void *array, size_t len);
 AWS_COMMON_API
 uint64_t aws_hash_byte_cursor_ptr_ignore_case(const void *item);
 
+/**
+ * Returns the lookup table used internally to do case-insensitive comparisons.  Allows it to be
+ * used in the lookup version of byte_buf append.
+ */
+AWS_COMMON_API
+const uint8_t *aws_lookup_table_to_lower_get(void);
```
graebm: [trivial] "Returns table where uppercase ascii characters are replaced by lowercase. Can be used with..."
instead of explaining how its used internally

also, I'd be fine with leaving "_get" off the end of the name. but that's just, like, my opinion man

include/aws/common/byte_buf.h
```diff
@@ -113,6 +113,18 @@ bool aws_array_eq_c_str_ignore_case(const void *array, size_t array_len, const c
 AWS_COMMON_API
 int aws_byte_buf_init(struct aws_byte_buf *buf, struct aws_allocator *allocator, size_t capacity);
 
+/**
+ * Set of properties of a valid aws_byte_buf.
+ */
+AWS_COMMON_API
+bool is_valid_byte_buf(const struct aws_byte_buf *buf);
```
graebm: aws_byte_buf_is_valid()

PR #304, opened 2019-04-10
byte_buf append w/lookup does not alias

marcomagdy:
*Description of changes:*
Mark the arguments with `restrict` keyword to generate better assembly when strict-aliasing is enabled.


ColdenCullen:
Does this break appending a buffer to itself?

marcomagdy:
> Does this break appending a buffer to itself?

The documentation https://github.com/awslabs/aws-c-common/pull/304/files#diff-b6bf40e2dd7c2a450ad1dc0080df6979R309 says that they should not overlap.

>... from and to should not be the same buffer (overlap is not handled) lookup_table must be at least 256 bytes

PR #307, opened 2019-04-10
CBMC got stricter, so touch up the stubs to be correct

danielsn:
*Issue #, if available:*

*Description of changes:*
This should fix the issues surfacing in https://github.com/awslabs/aws-c-common/pull/306


.cbmc-batch/jobs/aws_byte_buf_append_dynamic/cbmc-batch.yaml
```diff
@@ -1,4 +1,4 @@
 jobos: ubuntu16
-cbmcflags: "--bounds-check;--pointer-check;--div-by-zero-check;--signed-overflow-check;--unsigned-overflow-check;--pointer-overflow-check;--undefined-shift-check;--float-overflow-check;--nan-check;--unwinding-assertions;--function;aws_byte_buf_append_dynamic_harness"
+cbmcflags: "--bounds-check;--div-by-zero-check;--float-overflow-check;--nan-check;--pointer-check;--pointer-overflow-check;--signed-overflow-check;--undefined-shift-check;--unsigned-overflow-check;--unwind;1;--unwinding-assertions;--object-bits;8"
```
tautschnig: Why are the changes in those two .yaml files necessary? Removing `--function` and adding `--unwind` seem like major changes. And isn't `--object-bits 8` the default via your Makefile.common changes?

danielsn: These files are auto-generated by the Makefile.common  So something must have changed in Makefile.common between when this test was first written, and now.  My preference would be to merge this PR and then take a separate issue to investigate the Makefile.common flags and make sure they're the best ones.

PR #325, opened 2019-04-30
Fixes implementation of aws_byte_buf_init_copy function

feliperodri:

*Description of changes:*

Previous implementation did not copy the capacity field from src structure. In addition, it allowed the construction of an `aws_byte_buf` structure with a `buffer` field set to `NULL`.

### DIFF
--- a/.cbmc-batch/jobs/aws_byte_buf_init_copy/aws_byte_buf_init_copy_harness.c
+++ b/.cbmc-batch/jobs/aws_byte_buf_init_copy/aws_byte_buf_init_copy_harness.c
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/byte_buf.h>
+#include <proof_helpers/make_common_data_structures.h>
+#include <proof_helpers/proof_allocators.h>
+
+void aws_byte_buf_init_copy_harness() {
+    /* data structure */
+    struct aws_byte_buf *dest;
+
+    /* parameters */
+    struct aws_allocator *allocator;
+    struct aws_byte_buf src;
+
+    /* assumptions */
+    __CPROVER_assume(is_bounded_byte_buf(&src, MAX_BUFFER_SIZE));
+    ensure_byte_buf_has_allocated_buffer_member(&src);
+    __CPROVER_assume(aws_byte_buf_is_valid(&src));
+
+    ASSUME_VALID_MEMORY(dest);
+    ASSUME_CAN_FAIL_ALLOCATOR(allocator);
+
+    /* operation under verification */
+    if (!aws_byte_buf_init_copy(dest, allocator, &src)) {
+        /* assertions */
+        assert(aws_byte_buf_is_valid(dest));
+        assert(is_byte_buf_expected_alloc(dest));
+        assert(dest->len == src.len);
+        assert(dest->capacity == src.capacity);
+        assert_bytes_match(dest->buffer, src.buffer, dest->len);
+        assert(aws_byte_buf_is_valid(&src));
+    }
+}
--- a/.cbmc-batch/jobs/aws_byte_buf_init_copy/aws_byte_buf_init_copy_harness.c
+++ b/.cbmc-batch/jobs/aws_byte_buf_init_copy/aws_byte_buf_init_copy_harness.c
@@ -29,9 +29,13 @@ void aws_byte_buf_init_copy_harness() {
     __CPROVER_assume(is_bounded_byte_buf(&src, MAX_BUFFER_SIZE));
     ensure_byte_buf_has_allocated_buffer_member(&src);
     __CPROVER_assume(aws_byte_buf_is_valid(&src));
-
     ASSUME_VALID_MEMORY(dest);
     ASSUME_CAN_FAIL_ALLOCATOR(allocator);
+
+    /* save current state of the data structure */
+    struct aws_byte_buf old = src;
+    struct store_byte_from_buffer old_byte;
+    save_byte_from_array(src.buffer, src.len, &old_byte);
 
     /* operation under verification */
     if (!aws_byte_buf_init_copy(dest, allocator, &src)) {
@@ -42,5 +46,9 @@ void aws_byte_buf_init_copy_harness() {
         assert(dest->capacity == src.capacity);
         assert_bytes_match(dest->buffer, src.buffer, dest->len);
         assert(aws_byte_buf_is_valid(&src));
+        if (src.len > 0) {
+            assert_byte_from_buffer_matches(src.buffer, &old_byte);
+            assert_byte_from_buffer_matches(dest->buffer, &old_byte);
+        }
     }
 }
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -114,10 +114,18 @@ AWS_COMMON_API
 int aws_byte_buf_init(struct aws_byte_buf *buf, struct aws_allocator *allocator, size_t capacity);
 
 /**
+ * Initializes *dest by copying all the fields from *src.
+ */
+AWS_COMMON_API int aws_byte_buf_init_copy(
+    struct aws_byte_buf *dest,
+    struct aws_allocator *allocator,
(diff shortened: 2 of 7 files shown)


.cbmc-batch/jobs/aws_byte_buf_init_copy/aws_byte_buf_init_copy_harness.c
```diff
@@ -0,0 +1,46 @@
...
+
+    /* parameters */
+    struct aws_allocator *allocator;
+    struct aws_byte_buf src;
+
+    /* assumptions */
+    __CPROVER_assume(is_bounded_byte_buf(&src, MAX_BUFFER_SIZE));
+    ensure_byte_buf_has_allocated_buffer_member(&src);
+    __CPROVER_assume(aws_byte_buf_is_valid(&src));
+
```
danielsn: You could grab an old byte from ```src``` and make sure it didn't change.

include/aws/common/byte_buf.h
```diff
@@ -113,11 +113,19 @@ bool aws_array_eq_c_str_ignore_case(const void *array, size_t array_len, const c
 AWS_COMMON_API
 int aws_byte_buf_init(struct aws_byte_buf *buf, struct aws_allocator *allocator, size_t capacity);
 
+/**
+ * Initializes *dest by copying all the fields from *src.
```
danielsn: Better comment here: what are the requirements? When can it fail? it doesn't just copy the fields, it allocates a new backing array.

include/aws/common/byte_buf.h
```diff
@@ -113,11 +113,19 @@ bool aws_array_eq_c_str_ignore_case(const void *array, size_t array_len, const c
...
+ */
+AWS_COMMON_API int aws_byte_buf_init_copy(
+    struct aws_byte_buf *dest,
+    struct aws_allocator *allocator,
+    const struct aws_byte_buf *src);
+
 /**
  * Set of properties of a valid aws_byte_buf.
```
danielsn: This is a best effort check, right?

source/byte_buf.c
```diff
@@ -35,8 +35,32 @@ int aws_byte_buf_init(struct aws_byte_buf *buf, struct aws_allocator *allocator,
...
-bool aws_byte_buf_is_valid(const struct aws_byte_buf *buf) {
-    return (buf->len <= buf->capacity) && (buf->allocator != NULL) && AWS_MEM_IS_WRITABLE(buf->buffer, buf->len);
+int aws_byte_buf_init_copy(struct aws_byte_buf *dest, struct aws_allocator *allocator, const struct aws_byte_buf *src) {
+    if (!allocator || !dest || !aws_byte_buf_is_valid(src)) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+
+    *dest = *src;
+    dest->allocator = allocator;
+    dest->buffer = (uint8_t *)aws_mem_acquire(allocator, src->capacity);
```
danielsn: Is this cast needed?

source/byte_buf.c
```diff
@@ -89,26 +113,25 @@ int aws_byte_buf_init_copy_from_cursor(
...
-        return AWS_OP_SUCCESS;
+    if (!allocator || !dest || !aws_byte_cursor_is_valid(&src)) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
     }
 
-    dest->buffer = (uint8_t *)aws_mem_acquire(allocator, sizeof(uint8_t) * src.len);
+    /* clear the dest before updating it. */
+    *dest = (struct aws_byte_buf){0};
+
+    dest->buffer = (uint8_t *)aws_mem_acquire(allocator, src.len);
```
danielsn: Is this cast necessary?

include/aws/common/byte_buf.h
```diff
@@ -114,13 +114,26 @@ AWS_COMMON_API
...
+ * It is also a cheap check, in the sense it run in constant time (i.e., no loops or recursion).
  */
 AWS_COMMON_API
-bool aws_byte_buf_is_valid(const struct aws_byte_buf *buf);
+bool aws_byte_buf_is_valid(const struct aws_byte_buf *const buf);
 
 /**
- * Set of properties of a valid aws_byte_cursor.
+ * Evaluates the set of properties that define the shape of all valid aws_byte_cursor structures.
+ * It is also a cheap check, in the sense it run in constant time (i.e., no loops or recursion).
```
bretambrose: "it is run" or "that it runs"

source/byte_buf.c
```diff
@@ -35,8 +35,33 @@ int aws_byte_buf_init(struct aws_byte_buf *buf, struct aws_allocator *allocator,
...
+int aws_byte_buf_init_copy(struct aws_byte_buf *dest, struct aws_allocator *allocator, const struct aws_byte_buf *src) {
+    if (!allocator || !dest || !aws_byte_buf_is_valid(src)) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+
+    *dest = *src;
+    dest->allocator = allocator;
+    dest->buffer = (uint8_t *)aws_mem_acquire(allocator, src->capacity);
+    if (dest->buffer == NULL) {
+        *dest = (struct aws_byte_buf){0};
```
bretambrose: Is this needed for proofs or can it be replaced by AWS_ZERO_STRUCT?

source/byte_buf.c
```diff
@@ -89,26 +114,25 @@ int aws_byte_buf_init_copy_from_cursor(
...
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
     }
 
-    dest->buffer = (uint8_t *)aws_mem_acquire(allocator, sizeof(uint8_t) * src.len);
+    /* clear the dest before updating it. */
+    *dest = (struct aws_byte_buf){0};
+
+    dest->buffer = (uint8_t *)aws_mem_acquire(allocator, src.len);
     if (dest->buffer == NULL) {
+        AWS_POSTCONDITION(aws_byte_cursor_is_valid(&src));
```
bretambrose: Not sure this is worth doing when src is passed by value.

source/byte_buf.c
```diff
@@ -89,26 +114,25 @@ int aws_byte_buf_init_copy_from_cursor(
...
-    if (src.ptr == NULL) {
-        dest->buffer = NULL;
-        return AWS_OP_SUCCESS;
+    if (!allocator || !dest || !aws_byte_cursor_is_valid(&src)) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
     }
 
-    dest->buffer = (uint8_t *)aws_mem_acquire(allocator, sizeof(uint8_t) * src.len);
+    /* clear the dest before updating it. */
+    *dest = (struct aws_byte_buf){0};
```
bretambrose: AWS_ZERO_STRUCT preferred for clarity if possible

source/byte_buf.c
```diff
@@ -89,26 +114,25 @@ int aws_byte_buf_init_copy_from_cursor(
...
     if (dest->buffer == NULL) {
+        AWS_POSTCONDITION(aws_byte_cursor_is_valid(&src));
         return AWS_OP_ERR;
     }
 
     dest->len = src.len;
     dest->capacity = src.len;
     dest->allocator = allocator;
     memcpy(dest->buffer, src.ptr, src.len);
+    AWS_POSTCONDITION(aws_byte_cursor_is_valid(&src));
```
bretambrose: similar comment here

source/byte_buf.c
```diff
@@ -35,8 +35,33 @@ int aws_byte_buf_init(struct aws_byte_buf *buf, struct aws_allocator *allocator,
     return AWS_OP_SUCCESS;
 }
 
-bool aws_byte_buf_is_valid(const struct aws_byte_buf *buf) {
-    return (buf->len <= buf->capacity) && (buf->allocator != NULL) && AWS_MEM_IS_WRITABLE(buf->buffer, buf->len);
+int aws_byte_buf_init_copy(struct aws_byte_buf *dest, struct aws_allocator *allocator, const struct aws_byte_buf *src) {
```
bretambrose: Why did we unlink _copy from _copy_from_cursor?  Did that cause proof issues?  This also leads to a slight change in semantics because now capacity mirrors the original buffer's capacity rather than just the length.  

I was about to argue that I preferred the old approach rather than the new approach but then I changed my mind: if you want a shrunk copy use the cursor interface, if you want an exact copy use the byte buf interface.  I would like some additional reviewers to approve that change though.

feliperodri: The point to unlink `_copy` from `_copy_from_cursor` is the same that made you change your mind, i.e., if we want a exact copy use `_init_copy`, otherwise just use `_copy_from_cursor`

danielsn: To give some more background: our new proof style involves determining the invariants of the function, and then using CBMC to prove those hold.  For ```init_copy``` we had expected the result to be an exact copy of the initial, and were surprised when its backing array was smaller.  IMO, the new behaviour follows the principle of least surprise.

source/byte_buf.c
danielsn: Nice catch.

source/byte_buf.c
danielsn: Nice catch.

PR #326, opened 2019-04-30
Implement common codebuild generator/runner

ColdenCullen:


codebuild/common.yml
```diff
@@ -0,0 +1,11 @@
+version: 0.2
+#this buildspec assumes the ubuntu 14.04 trusty image
```
justinboswell: This isn't true anymore, surely :)

codebuild/builder.py
```diff
@@ -0,0 +1,693 @@
...
+        'account': github_account,
+        'spec': config['spec'].name(),
+    }
+
+    # This matches the CodeBuild API for expected format
+    CREATE_PARAM_TEMPLATE = {
+        'name': '{project}-{spec}',
+        'source': {
+            'type': 'GITHUB',
+            'location': 'https://github.com/{account}/{project}.git',
```
JonathanHenson: this is good, helps the GA switch be much smoother.

codebuild/common.yml
```diff
@@ -0,0 +1,11 @@
+version: 0.2
+phases:
+  install:
+    commands:
+# To be uncommented when the file is properly updated in master
+#      - python3 -c "from urllib.request import urlretrieve; urlretrieve('https://raw.githubusercontent.com/awslabs/aws-c-common/master/codebuild/builder.py', 'builder.py')"
```
JonathanHenson: Can we just rebuild the images to have elasticurl?

ColdenCullen: Paging mr @justinboswell

codebuild/builder.py
```diff
@@ -0,0 +1,731 @@
...
+########################################################################################################################
+
+def run_build(build_spec, is_dryrun):
+
+    if not is_dryrun:
+        import tempfile, shutil, subprocess
+
+    #TODO These platforms don't succeed when doing a RelWithDebInfo build
+    if build_spec.host in ("al2012", "manylinux"):
+        global BUILD_CONFIG
```
justinboswell: Why not just cache it into a local and not mess with the global?

ColdenCullen: This way when we fix these issues, I can just nuke this chunk and the problem goes away

PR #328, opened 2019-05-01
Hex encoding and byte buf utilities

bretambrose:
New hex encoding variant that does not 0-terminate, resizes as needed, and respects the buffers current position.

New reserve invariant that performs its resizing calculation relative to the current length.  Useful to cut down on the number of checked arithmetic operations needed when using reserve.


include/aws/common/byte_buf.h
```diff
@@ -348,6 +348,16 @@ int aws_byte_buf_append_dynamic(struct aws_byte_buf *to, const struct aws_byte_c
...
 
+/**
+ * Convenience function that attempts to increase the capacity of a buffer relative to the current
+ * length.
+ *
+ *  aws_byte_buf_reserve_from_len(buf, x) ~~ aws_byte_buf_reserve(buf, buf->len + x)
+ *
+ */
+AWS_COMMON_API
+int aws_byte_buf_reserve_from_len(struct aws_byte_buf *buffer, size_t additional_length);
```
justinboswell: [nitpick] Feels like _grow or _expand or _extend, do we not have this pattern anywhere else? reserve_from_len just feels a bit unintuitive to me offhand.

bretambrose: _reserve_relative was another possibility

PR #334, opened 2019-05-06
CBMC proof for hash_table_swap()

danielsn:
*Description of changes:*
CBMC proof for ```hash_table_swap()```.  

Due to the fact that either input to swap can be invalid, it's hard to come up with any good post-conditions for this function.  It might be possible to track some using ghost state, but that would put unnecessary overhead on the function itself.

### DIFF
--- a/.cbmc-batch/jobs/aws_hash_table_swap/aws_hash_table_swap_harness.c
+++ b/.cbmc-batch/jobs/aws_hash_table_swap/aws_hash_table_swap_harness.c
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use
+ * this file except in compliance with the License. A copy of the License is
+ * located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an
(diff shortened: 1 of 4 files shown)


source/hash_table.c
```diff
@@ -231,6 +231,7 @@ void aws_hash_table_clean_up(struct aws_hash_table *map) {
 }
 
 void aws_hash_table_swap(struct aws_hash_table *AWS_RESTRICT a, struct aws_hash_table *AWS_RESTRICT b) {
+    AWS_PRECONDITION(a != b);
```
feliperodri:
```
AWS_PRECONDITION(aws_hash_table_is_valid(a));
AWS_PRECONDITION(aws_hash_table_is_valid(b));
```
?

danielsn:
I had the same thought at first. But the spec for swap requires neither
input to be valid as per the .h file.

.cbmc-batch/jobs/aws_hash_table_swap/aws_hash_table_swap_harness.c
```diff
@@ -0,0 +1,53 @@
...
+
+void aws_hash_table_swap_harness() {
+    struct aws_hash_table a;
+    struct aws_hash_table b;
+    bool inita;
+    bool initb;
+    struct store_byte_from_buffer stored_byte_a;
+    struct store_byte_from_buffer stored_byte_b;
+
+    // There are no loops in the code under test, so use the biggest possible value
```
JonathanHenson: c++ comments...

danielsn: Fixed.

PR #338, opened 2019-05-07
Implement AWS_BYTE_CUR_INIT_FROM_LITERAL

ColdenCullen:
This is mostly for unit testing stuff.

### DIFF
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -441,6 +441,11 @@ const uint8_t *aws_lookup_table_to_lower_get(void);
 AWS_EXTERN_C_END
 
 /**
+ */
+#define AWS_BYTE_CUR_INIT_FROM_LITERAL(literal)                                                                  \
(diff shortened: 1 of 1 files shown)


ColdenCullen:
Hang on this isn't as useful as I thought it would be

ColdenCullen:
Changed my mind, rather than defining a static, it's more useful to have an init macro.

include/aws/common/byte_buf.h
```diff
@@ -440,6 +440,11 @@ const uint8_t *aws_lookup_table_to_lower_get(void);
 
 AWS_EXTERN_C_END
 
+/**
+ */
+#define AWS_BYTE_CUR_INIT_FROM_LITERAL(literal)                                                                  \
+    { .ptr = (uint8_t *)(literal), .len = AWS_ARRAY_SIZE(literal) }
```
graebm: won't AWS_ARRAY_SIZE of "a" be 2? because of the '\0'?

ColdenCullen: Yes! That's probably not what we want here huh.

PR #340, opened 2019-05-08
Proofs for hash_table_iterator functions

danielsn:
*Description of changes:*
1. Added proofs for ```hash_iter*``` functions
1. As part of this, made explicit the state machine that hash iterators are expected to follow
1. ~~In order to keep ABI compatability, I use an ```intptr_t``` to store the new enum.  If this is not essential, I'd rather use an ```enum``` to store it, to make the intent clearer.~~

### DIFF
--- a/.cbmc-batch/include/proof_helpers/make_common_data_structures.h
+++ b/.cbmc-batch/include/proof_helpers/make_common_data_structures.h
@@ -133,3 +133,19 @@ void ensure_allocated_hash_table(struct aws_hash_table *map, size_t max_table_en
  * Makes a valid c string, with as much nondet as possible, len < max
  */
 const char *make_arbitrary_c_str(size_t max_size);
+
+/**
+ * A correct hash table has max_load < size.  This means that there is always one slot empty.
+ * This function is useful for assuming that there is some (nondet) slot which is empty
+ * which is necessary to prove termination for hash-table deletion code.  Should only be used inside
+ * an assume because of the way it does nondet.
+ */
+bool aws_hash_table_has_an_empty_slot(struct aws_hash_table *map, size_t *rval);
+
+/**
+ * A correct implementation of the hash_destroy function should never have a memory
+ * error on valid input. There is the question of whether the destroy functions themselves
+ * are correctly called (i.e. only on valid input, no double free, etc.).  Testing this would
+ * require a stronger function here.
+ */
+void hash_proof_destroy_noop(void *p);
--- a/.cbmc-batch/include/proof_helpers/make_common_data_structures.h
+++ b/.cbmc-batch/include/proof_helpers/make_common_data_structures.h
@@ -140,7 +140,7 @@ const char *make_arbitrary_c_str(size_t max_size);
  * which is necessary to prove termination for hash-table deletion code.  Should only be used inside
  * an assume because of the way it does nondet.
  */
-bool aws_hash_table_has_an_empty_slot(struct aws_hash_table *map, size_t *rval);
+bool aws_hash_table_has_an_empty_slot(const struct aws_hash_table *const map, size_t *rval);
 
 /**
  * A correct implementation of the hash_destroy function should never have a memory
--- a/.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
+++ b/.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
@@ -12,9 +12,9 @@
 # limitations under the License.
 
 ###########
-#2: 1M 15s
-#4:  4m 40s
-MAX_TABLE_SIZE ?= 2
+#2: 1M 15s (tho note that this doens't give full coverage)
+#4:  4m 40s (gives full coverage)
+MAX_TABLE_SIZE ?= 4
 DEFINES += -DMAX_TABLE_SIZE=$(MAX_TABLE_SIZE) 
 
 UNWINDSET += s_remove_entry.0:$(shell echo $$((1 + $(MAX_TABLE_SIZE))))
--- a/.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
+++ b/.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
@@ -14,7 +14,7 @@
 ###########
 #2: 1M 15s
 #4:  4m 40s
-MAX_TABLE_SIZE ?= 4
+MAX_TABLE_SIZE ?= 2
 DEFINES += -DMAX_TABLE_SIZE=$(MAX_TABLE_SIZE) 
 
 UNWINDSET += s_remove_entry.0:$(shell echo $$((1 + $(MAX_TABLE_SIZE))))
--- a/.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
+++ b/.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
@@ -12,8 +12,7 @@
 # limitations under the License.
 
 ###########
-#2: 1M 15s (tho note that this doens't give full coverage)
-#4:  4m 40s (gives full coverage)
+#4: 4m 40s (smallest size that gives full coverage)
 MAX_TABLE_SIZE ?= 4
 DEFINES += -DMAX_TABLE_SIZE=$(MAX_TABLE_SIZE) 
 
--- a/.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
+++ b/.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
@@ -12,7 +12,7 @@
 # limitations under the License.
 
 ###########
-#4: 4m 40s (smallest size that gives full coverage)
+#4: 4m40s (smallest size that gives full coverage)
 MAX_TABLE_SIZE ?= 4
 DEFINES += -DMAX_TABLE_SIZE=$(MAX_TABLE_SIZE) 
 
--- a/.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
+++ b/.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
@@ -1,0 +1,34 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use
+# this file except in compliance with the License. A copy of the License is
+# located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+###########
+#2: 1M 15s
+#4:  4m 40s
+MAX_TABLE_SIZE ?= 4
+DEFINES += -DMAX_TABLE_SIZE=$(MAX_TABLE_SIZE) 
+
+UNWINDSET += s_remove_entry.0:$(shell echo $$((1 + $(MAX_TABLE_SIZE))))
+
+CBMCFLAGS +=
+
+DEPENDENCIES += $(HELPERDIR)/source/make_common_data_structures.c
+DEPENDENCIES += $(HELPERDIR)/source/proof_allocators.c
+DEPENDENCIES += $(HELPERDIR)/source/utils.c
+DEPENDENCIES += $(HELPERDIR)/stubs/error.c
+DEPENDENCIES += $(SRCDIR)/source/common.c
+DEPENDENCIES += $(SRCDIR)/source/hash_table.c
+
+ENTRY = aws_hash_iter_delete_harness
+###########
+
+include ../Makefile.common
--- a/.cbmc-batch/jobs/aws_hash_iter_delete/aws_hash_iter_delete_harness.c
+++ b/.cbmc-batch/jobs/aws_hash_iter_delete/aws_hash_iter_delete_harness.c
@@ -19,23 +19,6 @@
 #include <proof_helpers/proof_allocators.h>
 #include <proof_helpers/utils.h>
 
-bool has_an_empty_slot(struct aws_hash_table *map, size_t *rval) {
-    struct hash_table_state *state = map->p_impl;
-    __CPROVER_assume(state->entry_count > 0);
-    size_t empty_slot_idx;
-    __CPROVER_assume(empty_slot_idx < state->size);
-    *rval = empty_slot_idx;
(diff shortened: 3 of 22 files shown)


include/aws/common/hash_table.h
```diff
@@ -74,18 +75,30 @@ struct aws_hash_element {
...
+ */
+AWS_STATIC_ASSERT(sizeof(intptr_t) == sizeof(void *));
+
 struct aws_hash_iter {
     const struct aws_hash_table *map;
     struct aws_hash_element element;
     size_t slot;
     size_t limit;
+    /* Holds an enum aws_hash_iter_status */
+    intptr_t status;
```
justinboswell: enums are ints underneath, you could leave an int reserved and only use an int for status

danielsn: Done.

.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
```diff
@@ -0,0 +1,34 @@
...
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+###########
+#2: 1M 15s (tho note that this doens't give full coverage)
+#4:  4m 40s (gives full coverage)
```
feliperodri:
```
#2: 1m15s (tho note that this doens't give full coverage)
#4: 4m40s (gives full coverage)
```

.cbmc-batch/jobs/aws_hash_iter_delete/aws_hash_iter_delete_harness.c
```diff
@@ -0,0 +1,58 @@
...
+ * limitations under the License.
+ */
+
+#include <aws/common/hash_table.h>
+#include <aws/common/private/hash_table_impl.h>
+#include <proof_helpers/make_common_data_structures.h>
+#include <proof_helpers/proof_allocators.h>
+#include <proof_helpers/utils.h>
+
+bool has_an_empty_slot(struct aws_hash_table *map, size_t *rval) {
```
feliperodri: Will it always be used only here? If not, it should be in `utils.h`

.cbmc-batch/jobs/aws_hash_iter_delete/aws_hash_iter_delete_harness.c
```diff
@@ -0,0 +1,58 @@
...
+    return state->slots[empty_slot_idx].hash_code == 0;
+}
+
+/**
+ * A correct implementation of the hash_destroy function should never have a memory
+ * error on valid input. There is the question of whether the destroy functions themselves
+ * are correctly called (i.e. only on valid input, no double free, etc.).  This will be tested in
+ * future proofs.
+ */
+void hash_proof_destroy_noop(void *p) {}
```
feliperodri: Will it always be used only here? If not, it should be in `utils.h`

.cbmc-batch/jobs/aws_hash_iter_done/Makefile
```diff
@@ -0,0 +1,30 @@
...
+#
+# or in the "license" file accompanying this file. This file is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+###########
+DEFINES += -DMAX_TABLE_SIZE=SIZE_MAX 
+
+
```
feliperodri: Remove extra line

.cbmc-batch/jobs/aws_hash_iter_done/Makefile
```diff
@@ -0,0 +1,30 @@
...
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+###########
+DEFINES += -DMAX_TABLE_SIZE=SIZE_MAX
```
feliperodri: Why do we need to bound to `SIZE_MAX`? I'd expect that removing this line would have no affect in the proof.

source/hash_table.c
```diff
@@ -53,7 +53,10 @@ static uint64_t s_hash_for(struct hash_table_state *state, const void *key) {
 }
 
 static size_t s_index_for(struct hash_table_state *map, struct hash_table_entry *entry) {
-    return (size_t)(entry - map->slots);
+    AWS_PRECONDITION(hash_table_state_is_valid(map));
+    size_t index = entry - map->slots;
+    AWS_POSTCONDITION(index < map->size);
```
feliperodri: No need for `AWS_POSTCONDITION(hash_table_state_is_valid(map))`?

source/hash_table.c
```diff
@@ -642,8 +650,17 @@ bool aws_hash_table_eq(
...
+ * Otherwise, return an iter that will return true for aws_hash_iter_done().
+ * Note that aws_hash_iter_is_valid() need not hold on entry to the function, since
+ * it can be called on a partially constructed iter from aws_hash_iter_begin().
+ *
+ * Note that calling this on an iterator which is "done" is idempotent: it will return another
+ * iterator which is "done".
+ */
 static inline void s_get_next_element(struct aws_hash_iter *iter, size_t start_slot) {
-
+    AWS_PRECONDITION(iter != NULL);
```
feliperodri: No need for `AWS_PRECONDITION(aws_hash_iter_is_valid(iter))`?

source/hash_table.c
```diff
@@ -653,38 +670,61 @@ static inline void s_get_next_element(struct aws_hash_iter *iter, size_t start_s
...
     iter.map = map;
     iter.limit = state->size;
     s_get_next_element(&iter, 0);
+    AWS_POSTCONDITION(iter.status == AWS_HASH_ITER_STATUS_DONE || iter.status == AWS_HASH_ITER_STATUS_READY_FOR_USE);
+    AWS_POSTCONDITION(aws_hash_iter_is_valid(&iter));
     return iter;
 }
 
 bool aws_hash_iter_done(const struct aws_hash_iter *iter) {
+    AWS_PRECONDITION(iter);
```
feliperodri: This should be checked in `aws_hash_iter_is_valid(iter)`

.cbmc-batch/jobs/aws_hash_iter_delete/Makefile
danielsn: Fixed.

.cbmc-batch/jobs/aws_hash_iter_delete/aws_hash_iter_delete_harness.c
danielsn: Moved.

.cbmc-batch/jobs/aws_hash_iter_delete/aws_hash_iter_delete_harness.c
danielsn: Moved.

source/hash_table.c
danielsn: See the comment in the code right above.

.cbmc-batch/jobs/aws_hash_iter_done/Makefile
danielsn: Added comment to the makefile explaining why.

.cbmc-batch/jobs/aws_hash_iter_done/Makefile
danielsn: Done

source/hash_table.c
danielsn: Done

source/hash_table.c
danielsn: Done

.cbmc-batch/include/proof_helpers/make_common_data_structures.h
```diff
@@ -133,3 +133,19 @@ void ensure_allocated_hash_table(struct aws_hash_table *map, size_t max_table_en
...
  */
 const char *make_arbitrary_c_str(size_t max_size);
+
+/**
+ * A correct hash table has max_load < size.  This means that there is always one slot empty.
+ * This function is useful for assuming that there is some (nondet) slot which is empty
+ * which is necessary to prove termination for hash-table deletion code.  Should only be used inside
+ * an assume because of the way it does nondet.
+ */
+bool aws_hash_table_has_an_empty_slot(struct aws_hash_table *map, size_t *rval);
```
karkhaz: * Suggest `struct aws_hash_table const * const map, size_t * const rval`
* Naiive question: are you checking elsewhere that `max_load < size`? If that condition doesn't hold, what actually happens–does the proof break, or might it silently let bugs through?

include/aws/common/hash_table.h
```diff
@@ -74,16 +75,24 @@ struct aws_hash_element {
...
+    AWS_HASH_ITER_STATUS_DELETE_CALLED,
+    AWS_HASH_ITER_STATUS_READY_FOR_USE,
+};
+
 struct aws_hash_iter {
     const struct aws_hash_table *map;
     struct aws_hash_element element;
     size_t slot;
     size_t limit;
+    /* Holds an enum aws_hash_iter_status */
```
karkhaz: Is this comment necessary?

include/aws/common/private/hash_table_impl.h
```diff
@@ -90,6 +90,41 @@ bool aws_hash_table_is_valid(const struct aws_hash_table *map) {
...
+            /* Done iff slot == limit */
+            return iter->slot == iter->limit;
+        case AWS_HASH_ITER_STATUS_DELETE_CALLED:
+            /* iter->slot can underflow to SIZE_MAX after a delete
+             * see the comments for aws_hash_iter_delete() */
+            return iter->slot <= iter->limit || iter->slot == SIZE_MAX;
+        case AWS_HASH_ITER_STATUS_READY_FOR_USE:
+            /* A slot must point to a valid location (i.e. hash_code != 0) */
+            return iter->slot < iter->limit && iter->map->p_impl->slots[iter->slot].hash_code != 0;
+        default:
```
karkhaz: I wonder whether it makes sense to omit the default case, so that the compiler can immediately warn us when somebody adds a new status? We can still return false, but after the end of the switch statement...

source/hash_table.c
```diff
@@ -542,13 +550,14 @@ static size_t s_remove_entry(struct hash_table_state *state, struct hash_table_e
         }
 
         /* Okay, shift this one back */
-        memcpy(&state->slots[index], &state->slots[next_index], sizeof(*state->slots));
+        state->slots[index] = state->slots[next_index];
```
karkhaz: It's not obvious to me why you made this change. Does this fix a bug, or is it a semantics-preserving change that makes the proof possible or faster? Please add an explanation to the commit message

source/hash_table.c
```diff
@@ -653,38 +671,60 @@ static inline void s_get_next_element(struct aws_hash_iter *iter, size_t start_s
         if (entry->hash_code) {
             iter->element = entry->element;
             iter->slot = i;
+            iter->status = AWS_HASH_ITER_STATUS_READY_FOR_USE;
```
karkhaz: Is this a bug fix? Please mention in the commit message if so

source/hash_table.c
```diff
@@ -653,38 +671,60 @@ static inline void s_get_next_element(struct aws_hash_iter *iter, size_t start_s
...
             return;
         }
     }
     iter->element.key = NULL;
     iter->element.value = NULL;
     iter->slot = iter->limit;
+    iter->status = AWS_HASH_ITER_STATUS_DONE;
```
karkhaz: Is this a bug fix? Please mention in the commit message if so

source/hash_table.c
```diff
@@ -722,7 +762,13 @@ void aws_hash_iter_delete(struct aws_hash_iter *iter, bool destroy_contents) {
      * underflowing to SIZE_MAX; we have to take care in aws_hash_iter_done to avoid
      * treating this as an end-of-iteration condition.
      */
+#pragma CPROVER check push
+#pragma CPROVER check disable "unsigned-overflow"
     iter->slot--;
+#pragma CPROVER check pop
+    iter->status = AWS_HASH_ITER_STATUS_DELETE_CALLED;
```
karkhaz: Is this a bug fix? Please mention in the commit message if so

.cbmc-batch/include/proof_helpers/make_common_data_structures.h
danielsn: 1. done
1. That is checked as a pre-condition to the function, which translates to an ```assert``` at runtime.  I'm also checking that every function which acts on hash-tables does that as a post-condition, so once I finish those proofs we will have a functional proof of this property.

include/aws/common/hash_table.h
danielsn: Nope. That's from an old version of the code. Removed.

include/aws/common/private/hash_table_impl.h
danielsn: Done.

source/hash_table.c
danielsn: This is semantics preserving.  In my mind, it has two advantages: 
1. A simple assignment is more obviously correct than the memcpy (where an ```&``` or a ```*``` could easily be added or missed)
2. In some cases, I want the ability to override ```memcpy``` as part of CBMC proofs. The fewer times ```memcpy``` appears, the less I/CBMC have to reason about what it is doing.

source/hash_table.c
danielsn: This is more of an improvement than a bug fix.  There is an implicit state machine for using an ```aws_hash_iter```: for example, it is an error to call ```iter_delete``` on an iter which is in the ```done``` state, and after ```iter_delete```  has been called, the only valid thing to do is to call ```iter_next```.  By making the state machine explicit, it is possible to both test it during debug builds, and to prove properties about consumers of the hash-table.  In particular, this sets up aws-encryption-sdk proofs to check this property.

About the commit message: all commits will be squashed when the PR is merged.  I will write a better commit message for the squash commit.

source/hash_table.c
danielsn: See above.

source/hash_table.c
danielsn: see above.

tautschnig:
@danielsn CI has now succeeded.

PR #342, opened 2019-05-09
Add aws_byte_buf_reset() function

danielsn:
*Description of changes:*
There are cases where the user wants to reuse the same buffer (we see this in some c-common tests, for example). Instead of either having to
1. poke inside the buf internals
1. call secure_zero, which has the added cost of a memset.

this gives the user an API function that does the right thing.

### DIFF
--- a/.cbmc-batch/jobs/aws_byte_buf_reset/aws_byte_buf_reset_harness.c
+++ b/.cbmc-batch/jobs/aws_byte_buf_reset/aws_byte_buf_reset_harness.c
@@ -20,10 +20,9 @@
 void aws_byte_buf_reset_harness() {
     struct aws_byte_buf buf;
 
-    __CPROVER_assume(is_bounded_byte_buf(&buf, MAX_BUFFER_SIZE));
+    __CPROVER_assume(aws_byte_buf_is_bounded(&buf, MAX_BUFFER_SIZE));
     ensure_byte_buf_has_allocated_buffer_member(&buf);
     __CPROVER_assume(aws_byte_buf_is_valid(&buf));
-    assert(buf.capacity <= MAX_BUFFER_SIZE);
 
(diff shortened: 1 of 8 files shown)


.cbmc-batch/jobs/aws_byte_buf_reset/aws_byte_buf_reset_harness.c
```diff
@@ -0,0 +1,38 @@
...
+#include <proof_helpers/make_common_data_structures.h>
+#include <proof_helpers/utils.h>
+
+void aws_byte_buf_reset_harness() {
+    struct aws_byte_buf buf;
+
+    __CPROVER_assume(is_bounded_byte_buf(&buf, MAX_BUFFER_SIZE));
+    ensure_byte_buf_has_allocated_buffer_member(&buf);
+    __CPROVER_assume(aws_byte_buf_is_valid(&buf));
+    assert(buf.capacity <= MAX_BUFFER_SIZE);
```
feliperodri: This is not necessary

.cbmc-batch/jobs/Makefile.aws_byte_buf
```diff
@@ -0,0 +1,21 @@
+
```
feliperodri: Remove extra line

.cbmc-batch/jobs/aws_byte_buf_reset/aws_byte_buf_reset_harness.c
danielsn: Removed.

.cbmc-batch/jobs/Makefile.aws_byte_buf
danielsn: Done

PR #346, opened 2019-05-13
Updates implementation of aws_byte_buf_from_c_str function

feliperodri:

*Description of changes:*

1. Updates implementation of `aws_byte_buf_from_c_str` function;
2. Adds a proof harness for `aws_byte_buf_from_c_str` function;
3. Updates the `test_buffer_init_copy` test.

### DIFF
--- a/.cbmc-batch/jobs/aws_byte_buf_from_c_str/aws_byte_buf_from_c_str_harness.c
+++ b/.cbmc-batch/jobs/aws_byte_buf_from_c_str/aws_byte_buf_from_c_str_harness.c
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/byte_buf.h>
+#include <proof_helpers/make_common_data_structures.h>
+
+void aws_byte_buf_from_c_str_harness() {
+    size_t len;
+    char *c_str;
+
+    if (nondet_bool()) {
+        c_str = NULL;
+    } else {
+        /* Assume the length of the string is bounded by some max length */
+        __CPROVER_assume(len <= MAX_BUFFER_SIZE);
+        ASSUME_VALID_MEMORY_COUNT(c_str, len);
+
+        /* Need *c_str to be a '\0'-terminated C string, so assume an arbitrary character is 0 */
+        int index;
+        __CPROVER_assume(index >= 0 && index < len);
+        c_str[index] = 0;
+    }
+
+    /* operation under verification */
+    struct aws_byte_buf buf = aws_byte_buf_from_c_str(c_str);
+
+    /* assertions */
+    assert(aws_byte_buf_is_valid(&buf));
+    assert(buf.allocator == NULL);
+    if (buf.buffer) {
+        assert(buf.len == strlen(c_str));
+        assert(buf.capacity == buf.len + 1);
+        assert_bytes_match(buf.buffer, (uint8_t *)c_str, buf.len);
+    } else {
+        assert(buf.len == 0);
+        assert(buf.capacity == 0);
+    }
+}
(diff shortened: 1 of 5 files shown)


feliperodri:
@karkhaz @danielsn any comments?

feliperodri:
@baha @dlohar any comments?

include/aws/common/byte_buf.h
```diff
@@ -451,9 +451,16 @@ AWS_EXTERN_C_END
...
     buf.buffer = (uint8_t *)c_str;
-    buf.len = strlen(c_str);
-    buf.capacity = buf.len;
+    if (!buf.buffer) {
+        buf.len = 0;
+        buf.capacity = 0;
+    } else {
+        buf.len = strlen(c_str);
+        /* considering the byte for the terminating null character */
+        buf.capacity = buf.len + 1;
```
ColdenCullen: I worry about this, because having the capacity be higher than len implies that there's a byte at the end that's writable, but that's probably not safe, especially in cases where the cstr is a literal in the .data section. Can we instead not add 1 and just pretend the null doesn't exist?

feliperodri: In that case, if `c_str` is an empty string, i.e., `strlen(c_str) == 0`, then the correct behavior would be to set `buf.buffer == NULL`? Or should I update `aws_byte_buf_is_valid` to consider valid either `buffer == NULL` or `strlen(buffer) == 0` when `len == 0`  and `capacity == 0`.

ColdenCullen: I think that behavior here is fine, I'm just suggesting that you remove the `+ 1` from the capacity.

.cbmc-batch/jobs/aws_byte_buf_from_c_str/aws_byte_buf_from_c_str_harness.c
```diff
@@ -0,0 +1,50 @@
...
+
+void aws_byte_buf_from_c_str_harness() {
+    size_t len;
+    char *c_str;
+
+    if (nondet_bool()) {
+        c_str = NULL;
+    } else {
+        /* Assume the length of the string is bounded by some max length */
+        __CPROVER_assume(len <= MAX_BUFFER_SIZE);
```
danielsn: There is a function for this https://github.com/awslabs/aws-c-common/blob/master/.cbmc-batch/include/proof_helpers/make_common_data_structures.h#L135

.cbmc-batch/jobs/aws_byte_buf_from_c_str/aws_byte_buf_from_c_str_harness.c
```diff
@@ -0,0 +1,50 @@
...
+
+    /* assertions */
+    assert(aws_byte_buf_is_valid(&buf));
+    assert(buf.allocator == NULL);
+    if (buf.buffer) {
+        assert(buf.len == strlen(c_str));
+        assert(buf.capacity == buf.len);
+        assert_bytes_match(buf.buffer, (uint8_t *)c_str, buf.len);
+    } else {
+        assert(buf.len == 0);
```
danielsn: ```assert(strlen(c_str) == 0)```

.cbmc-batch/jobs/aws_byte_buf_from_c_str/aws_byte_buf_from_c_str_harness.c
```diff
@@ -0,0 +1,50 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/byte_buf.h>
+#include <proof_helpers/make_common_data_structures.h>
+
+void aws_byte_buf_from_c_str_harness() {
+    size_t len;
+    char *c_str;
```
danielsn: Once you use the function below, you can make this into a one liner.

feliperodri:
@ColdenCullen ready for review

include/aws/common/byte_buf.h
```diff
@@ -450,10 +450,11 @@ AWS_EXTERN_C_END
  */
 AWS_STATIC_IMPL struct aws_byte_buf aws_byte_buf_from_c_str(const char *c_str) {
     struct aws_byte_buf buf;
-    buf.buffer = (uint8_t *)c_str;
-    buf.len = strlen(c_str);
+    buf.len = (!c_str) ? 0 : strlen(c_str);
     buf.capacity = buf.len;
+    buf.buffer = (buf.capacity == 0) ? NULL : (uint8_t *)c_str;
```
ColdenCullen: `buf.buffer = (uint8_t *)c_str` should be fine, `NULL` casted to `uint8_t *` is still `NULL` 😛

ColdenCullen: Oh I see you're branching on size, not pointer. I don't love this, but I guess it is at least consistent with from_array.

PR #349, opened 2019-05-13
Adds proof harness for aws_nospec_mask function

feliperodri:

*Description of changes:*
1. Updates implementation of `aws_nospec_mask` function with pre- and post-conditions;
2. Adds a proof harness for `aws_nospec_mask` function;

### DIFF
--- a/.cbmc-batch/jobs/aws_nospec_mask/cbmc-batch.yaml
+++ b/.cbmc-batch/jobs/aws_nospec_mask/cbmc-batch.yaml
@@ -1,0 +1,4 @@
+jobos: ubuntu16
+cbmcflags: "--bounds-check;--div-by-zero-check;--float-overflow-check;--nan-check;--pointer-check;--pointer-overflow-check;--signed-overflow-check;--undefined-shift-check;--unsigned-overflow-check;--unwinding-assertions;--unwind;1"
+goto: aws_nospec_mask_harness.goto
+expected: "SUCCESSFUL"
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -501,6 +501,8 @@ AWS_STATIC_IMPL struct aws_byte_cursor aws_byte_cursor_from_array(const void *by
     return cur;
 }
 
+#pragma CPROVER check push
+#pragma CPROVER check disable "unsigned-overflow"
 /**
  * If index >= bound, bound > (SIZE_MAX / 2), or index > (SIZE_MAX / 2), returns
  * 0. Otherwise, returns UINTPTR_MAX.  This function is designed to return the correct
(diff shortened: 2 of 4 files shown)


feliperodri:
@karkhaz @danielsn @Baha @dlohar @mmuesly any comments?

include/aws/common/byte_buf.h
```diff
@@ -507,6 +507,8 @@ AWS_STATIC_IMPL struct aws_byte_cursor aws_byte_cursor_from_array(const void *by
  * SPECTRE mitigation purposes.
  */
 AWS_STATIC_IMPL size_t aws_nospec_mask(size_t index, size_t bound) {
```
feliperodri: Can I ignore any unsigned overflow in this function? @danielsn @justinboswell @JonathanHenson @bretambrose

bdonlan: Unsigned over/underflow is deliberately used as part of this function's logic. You probably should not be warning on the _presence_ of over/underflow, but it is incorrect to _assume_ that over/underflow won't occur.

.cbmc-batch/jobs/aws_nospec_mask/cbmc-batch.yaml
```diff
@@ -0,0 +1,4 @@
+jobos: ubuntu16
+cbmcflags: "--bounds-check;--div-by-zero-check;--float-overflow-check;--nan-check;--pointer-check;--pointer-overflow-check;--signed-overflow-check;--undefined-shift-check;--unsigned-overflow-check;--unwinding-assertions;--unwind;1"
```
bdonlan: Should we just remove `--unsigned-overflow-check` here?

feliperodri: This would disable the unsigned overflow check for all code covered by this proof. However, the right thing would be to disable the unsigned overflow check only in the portion of the code that deliberately uses these overflows. To that end, we have the following pragmas

```
#pragma CPROVER check push
#pragma CPROVER check disable "unsigned-overflow"

// Code with known unsigned overflows

#pragma CPROVER check pop
```

include/aws/common/byte_buf.h
```diff
@@ -507,6 +507,8 @@ AWS_STATIC_IMPL struct aws_byte_cursor aws_byte_cursor_from_array(const void *by
...
+#pragma CPROVER check push
+#pragma CPROVER check disable "unsigned-overflow"
```
feliperodri: @bdonlan This pragma disables the unsigned overflow check for this particular code.

include/aws/common/byte_buf.h
```diff
@@ -562,6 +564,7 @@ AWS_STATIC_IMPL size_t aws_nospec_mask(size_t index, size_t bound) {
      * an arithmetic right shift operation on x86.
      */
     combined_mask = combined_mask * UINTPTR_MAX;
+#pragma CPROVER check pop
```
danielsn: I'd find it cleaner to put the pop on the last line of the function, after the return.  Alternatively, put the pragma around the while function, including the signature.

include/aws/common/byte_buf.h
feliperodri: We now consider all unsigned overflow in this function to be benign, since they are used as part of the function's logic.

PR #359, opened 2019-05-16
Adds assert_byte_buf_equivalence function to proof helpers

feliperodri:

*Description of changes:*
Adds `assert_byte_buf_equivalence` function.


feliperodri:
@karkhaz @danielsn any comments?

.cbmc-batch/source/utils.c
```diff
@@ -64,6 +64,18 @@ void assert_array_list_equivalence(
     }
 }
 
+void assert_byte_buf_equivalence(
+    const struct aws_byte_buf *const lhs,
+    const struct aws_byte_buf *const rhs,
+    const struct store_byte_from_buffer *const rhs_byte) {
+    assert(lhs->len == rhs->len);
```
karkhaz: Don't you need to check that `lhs` and `rhs` are not null before you dereference `len`?

feliperodri: Yes, thank you!

.cbmc-batch/include/proof_helpers/utils.h
```diff
@@ -64,6 +65,16 @@ void assert_array_list_equivalence(
     const struct aws_array_list *const rhs,
     const struct store_byte_from_buffer *const rhs_byte);
 
+/**
```
danielsn: See comment on previous review.

.cbmc-batch/source/utils.c
```diff
@@ -38,7 +38,7 @@ void assert_all_zeroes(const uint8_t *const a, const size_t len) {
 }
 
 void assert_byte_from_buffer_matches(const uint8_t *const buffer, const struct store_byte_from_buffer *const b) {
-    if (buffer) {
+    if (buffer && b) {
```
danielsn: when would we ever not have a b?

.cbmc-batch/source/utils.c
```diff
@@ -55,12 +55,30 @@ void assert_array_list_equivalence(
     const struct aws_array_list *const lhs,
     const struct aws_array_list *const rhs,
     const struct store_byte_from_buffer *const rhs_byte) {
-    assert(lhs->alloc == rhs->alloc);
-    assert(lhs->current_size == rhs->current_size);
-    assert(lhs->length == rhs->length);
-    assert(lhs->item_size == rhs->item_size);
-    if (lhs->current_size > 0) {
-        assert_byte_from_buffer_matches((uint8_t *)lhs->data, rhs_byte);
+    assert(!lhs == !rhs);
```
danielsn: Clever.  But a comment might help

.cbmc-batch/source/utils.c
```diff
@@ -55,12 +55,30 @@ void assert_array_list_equivalence(
...
+            assert_byte_from_buffer_matches((uint8_t *)lhs->data, rhs_byte);
+        }
+    }
+}
+
+void assert_byte_buf_equivalence(
+    const struct aws_byte_buf *const lhs,
+    const struct aws_byte_buf *const rhs,
+    const struct store_byte_from_buffer *const rhs_byte) {
+    assert(!lhs == !rhs);
```
danielsn: Comment here.

.cbmc-batch/source/utils.c
```diff
@@ -55,12 +55,30 @@ void assert_array_list_equivalence(
...
+    if (lhs && rhs) {
```
danielsn: If we have ```lhs``` but not ```rhs```, what are the sematics.

.cbmc-batch/source/utils.c
feliperodri: When one tries to save a byte from a NULL buffer using `save_byte_from_array`

.cbmc-batch/source/utils.c
feliperodri: What do you mean? The assertion on the previous line ensures either both are null or both are non-null. Could you clarify the comment?

danielsn:
```
if (lhs == rhs) return;
assert (lhs && rhs) // if only one is null, they're different
//they're both non-null, now, check the internals.
```

PR #366, opened 2019-05-21
Calloc is now fixed in CBMC, so use it. 

danielsn:
*Description of changes:*
* Use calloc from CBMC
* Fix up the aws_hash_table_init proof: now runs in seconds instead of minutes.

### DIFF
--- a/.cbmc-batch/jobs/aws_hash_table_init-bounded/Makefile
+++ b/.cbmc-batch/jobs/aws_hash_table_init-bounded/Makefile
@@ -12,25 +12,21 @@
 # limitations under the License.
 
 ###########
-# Sufficient for coverage for bounded inputs
-# Additional coverage is given using the -unbounded version of this proof
-# 4 - 1 min
-# 8 - 2 min
-# 12 - 6 min
-MAX_TABLE_SIZE ?= 8
+# Using the new calloc model, get full coverage in about 5s on unbounded sizes!
+MAX_TABLE_SIZE ?= SIZE_MAX
 DEFINES += -DMAX_TABLE_SIZE=$(MAX_TABLE_SIZE) 
-
-#The hash table struct itself is 80 bytes, which is 10 iters 
-#A hash entry is 24 bytes, which is 10 iters. So we need 3 * MAX_TABLE_SIZE + 10 + 1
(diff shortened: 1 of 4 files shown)


.cbmc-batch/jobs/aws_hash_table_init-bounded/Makefile
```diff
@@ -12,26 +12,22 @@
...
 DEPENDENCIES += $(HELPERDIR)/source/utils.c
 DEPENDENCIES += $(HELPERDIR)/stubs/error.c
-DEPENDENCIES += $(HELPERDIR)/stubs/memset_override_0.c
 DEPENDENCIES += $(SRCDIR)/source/common.c
 DEPENDENCIES += $(SRCDIR)/source/hash_table.c
 
+ADDITIONAL_REMOVE_FUNCTION_BODY += --remove-function-body bounded_malloc
+ADDITIONAL_REMOVE_FUNCTION_BODY += --remove-function-body can_fail_malloc
+ADDITIONAL_REMOVE_FUNCTION_BODY += --remove-function-body nondet_compare
+ADDITIONAL_REMOVE_FUNCTION_BODY += --remove-function-body s_can_fail_malloc_allocator
```
mmuesly: Why do you need to drop these function bodies?

danielsn: They're not used in the proof, and having them around was causing the function pointer analysis to consider, even tho they weren't used (a Mark Tuttle suggestion, to give credit where credit is due)

.cbmc-batch/jobs/aws_hash_table_init-bounded/Makefile
```diff
@@ -12,26 +12,22 @@
...
 
 ###########
-# Sufficient for coverage for bounded inputs
-# Additional coverage is given using the -unbounded version of this proof
-# 4 - 1 min
-# 8 - 2 min
-# 12 - 6 min
-MAX_TABLE_SIZE ?= 8
+# Using the new calloc model, get full coverage in about 5s on unbounded sizes!
+MAX_TABLE_SIZE ?= SIZE_MAX
```
feliperodri: Why do you need to explicitly assign `SIZE_MAX` instead of just considering non-deterministic values?

PR #370, opened 2019-05-23
Update travis to use builder.py

ColdenCullen:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,4 @@ install:
 
 script:
   - export CODEBUILD_SRC_DIR=$(pwd)
-  - ./codebuild/common-posix.sh -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+  - python3 ./codebuild/builder.py build macos-clang-default-macos-x64
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,12 @@ osx_image:
   - xcode10.1
 compiler: clang
 
+addons:
+  homebrew:
+    update: true
+    packages:
+      - python@3
+
 install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,9 +8,15 @@ osx_image:
   - xcode10.1
 compiler: clang
 
+addons:
+  homebrew:
+    update: true
+    packages:
```
justinboswell: so do we no longer install cmake? Is there a system one to use?

codebuild/builder.py
```diff
@@ -143,6 +148,16 @@ def __repr__(self):
             "-DENABLE_SANITIZERS=ON",
         ],
     },
+    'macos': {
+        'architectures': {
+            'x86': {
```
justinboswell: Wait, we support 32-bit OSX? Is that even a thing at this point?

.travis.yml
ColdenCullen: I mean, I guess? I didn't remove anything, and the builds passed, so praise the travis gods?

codebuild/builder.py
ColdenCullen: Hell if I know, but couldn't see a reason not to put it in there

.travis.yml
justinboswell: :muscle:

codebuild/builder.py
justinboswell: won't this double our travis builds?

ColdenCullen: No, because travis builds aren't defined by this script, they're defined by travis.yml.

Issue #123, opened 2018-08-15
aws_base64_decode rejects the output of aws_base64_encode


bdonlan:
`aws_base64_encode` appends a NUL-terminator to its output, which is included in the byte_buf's len count. However, `aws_base64_decode` calls out to `aws_base64_compute_decoded_len`, which assumes the length counts only the actual base64 encoded data, and rejects if this data is not a length that is a multiple of 4. With the NUL terminator, this check never passes, and so passing the output buffer from `aws_base64_encode` directly to `aws_base64_decode` always fails.

ColdenCullen:
Why are we writing a terminator at all? I can't find any docs online that suggest doing that.

bdonlan:
It's useful if you want to use the buffer in a C routine (fprintf, etc), but probably should not be counted against len.

ColdenCullen:
Could you do something like we did in awslabs/aws-c-io#29 and append a null terminator, but not include it in the length?

jonahharris:
Agreed. After looking through it, as aws_base64_compute_encoded_len is public and used (by the developer) to allocate the byte buffer, thinking we should let that still account for the null terminator, but update aws_base64_encode to set the byte buffer's len property such that the terminator is unaccounted for. All usages I can see in the test cases account for it by decrementing len prior to passing to aws_base64_decode, as referenced by op. Those are easy to update.

It appears this is also used in aws-crt-cpp and aws-c-event-stream libraries. In aws-crt-cpp, it's called in Base64Encode, there are no cases I can find of its use in tests or otherwise. In aws-c-event-stream, it's passed to fprintf and will work as-expected with the change.

A cursory search of base-64 encoding/decoding implementations seems to be a 60/40 split as far as not including vs including the null terminator in the length. So, I think we'd be fine not including it, which is also my preference. I did, however, notice [another AWS library implementation of b64](https://github.com/awslabs/amazon-kinesis-video-streams-producer-sdk-cpp/blob/master/kinesis-video-pic/src/utils/src/Base64.cpp#L52) that _does_ include the null terminator in the length.

Thoughts? I can code this fix up quickly if a decision is made. Likewise, should the same change be made for hex encoding?

JonathanHenson:
@bdonlan agreed, null terminators should NEVER be counted against length. This one is my error, because the RFC unit test cases had empty string as a vector. We should leave the NULL terminator but not count it against length.

ColdenCullen:
Hey this seems fixed?

PR #376, opened 2019-05-24
Marking unused parameters with AWS_UNUSED_PARAM macro.

jonahharris:
*Description of changes:*

This PR adds an AWS_UNUSED_PARAM(param) macro and respectively replaces all usages of the (void)param pattern. This makes identification of unused parameters simpler.

For context, I've been migrating a large database proxy system and compiler (of interest to AWS) from libevent/libuv over to AWS-C-IO. The code uses a number of portability/convenience macros, similar to those found in [Hedley](https://nemequ.github.io/hedley/api-reference.html), and I'm looking to add the commonly-useful ones to AWS-C-COMMON. Accordingly, I'll update the other common-based libraries (io, http, mqtt, etc.) to use any new addition as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,7 @@ and let the user figure it out.
 * There is only one valid character encoding-- UTF-8. Try not to ever need to care about character encodings, but
 where you do, the working assumption should always be UTF-8 unless it's something we don't get a choice in (e.g. a protocol
 explicitly mandates a character set).
+* If you are not using one or more parameters of a function within its body, use the `AWS_UNUSED_PARAM` macro to communicate explicit intent and avoid possible compilation warnings.
 * If you are adding/using a compiler specific keyword, macro, or intrinsic, hide it behind a platform independent macro
 definition. This mainly applies to header files. Obviously, if you are writing a file that will only be built on a certain
 platform, you have more liberty on this.
--- a/include/aws/common/atomics_msvc.inl
+++ b/include/aws/common/atomics_msvc.inl
@@ -89,7 +89,7 @@ static inline void aws_atomic_priv_check_order(enum aws_memory_order order) {
             abort();
     }
 #endif
-    (void)order;
+    AWS_UNUSED_PARAM(order);
 }
 
 enum aws_atomic_mode_priv { aws_atomic_priv_load, aws_atomic_priv_store };
--- a/include/aws/common/common.h
+++ b/include/aws/common/common.h
@@ -51,7 +51,7 @@
             memset((void *)array, 0, sizeof(array));                                                                   \
         } while (0)
 #define AWS_ARRAY_SIZE(array) (sizeof(array) / sizeof(array[0]))
-
+#define AWS_UNUSED_PARAM(param) (void)(param)
 
 #define AWS_CACHE_LINE 64
 
--- a/include/aws/common/common.h
+++ b/include/aws/common/common.h
@@ -51,7 +51,11 @@
             memset((void *)array, 0, sizeof(array));                                                                   \
         } while (0)
 #define AWS_ARRAY_SIZE(array) (sizeof(array) / sizeof(array[0]))
-#define AWS_UNUSED_PARAM(param) (void)(param)
+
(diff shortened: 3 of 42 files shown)


jonahharris:
@ColdenCullen - I'm interested in your opinions/suggestions on how I should proceed with not only this PR, but also any potential additions to this (and related) libraries. Prior to any larger changes, other than bug-fixes or test additions, should I want to make a change, would you prefer I put it in an issue for discussion pre-development? I have no desire to bloat your libraries, I just want to add things I think are generally useful to other developers and that will be used/needed within my codebase.

ColdenCullen:
Not sure why my opinion is worth bonus points, but my feelings are:
Pros:
* The macro is slightly clearer

Cons:
* Canceling out a parameter now requires significantly more text
* This change involves LOTS of churn

When in doubt, I tend to lean towards not making changes, but if anyone else on @awslabs/aws-sdk-common-runtime-team has feelings (looking at you @JonathanHenson), I could be swayed.

jonahharris:
> Not sure why my opinion is worth bonus points

I went off the contributor graph and the cursory observation that I've seen the latest responses from you in terms of communication regarding issues and such :)

JonathanHenson:
I don’t really care one way or the other as long as it is consistent. We did something similar in the c++ sdk years ago. One benefit of the macro is that it’s explicit about intent, and reduces cognitive load. If we do move forward with this, let’s update the coding guidelines as well. Also keep in mind, these header files will be included in the C++ SDK at some point and this macro will likely clash with its macro of the same name. Maybe put an ifdef check in there as well.

JonathanHenson:
This looks fine to me. I’ll talk with the team today and make sure everyone wants to move forward with the change.

jonahharris:
Cool. Thanks!

JonathanHenson:
talked to team about it this morning. We can't find a compelling reason why this is better than it currently is. And while the change is fine, since we're neutral on it, we prefer to provide Occam's razor on changes.

Can you provide us with an argument for beyond the subjective argument I made about "explicit intent?"

jonahharris:
The code I'm working on will soon be handed-over to AWS. To that end, I've already been making a considerable number of changes to switch from third-party libraries over to AWS-specific ones along with performing general clean-ups. Trust me, I get it: my position is similar to your team's in terms of, where possible, making the fewest changes required. In this case, for those things in my codebase that already seem cleaner, I'd just prefer not to make them less-so.

While I can always keep custom definitions within the codebase, it seemed this would be a good candidate for common, which I assume is why it was also added to the aforementioned C++ library.

While I get the default stance, aside from churn, I don't see a negative. Having made the related update to all but one other (public) library based on common, I've ensured 100% passing test coverage.

Regardless of this PR, I guess my question going forward is: how do we avoid this in the future? I'd like to hand over code that is relatively-clean and I'm happy to contribute to the base (-common and -io libraries on which this will be based), but I don't want to spend a couple hours working on something that isn't likely to be accepted. Any suggestions you or your team has would be appreciated.

JonathanHenson:
We'd obviously always love your contributions. They're very much welcome. We also have a long roadmap of things that need improvement/implementation if you're looking for something in your spare time.

Regarding your last paragraph, I'd suggest opening an issue for us to dialog on to avoid churn for a large change. 

I'm still not against this, we're just mulling it over. I agree there's not really a negative here, we just also don't see the positive.

jonahharris:
Cool. I'll open an issue on any future proposed change. This was a simple one, but I was definitely wondering how you would like to handle anything not-straight-forward. As an example: https://github.com/awslabs/aws-c-common/issues/123#issuecomment-495848822

jonahharris:
While going through the IOT library samples I noticed it also has a similar macro, IOT_UNUSED.

jonahharris:
I now have three full database protocol implementations rewritten with aws-c-common and aws-c-io; is this still iffy in terms of being merged? If the team if pro-merge, I'll run through and make sure it's up-to-date with master. If not, I'll just update my local definitions to use something other than AWS_UNUSED.

PR #394, opened 2019-05-30
Implement aws_ntoh24 and aws_hton24

ColdenCullen:
HTTP/2 stores frame length as a 24 bit integer, so here we are.

### DIFF
--- a/include/aws/common/byte_order.h
+++ b/include/aws/common/byte_order.h
@@ -85,6 +85,44 @@ AWS_STATIC_IMPL uint32_t aws_ntoh32(uint32_t x) {
 }
 
 /**
+ * Convert 24 bit integer from host to network byte order.
+ */
+AWS_STATIC_IMPL uint32_t aws_hton24(uint32_t x) {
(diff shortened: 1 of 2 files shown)


include/aws/common/byte_order.h
```diff
@@ -84,6 +84,44 @@ AWS_STATIC_IMPL uint32_t aws_ntoh32(uint32_t x) {
...
+        return x;
+    } else {
+        uint32_t result = 0;
+        uint8_t *result_ptr = (uint8_t *)&result;
+
+        *result_ptr++ = (x >> 16) & 0xFF;
+        *result_ptr++ = (x >> 8) & 0xFF;
+        *result_ptr++ = (x >> 0) & 0xFF;
+
+        return result;
```
marcomagdy: is this the same as aws_ntoh32 followed by an 8 bit right shift?

ColdenCullen: No you're right, it's exactly the same! Re-implementing...

PR #377, opened 2019-05-24
Implement a harness for s_swap in priority_queue.c

angelhof:
Implement a proof harness for `s_swap` in `priority_queue.c`.

In the process I extended `priority_queue_is_valid` to check
whether the backpointers array_list is correctly allocated to be of equal
size to the container array_list.

Note that the `Makefile.common` is taken from #373 because `s_swap` is a static function.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #369, opened 2019-05-22
Redefine AWS_PRECONDITION and add an explanation to all preconditions

angelhof:
Redefine `AWS_PRECONDITION` using `__CPROVER_precondition` and add an explanation string to all preconditions in the library.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #397, opened 2019-05-31
Harnesses for prio queue push and push_ref

angelhof:
Implement harnesses for `aws_priority_queue_push` and `aws_priority_queue_push_ref`.

Depends on #377 


.cbmc-batch/stubs/aws_array_list_ensure_capacity_override.c
```diff
@@ -0,0 +1,22 @@
+#include <aws/common/array_list.h>
```
tautschnig: 1) Missing license header
2) Could you comment on what exactly is different in this implementation over the original one, and maybe also why we need the stub?

.cbmc-batch/stubs/s_sift_up_override.c
```diff
@@ -0,0 +1,10 @@
+
```
tautschnig: License header

source/priority_queue.c
```diff
@@ -145,11 +145,7 @@ int aws_priority_queue_init_dynamic(
     queue->pred = pred;
     AWS_ZERO_STRUCT(queue->backpointers);
 
-    
-    bool ret = aws_array_list_init_dynamic(&queue->container,
-                                           alloc,
-                                           default_size,
-                                           item_size);
+    bool ret = aws_array_list_init_dynamic(&queue->container, alloc, default_size, item_size);
```
tautschnig: Formatting change?

angelhof: I think that clang-format did that change.

.cbmc-batch/stubs/aws_array_list_ensure_capacity_override.c
angelhof: The stub was used to improve verification time, but as it seems, verification time is reasonable with the standard implementation too, so I will remove it.

angelhof:
@ColdenCullen @JonathanHenson @justinboswell @graebm could you review and merge this PR?

angelhof:
The errors seem to be unrelated with the changes that this PR makes.

source/priority_queue.c
```diff
@@ -176,8 +176,6 @@ void aws_priority_queue_init_static(
     AWS_ZERO_STRUCT(queue->backpointers);
 
     aws_array_list_init_static(&queue->container, heap, item_count, item_size);
-
-    AWS_POSTCONDITION(aws_priority_queue_is_valid(queue));
```
ColdenCullen: Why remove this?

angelhof: i didn't mean to remove this, thanks for the catch.

.cbmc-batch/jobs/aws_priority_queue_push/aws_priority_queue_push_harness.c
```diff
@@ -0,0 +1,37 @@
...
+    struct aws_priority_queue queue;
+
+    /* Assumptions */
+    __CPROVER_assume(aws_priority_queue_is_bounded(&queue, MAX_INITIAL_ITEM_ALLOCATION, MAX_ITEM_SIZE));
+    ensure_priority_queue_has_allocated_members(&queue);
+    __CPROVER_assume(aws_priority_queue_is_valid(&queue));
+    void *item = can_fail_malloc(queue.container.item_size);
+
+    /* Assume the function preconditions */
+    __CPROVER_assume(item && AWS_MEM_IS_READABLE(item, queue.container.item_size));
```
karkhaz: I'm not asking for changes here. But out of curiosity: is it documented somewhere that (for example) `item` should be non-NULL as a precondition?

I had an idea (for the distant future) that we should generate a human-readable report from these harnesses. We'd scrape all the __CPROVER_assumes that happen before the function call and generate a cute report that says "we've proved foo, assuming bar and baz". But I suppose this is slightly less useful for c-common as opposed to FreeRTOS, if it's already documented somewhere.

.cbmc-batch/stubs/s_sift_up_override.c
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#include <aws/common/priority_queue.h>
+
+bool __CPROVER_file_local_priority_queue_c_s_sift_up(struct aws_priority_queue *queue, size_t index) {
+    AWS_PRECONDITION(aws_priority_queue_is_valid(queue));
+    AWS_PRECONDITION(index < queue->container.length);
+    bool did_move;
+    AWS_POSTCONDITION(aws_priority_queue_is_valid(queue));
+    return did_move;
```
karkhaz: This isn't really blocking, but for future---I believe this might read more simply as
```c
{
    AWS_PRECONDITION(aws_priority_queue_is_valid(queue));
    AWS_PRECONDITION(index < queue->container.length);
    return nondet_bool();
}
```
(but that's probably a matter of taste, your implementation is fine.)

.cbmc-batch/jobs/aws_priority_queue_push/aws_priority_queue_push_harness.c
angelhof: At the moment, `AWS_PRECONDITION` and `AWS_POSTCONDITION` in the function code serve as documentation (especially after #369, which allows for explanations in pre/postconditions, was merged). However, as you noted the "real" assumptions are written in the harnesses and thus they could "drift" out-of-sync from preconditions. It would be really nice (as future work) to generate documentation from `__CPROVER_assume`s or to find a way to keep assumes and preconditions in sync.

PR #398, opened 2019-05-31
Harnesses for priority queue remove/pop/s_remove

angelhof:
Depends on #377 


angelhof:
These CBMC errors seem unrelated to this PR.

.cbmc-batch/include/proof_helpers/make_common_data_structures.h
```diff
@@ -79,9 +79,11 @@ bool aws_priority_queue_is_bounded(
     const size_t max_item_size);
 
 /**
- * Ensures members of an aws_priority_queue structure are correctly allocated
+ * Ensures members of an aws_priority_queue structure are correctly
+ * allocated and returns whether the backpointers list was allocated
+ * or not.
  */
-void ensure_priority_queue_has_allocated_members(struct aws_priority_queue *const queue);
+bool ensure_priority_queue_has_allocated_members(struct aws_priority_queue *const queue);
```
feliperodri: Functions like `ensure_*` should not return bool.

.cbmc-batch/jobs/aws_priority_queue_pop/aws_priority_queue_pop_harness.c
```diff
@@ -0,0 +1,51 @@
...
+
+#include <aws/common/priority_queue.h>
+#include <proof_helpers/make_common_data_structures.h>
+
+void aws_priority_queue_pop_harness() {
+    /* Data structure */
+    struct aws_priority_queue queue;
+
+    /* Assumptions */
+    __CPROVER_assume(aws_priority_queue_is_bounded(&queue, MAX_INITIAL_ITEM_ALLOCATION, MAX_ITEM_SIZE));
```
feliperodri: Why do you need to bound this queue? I didn't see any bounds on the `Makefile`. Could you do it without the bounds?

.cbmc-batch/jobs/aws_priority_queue_pop/aws_priority_queue_pop_harness.c
```diff
@@ -0,0 +1,51 @@
...
+    bool backpointers_allocated = ensure_priority_queue_has_allocated_members(&queue);
```
feliperodri: `ensure_priority_queue_has_allocated_members ` should not return a `bool` value.

.cbmc-batch/jobs/aws_priority_queue_pop/aws_priority_queue_pop_harness.c
```diff
@@ -0,0 +1,51 @@
...
+
+    /* Assume the function preconditions */
+    __CPROVER_assume(aws_priority_queue_is_valid(&queue));
+    void *item = can_fail_malloc(queue.container.item_size);
+
+    if (backpointers_allocated) {
```
feliperodri: Could you check whether the `backpointers` were allocated or not only by checking the data structure?

.cbmc-batch/jobs/aws_priority_queue_pop/aws_priority_queue_pop_harness.c
```diff
@@ -0,0 +1,51 @@
...
+                    can_fail_malloc(sizeof(struct aws_priority_queue_node));
+            }
+        }
+    }
+
+    /* Perform operation under verification */
+    aws_priority_queue_pop(&queue, item);
+
+    /* Assert the postconditions */
+    assert(aws_priority_queue_is_valid(&queue));
```
feliperodri: Are there any other assertions? For instance, check whether the size of the underlying `aws_array_list` decrease by 1. See `aws_array_list_pop_harness` for more possible assertions.

.cbmc-batch/jobs/aws_priority_queue_remove/Makefile
```diff
@@ -0,0 +1,41 @@
...
+# or in the "license" file accompanying this file. This file is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+###########
+include ../Makefile.aws_array_list
+
+###########
+#
```
feliperodri: Remove this line.

.cbmc-batch/jobs/aws_priority_queue_remove/aws_priority_queue_remove_harness.c
```diff
@@ -0,0 +1,55 @@
...
+
+#include <aws/common/priority_queue.h>
+#include <proof_helpers/make_common_data_structures.h>
+
+void aws_priority_queue_remove_harness() {
+    /* Data structure */
+    struct aws_priority_queue queue;
+
+    /* Assumptions */
+    __CPROVER_assume(aws_priority_queue_is_bounded(&queue, MAX_INITIAL_ITEM_ALLOCATION, MAX_ITEM_SIZE));
```
feliperodri: Again, do we really need these bounds here? I didn't see any bounds in the `Makefile`.

.cbmc-batch/jobs/aws_priority_queue_s_remove_node/aws_priority_queue_s_remove_node_harness.c
```diff
@@ -0,0 +1,55 @@
...
+    __CPROVER_assume(aws_priority_queue_is_bounded(&queue, MAX_INITIAL_ITEM_ALLOCATION, MAX_ITEM_SIZE));
+    bool backpointers_allocated = ensure_priority_queue_has_allocated_members(&queue);
+
+    /* Assume the function preconditions */
+    __CPROVER_assume(aws_priority_queue_is_valid(&queue));
+    void *item = can_fail_malloc(queue.container.item_size);
+    size_t index;
+    __CPROVER_assume(index < queue.container.length);
+
+    if (backpointers_allocated) {
```
feliperodri: See previous comments.

.cbmc-batch/jobs/aws_priority_queue_s_remove_node/aws_priority_queue_s_remove_node_harness.c
```diff
@@ -0,0 +1,55 @@
...
+#include <proof_helpers/make_common_data_structures.h>
+
+void __CPROVER_file_local_priority_queue_c_s_remove_node(struct aws_priority_queue *queue, void *item, size_t index);
+
+void aws_priority_queue_s_remove_node_harness() {
+    /* Data structure */
+    struct aws_priority_queue queue;
+
+    /* Assumptions */
+    __CPROVER_assume(aws_priority_queue_is_bounded(&queue, MAX_INITIAL_ITEM_ALLOCATION, MAX_ITEM_SIZE));
```
feliperodri: Do we really these bounds here? I didn't see any bounds in the `Makefile`.

.cbmc-batch/stubs/s_sift_either_override.c
```diff
@@ -0,0 +1,10 @@
+
+#include <aws/common/priority_queue.h>
+
+bool __CPROVER_file_local_priority_queue_c_s_sift_either(struct aws_priority_queue *queue, size_t index) {
```
feliperodri: I comment would be helpful here for documentation. Why do we need this stub?

.cbmc-batch/jobs/aws_priority_queue_pop/aws_priority_queue_pop_harness.c
angelhof: I was thinking of keeping the first version of all harnesses in `priority_queue.c` about validity and have a separate pass later for proving functional correctness properties. Especially since a lot of functions such as `memcpy` and `sift` are stubbed out in these harnesses to make them run in reasonable time.

.cbmc-batch/jobs/aws_priority_queue_s_remove_node/aws_priority_queue_s_remove_node_harness.c
angelhof: It seems that we do need it, as it bounds the containing array list and the size of its items. The bounds are defined in `Makefile.aws_array_list`

.cbmc-batch/jobs/aws_priority_queue_pop/aws_priority_queue_pop_harness.c
```diff
@@ -0,0 +1,51 @@
...
+    if (queue.backpointers.data) {
+        /* Assume that the two backpointers 0, len-1 are valid,
+         * either by being NULL or by allocating their objects. This
+         * is important for the s_swap that happens in s_remove. */
+        size_t len = queue.backpointers.length;
+        if (0 < len) {
+            ((struct aws_priority_queue_node **)queue.backpointers.data)[0] =
+                can_fail_malloc(sizeof(struct aws_priority_queue_node));
+            if (0 != len - 1) {
+                ((struct aws_priority_queue_node **)queue.backpointers.data)[len - 1] =
```
karkhaz: Note to self: I wasn't sure why `queue.backpointers.length` is guaranteed to be at least as large as `queue.backpointers.data`, and therefore why `queue.backpointers.data[queue.backpointers.length -1]` is safe.

This is actually guaranteed by the call to `aws_priority_queue_is_bounded` above (implementation [here](https://github.com/awslabs/aws-c-common/blob/master/.cbmc-batch/source/make_common_data_structures.c#L63)).

.cbmc-batch/stubs/s_sift_either_override.c
karkhaz: I agree with @feliperodri, the current comment is not sufficient. I would like to see two things please:
1. Why do we need the stub?
1. Why is it safe to stub the original function out?

.cbmc-batch/jobs/aws_priority_queue_s_remove_node/Makefile
```diff
@@ -0,0 +1,47 @@
...
+include ../Makefile.aws_array_list
+
+###########
+#
+# Runtime: 500s
+
+## This is here for the backpointer array list which contains pointers
+## to aws_priority_queue_node elements. As pointers are 64bits, it is
+## adequate to unroll the memcpy loop twice.
+UNWINDSET += memcpy_using_uint64_impl.0:2
```
angelhof: @karkhaz @tautschnig is it reasonable to assume that pointers are 64bit sized? This unwind here is adequate because it holds when pointer sizes are <= 64 bits.

.cbmc-batch/source/make_common_data_structures.c
```diff
@@ -75,20 +75,10 @@ bool aws_priority_queue_is_bounded(
     return container_is_bounded && backpointers_is_bounded;
 }
 
-bool ensure_priority_queue_has_allocated_members(struct aws_priority_queue *const queue) {
+void ensure_priority_queue_has_allocated_members(struct aws_priority_queue *const queue) {
```
karkhaz: Thanks for this change, this is a lot more sensible. So I suppose the idea is that we should never be using `ensure_*` functions inside a `__CPROVER_assume()`?

Do all the other `ensure_*` functions have void return type too? If not, it might be worth opening a PR to change all of them and fix their callers at the same time.

.cbmc-batch/stubs/aws_array_list_swap_override.c
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * FUNCTION: aws_array_list_swap
+ *
+ * We override aws_array_list_swap because mem_swap makes CBMC
+ * struggle (because of the many memcpys) and because the
+ * array_list_get_at in before the mem_swap are unneccessary if we
+ * stub out mem_swap. Instead we add a havoc assertion on the two
```
karkhaz: Where is the havoc assertion?

.cbmc-batch/stubs/s_remove_node_override.c
```diff
@@ -0,0 +1,50 @@
...
+    size_t item_index) {
+    assert(aws_priority_queue_is_valid(queue));
+    assert(item && AWS_MEM_IS_WRITABLE(item, queue->container.item_size));
+    if (aws_array_list_get_at(&queue->container, item, item_index)) {
+        /* shouldn't happen, but if it does we've already raised an error... */
+        assert(aws_priority_queue_is_valid(queue));
+        return AWS_OP_ERR;
+    }
+
+    queue->container.length -= 1;
```
karkhaz: Hmmm. We have only asserted that `queue` is valid. But it can be valid while having a length of zero, right? So this might underflow?

Please write a comment explaining why it's always okay to decrement `length`, it isn't obvious to me.

(I suppose the decrement below is safe because if `data` is non-null then length is at least 1, and that is checked by `aws_priority_queue_is_valid`?)

.cbmc-batch/stubs/s_sift_either_override.c
karkhaz: You asked for a re-review but there's still no comment here, please add one.

.cbmc-batch/source/make_common_data_structures.c
angelhof: You are right. Every other `ensure_*` is used in this manner, so there is no change to be made.

.cbmc-batch/stubs/s_sift_either_override.c
angelhof: I think I added comments in the stub. This is the [link](https://github.com/awslabs/aws-c-common/pull/398/files#diff-2d8a8fc92c4ceab7fa583e22798ca1ff) with the file changes.

.cbmc-batch/stubs/s_remove_node_override.c
angelhof: This cannot underflow because `aws_array_list_get_at` would have failed (thus exiting early) if we tried to get an index that is not in the range of the container.

I will add a comment explaining the above.

.cbmc-batch/stubs/aws_array_list_swap_override.c
angelhof: I forgot to add this before. Thanks for the catch.

PR #400, opened 2019-05-31
Adds suitable macros to check for endianness

feliperodri:

*Description of changes:*

Some of the CBMC proofs detected that a portion of the code in `lookup3.c` was not used at all, because the check for big-endian or little-endian did not use the correct flags.

Thus, we should also consider `__BYTE_ORDER__` and `__ORDER_LITTLE_ENDIAN__`. For more info, see https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html 

As a simple test, the following command should report the flags set by the C Preprocessor on a given machine:
```
cpp -dM < /dev/null | egrep '(86|BYTE)'
```


feliperodri:
@tautschnig @JonathanHenson @ColdenCullen @justinboswell @danielsn could you review this PR?

tautschnig:
I'm wondering whether we should go further and also make sure we have Visual Studio covered? See the answer by Adam Rosenfield in https://stackoverflow.com/questions/4239993/determining-endianness-at-compile-time.

feliperodri:
> Does correctness rely on one of HASH_LITTLE_ENDIAN, HASH_BIG_ENDIAN being set to 1? If so, how confident are we of completeness? Skimming SO indicates that sufficiently old compilers might not have an indicator symbol available.

@bretambrose No, this change isn't related to correctness. As far as I understand, the code take advantage of big endian or little endian to provide a more efficient implementation for `hashlittle` and `hashlittle2` functions, but they can still calculate the hash values even though `HASH_LITTLE_ENDIAN` and `HASH_BIG_ENDIAN` are set to 0. The idea of this PR is only to provide more flag options on the endianness check and then take advantage of the most efficient implementation.

ColdenCullen:
> LGTM. Shameless plug, [here](https://github.com/diffblue/cbmc/blob/develop/src/goto-cc/gcc_mode.cpp#L131) is a thing that I wrote for CBMC that maps the names of GCC architectures to pretty, simplified CBMC versions. The simplified versions also have information on whether they are little- or big-endian somewhere else in the file.

That's cool! Just curious, how would we go about using that in this case? Can we use it as anything other than just a reference?

feliperodri:
@ColdenCullen @tautschnig @karkhaz @danielsn @bretambrose could you merge this PR?

PR #406, opened 2019-06-06
Adds proof harness for aws_byte_cursor_compare* functions

feliperodri:
*Description of changes:*
1. Updates implementation of `aws_byte_cursor_compare*` functions with pre- and post-conditions;
2. Adds proof harnesses for `aws_byte_cursor_compare*` functions;

### DIFF
--- a/.cbmc-batch/jobs/aws_byte_cursor_compare_lexical/aws_byte_cursor_compare_lexical_harness.c
+++ b/.cbmc-batch/jobs/aws_byte_cursor_compare_lexical/aws_byte_cursor_compare_lexical_harness.c
@@ -43,11 +43,11 @@ void aws_byte_cursor_compare_lexical_harness() {
     save_byte_from_array(rhs.ptr, rhs.len, &old_byte_from_rhs);
 
     /* operation under verification */
-    if (aws_byte_cursor_compare_lexical((nondet_bool() ? &lhs : NULL), (nondet_bool() ? &rhs : NULL))) {
+    if (aws_byte_cursor_compare_lexical((nondet_bool() ? &lhs : NULL), (nondet_bool() ? &rhs : NULL)) == 0) {
         assert(lhs.len == rhs.len);
-        /*if (lhs.len > 0) {
+        if (lhs.len > 0) {
             assert_bytes_match(lhs.ptr, rhs.ptr, lhs.len);
-        }*/
+        }
     }
 
     /* assertions */
--- a/.cbmc-batch/jobs/aws_byte_cursor_compare_lexical/aws_byte_cursor_compare_lexical_harness.c
+++ b/.cbmc-batch/jobs/aws_byte_cursor_compare_lexical/aws_byte_cursor_compare_lexical_harness.c
@@ -51,6 +51,7 @@ void aws_byte_cursor_compare_lexical_harness() {
     }
 
     /* assertions */
+    assert(aws_byte_cursor_compare_lexical(&lhs, &lhs) == 0);
     assert(aws_byte_cursor_is_valid(&lhs));
     assert(aws_byte_cursor_is_valid(&rhs));
     if (lhs.len != 0) {
--- a/.cbmc-batch/jobs/aws_byte_cursor_compare_lexical/aws_byte_cursor_compare_lexical_harness.c
+++ b/.cbmc-batch/jobs/aws_byte_cursor_compare_lexical/aws_byte_cursor_compare_lexical_harness.c
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/byte_buf.h>
+#include <proof_helpers/make_common_data_structures.h>
+#include <proof_helpers/proof_allocators.h>
+
+void aws_byte_cursor_compare_lexical_harness() {
+    /* parameters */
+    struct aws_byte_cursor lhs;
+    struct aws_byte_cursor rhs;
+
+    /* assumptions */
+    __CPROVER_assume(aws_byte_cursor_is_bounded(&lhs, MAX_BUFFER_SIZE));
+    ensure_byte_cursor_has_allocated_buffer_member(&lhs);
+    __CPROVER_assume(aws_byte_cursor_is_valid(&lhs));
+    if (nondet_bool()) {
+        rhs = lhs;
+    } else {
+        __CPROVER_assume(aws_byte_cursor_is_bounded(&rhs, MAX_BUFFER_SIZE));
+        ensure_byte_cursor_has_allocated_buffer_member(&rhs);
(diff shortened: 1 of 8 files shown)


feliperodri:
@tautschnig @karkhaz @angelhof @danielsn @markrtuttle @lucasmt could you review the proof harnesses?

source/byte_buf.c
```diff
@@ -665,29 +665,39 @@ bool aws_byte_cursor_satisfies_pred(const struct aws_byte_cursor *source, aws_by
 }
 
 int aws_byte_cursor_compare_lexical(const struct aws_byte_cursor *lhs, const struct aws_byte_cursor *rhs) {
-
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(lhs));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(rhs));
+    AWS_PRECONDITION(AWS_MEM_IS_READABLE(lhs->ptr, lhs->len));
```
angelhof: As this is included in `aws_byte_cursor_is_valid` when `len > 0`, I think it is important to at least include a comment here that explains the intent of this precondition.

.cbmc-batch/jobs/aws_byte_cursor_compare_lexical/aws_byte_cursor_compare_lexical_harness.c
```diff
@@ -0,0 +1,62 @@
...
+    /* save current state of the data structure */
+    struct aws_byte_cursor old_lhs = lhs;
+    struct store_byte_from_buffer old_byte_from_lhs;
+    save_byte_from_array(lhs.ptr, lhs.len, &old_byte_from_lhs);
+    struct aws_byte_cursor old_rhs = rhs;
+    struct store_byte_from_buffer old_byte_from_rhs;
+    save_byte_from_array(rhs.ptr, rhs.len, &old_byte_from_rhs);
+
+    /* operation under verification */
+    if (aws_byte_cursor_compare_lexical((nondet_bool() ? &lhs : NULL), (nondet_bool() ? &rhs : NULL)) == 0) {
```
lucasmt: An interesting case to check would also be when the same pointer is passed twice to the function (e.g. `aws_byte_cursor_compare_lexical(&lhs, &lhs)`).

.cbmc-batch/jobs/aws_byte_cursor_compare_lookup/aws_byte_cursor_compare_lookup_harness.c
```diff
@@ -0,0 +1,65 @@
...
+
+void aws_byte_cursor_compare_lookup_harness() {
+    /* parameters */
+    struct aws_byte_cursor lhs;
+    struct aws_byte_cursor rhs;
+    /**
+     * The specification for the function requires that the buffer
+     * be at least 256 bytes.
+     */
+    uint8_t *lookup_table[256];
```
lucasmt: This should be `uint8_t lookup_table[256];` (no *).

.cbmc-batch/jobs/aws_byte_cursor_compare_lexical/aws_byte_cursor_compare_lexical_harness.c
```diff
@@ -0,0 +1,63 @@
...
+        assert(lhs.len == rhs.len);
+        if (lhs.len > 0) {
+            assert_bytes_match(lhs.ptr, rhs.ptr, lhs.len);
+        }
+    }
+
+    /* assertions */
+    assert(aws_byte_cursor_compare_lexical(&lhs, &lhs) == 0);
```
karkhaz: This seems like an "operation under verification" rather than an "assertion". I suggest that you move it further up, so that it's associated with the other call to `aws_byte_cursor_compare_lexical`.

Also, I don't think it's okay to do
```c
aws_byte_cursor_compare_lexical(&lhs, &rhs);
aws_byte_cursor_compare_lexical(&lhs, &lhs);
assert(aws_byte_cursor_is_valid(&lhs));
```
because what if the first call to `aws_byte_cursor_compare_lexical` makes `lhs` invalid, but the second one makes it valid again? The harness doesn't catch this case. I think we should check for validity immediately after we do *each* of the function calls.

My suggestion is that you only make a single call to `aws_byte_cursor_compare_lexical` which deals with both identical and distinct `lhs` and `rhs`. Suggest:
```c
if (aws_byte_cursor_compare_lexical(
      (nondet_bool() ? &lhs : NULL),
      (nondet_bool() ? &lhs
                     : nondet_bool() ? rhs
                                     : NULL)) == 0) {
```

.cbmc-batch/jobs/aws_byte_cursor_compare_lookup/aws_byte_cursor_compare_lookup_harness.c
```diff
@@ -0,0 +1,66 @@
...
+    save_byte_from_array(rhs.ptr, rhs.len, &old_byte_from_rhs);
+
+    /* operation under verification */
+    if (aws_byte_cursor_compare_lookup((nondet_bool() ? &lhs : NULL), (nondet_bool() ? &rhs : NULL), lookup_table) ==
+        0) {
+        assert(lhs.len == rhs.len);
+    }
+
+    /* assertions */
+    assert(aws_byte_cursor_compare_lookup(&lhs, &lhs, lookup_table) == 0);
```
karkhaz: Same comment as above, we're not actually checking the validity of the cursor after the first call to `aws_byte_cursor_compare_lookup` here.

source/byte_buf.c
```diff
@@ -711,12 +726,17 @@ int aws_byte_cursor_compare_lookup(
     }
 
     if (lhs_curr < lhs_end) {
+        AWS_POSTCONDITION(aws_byte_cursor_is_valid(lhs));
```
karkhaz: Do we need to write these postconditions three times? It seems that we assert them no matter what we're returning...suggest writing the postconditions just above `if (lhs_curr < lhs_end)`.

.cbmc-batch/jobs/aws_byte_cursor_compare_lexical/aws_byte_cursor_compare_lexical_harness.c
feliperodri: That's great! However, keep in mind that in many cases now we should not pass `NULL` pointers to entry points in proofs. Since all `AWS_PRECONDITIONS` turned into assertions, we need to keep the assumptions in the proof harness; therefore, one of the assumptions here is that `lhs` and `rhs` must be different than NULL.

feliperodri: Actually, I want to keep both calls otherwise I won't be able to check the following assertions when calling `aws_byte_cursor_compare_lexical(&lhs, &lhs);`

```
        assert(lhs.len == rhs.len);
        if (lhs.len > 0) {
            assert_bytes_match(lhs.ptr, rhs.ptr, lhs.len);
        }
```

In addition, the first call cannot make `lhs` or `rhs` invalid, because we ensure their validity in the postconditions.

.cbmc-batch/jobs/aws_byte_cursor_compare_lookup/aws_byte_cursor_compare_lookup_harness.c
feliperodri: Same comment as above. We check the validity inside the function as a postcondition.

source/byte_buf.c
feliperodri: Nice catch!

feliperodri:
@karkhaz could you review this PR? I've implemented your comments.

feliperodri:
@ColdenCullen @justinboswell @bretambrose could you review the changes in `byte_buf.c`?

ColdenCullen:
Hey @karkhaz, still waiting on you

PR #407, opened 2019-06-06
Assert unchanged byte buffers in the definition of aws_byte_buf_eq

angelhof:
This PR is an experiment as a first step of a proposal to "push" assertions (and assumptions) to the function definitions. At the moment, function pre and postconditions check whether the arguments and return values satisfy simple validity constraints. In contrast, more complex assertions are checked in harnesses. By moving assertions to the definitions, the code becomes self documented and the risk of assertions (and assumptions) in harnesses lagging behind the definition is reduced.

This PR mostly serves as a discussion thread for the proposal.

There are two main discussions points:

1. What is the ideal amount of work to do when testing? Is it reasonable to copy the input argument byte buffer and check that every byte in the byte buffer after the execution of the function is the same as before? In my opinion we should check the assertion while testing by default, and only check some simplified version of it when a specific test takes too much time to finish.

2. What should the macros used for these conditions look like? We have to find a balance between expressiveness and visual clarity, as it wouldn't make sense to introduce heavyweight conditions in every function definition. In my opinion, a good first step would be to introduce macros for frequently used assertions (such as equivalence of an argument before and after the execution of a function), so that they can be easily used by anyone developing the code.

### DIFF
--- a/include/aws/common/common.h
+++ b/include/aws/common/common.h
@@ -41,6 +41,7 @@
 #endif
 
 #define AWS_CONCAT(A, B) A ## B
+#define GHOST(x) ghost_ ## x
 
 #define AWS_ZERO_STRUCT(object)                                                                                        \
         do {                                                                                                           \
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -913,4 +913,20 @@ AWS_STATIC_IMPL bool aws_byte_buf_write_be64(struct aws_byte_buf *buf, uint64_t 
     return aws_byte_buf_write(buf, (uint8_t *)&x, 8);
 }
 
+#ifdef CBMC
+#    include <proof_helpers/make_common_data_structures.h>
+#    include <proof_helpers/proof_allocators.h>
+#    define GHOST_BYTE(buf) ghost_##buf##_byte
+#    define AWS_GHOST_BYTE_BUF(buf)                                                                                    \
+        struct aws_byte_buf GHOST(buf) = *buf;                                                                         \
+        struct store_byte_from_buffer GHOST_BYTE(buf);                                                                 \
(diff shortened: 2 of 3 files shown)


angelhof:
@tautschnig @danielsn @feliperodri could you share your ideas about the above?

angelhof:
@karkhaz, @markrtuttle could you review and comment on the above?

karkhaz:
Thanks for taking the initiative to do this. I think this is an excellent idea, modulo the caveats you mentioned. I think the key point is
> a good first step would be to introduce macros for frequently used assertions

Let's be data-driven about this. We now have a fairly wide variety of proofs in different parts of the codebase, so we're in a position to survey our proofs and figure out what the most commonly-used constructs are. So: I suggest that you compile a list of frequently-used constructs, including data on how often we use them, how cheap they are to execute, and how often we *could* potentially use them in the codebase (as opposed to the harnesses). The list doesn't have to be perfect, but just to give us an idea of where to start.

For the macro that you've proposed in particular: I'm not sure why it's necessary in that particular part of the code. Is it not sufficient to just use `const` everywhere (which is what the function already does)?

angelhof:
@karkhaz After taking a look in the array list module, the following two assertions seem to be used very often.

- Equivalence of the data structure in case a set/pop/push operation failed. In these functions the structure is not const, but in the failing case it should remain as it was before.
- Relation of the length of the list before and after the operation. This is particularly used in the successful paths of the set/pop/push operations.

Do you think the above two assertions are useful enough to have in the code (instead of just in the harnesses)?

include/aws/common/common.h
```diff
@@ -41,6 +41,7 @@
 #endif
 
 #define AWS_CONCAT(A, B) A ## B
+#define GHOST(x) ghost_ ## x
```
ColdenCullen: Please put this in the CBMC conditional block in byte_buf.h, this will not be useful for anyone else.

PR #409, opened 2019-06-06
Ring buffer implementation supporting single producer, single consumer locklessly.

JonathanHenson:


source/ring_buffer.c
```diff
@@ -0,0 +1,196 @@
...
+
+void aws_ring_buffer_clean_up(struct aws_ring_buffer *ring_buf) {
+    if (ring_buf->allocation) {
+        aws_mem_release(ring_buf->allocator, ring_buf->allocation);
+    }
+
+    AWS_ZERO_STRUCT(*ring_buf);
+}
+
+int aws_ring_buffer_acquire(struct aws_ring_buffer *ring_buf, size_t requested_size, struct aws_byte_buf *dest) {
```
justinboswell: Since ```acquire``` and ```acquire_up_to``` are basically the same except for the last section, why not just make one function that takes a ```size_t *returned_size```. If ```returned_size``` is non-null, do the best effort allocation. Then implement both of these in terms of that one function, less copy-pasta, only have to get it right once.

include/aws/common/ring_buffer.h
```diff
@@ -0,0 +1,81 @@
...
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/atomics.h>
+#include <aws/common/byte_buf.h>
+
+/**
+ * Lockless ring buffer implementation that is thread safe assuming a single thread acquires and a single thread
+ * releases. For any other use case (other than the single-threaded use-case), you must manage thread-safety manually.
+ * Also note: release must happen in the same order as acquire.
```
justinboswell: We do nothing to enforce this, which could lead to absolutely lawless overwrites. We also don't store the size or a nonce at the head of each buffer, so we can't check to see if release is safe/correctly ordered. Are we ok with that?

tests/ring_buffer_test.c
```diff
@@ -0,0 +1,382 @@
...
+    ASSERT_SUCCESS(aws_ring_buffer_acquire(&ring_buffer, 4, &vended_buffer_2));
+    ASSERT_PTR_EQUALS(ptr + 12, vended_buffer_2.buffer);
+    ASSERT_UINT_EQUALS(4, vended_buffer_2.capacity);
+    ASSERT_TRUE(aws_ring_buffer_buf_belongs_to_pool(&ring_buffer, &vended_buffer_2));
+
+    ASSERT_ERROR(AWS_ERROR_NO_AVAILABLE_BUFFERS, aws_ring_buffer_acquire(&ring_buffer, 1, &vended_buffer_1));
+
+    aws_ring_buffer_release(&ring_buffer, &vended_buffer_1);
+
+    /* we should turn over right here. and capacity should now be one less than*/
```
justinboswell: What does this capacity comment mean?

source/ring_buffer.c
justinboswell: Nevermind, they're not as identical as I thought.

include/aws/common/ring_buffer.h
JonathanHenson: we don't need size or nonce on head because we're filling in byte buf.

I'm not sure how we would check the ordering with that anyways.

source/ring_buffer.c
JonathanHenson: because the code is complicated enough as it is, and I didn't want to make it worse by framenting it across functions. There's subtle differences now, and it would be too messy to refactor it just to share less than 20 logical lines of code.

tests/ring_buffer_test.c
JonathanHenson: messed up copy/paste

include/aws/common/ring_buffer.h
JonathanHenson: also, I don't think we'd be able to have a lockless release at that point.

source/ring_buffer.c
```diff
@@ -0,0 +1,206 @@
...
+            *dest = aws_byte_buf_from_empty_array(head_cpy, requested_size);
+            return AWS_OP_SUCCESS;
+        }
+        /* After N wraps */
+    } else if (tail_cpy < head_cpy) {
+        /* prefer the head space for efficiency. */
+        if ((size_t)(ring_buf->allocation_end - head_cpy) > requested_size) {
+            aws_atomic_store_ptr(&ring_buf->head, head_cpy + requested_size);
+            *dest = aws_byte_buf_from_empty_array(head_cpy, requested_size);
+            return AWS_OP_SUCCESS;
```
graebm: [trivial] I **think** this could be: `if ((allocation_end - head_cpy) >= requested_size)`
can't we vend all the way to the end of the buffer, without leaving a gap, since we'd still end up with head != tail?

source/ring_buffer.c
```diff
@@ -0,0 +1,206 @@
...
+
+        if ((size_t)(tail_cpy - ring_buf->allocation) > requested_size) {
+            aws_atomic_store_ptr(&ring_buf->head, ring_buf->allocation + requested_size);
+            *dest = aws_byte_buf_from_empty_array(ring_buf->allocation, requested_size);
+            return AWS_OP_SUCCESS;
+        }
+
+        /* now vend as much as possible, once again preferring head space. */
+        if (head_space > 1 && head_space > tail_space) {
+            size_t returnable_size = head_space - 1;
```
graebm: trivial: `head_space >= tail_space` or we'd choose tail_space when they're equal

source/ring_buffer.c
```diff
@@ -0,0 +1,206 @@
...
+            return AWS_OP_SUCCESS;
+        }
+        /* after N wraps */
+    } else if (tail_cpy < head_cpy) {
+        size_t head_space = ring_buf->allocation_end - head_cpy;
+        size_t tail_space = tail_cpy - ring_buf->allocation;
+
+        if (!head_space && tail_space <= 1) {
+            return aws_raise_error(AWS_ERROR_NO_AVAILABLE_BUFFERS);
+        }
```
graebm: if my suspicion about not needing to leave an empty byte at the end is true, we can alter a bunch of logic in this branch...

source/common.c
```diff
@@ -428,6 +428,10 @@ static struct aws_error_info errors[] = {
         AWS_ERROR_ENVIRONMENT_UNSET,
         "System call failure when unsetting an environment variable."
     ),
+    AWS_DEFINE_ERROR_INFO_COMMON(
+        AWS_ERROR_NO_AVAILABLE_BUFFERS,
+        "A buffer acquisition source (e.g. a ring buffer or something), does not have available buffers to vend."
+    ),
```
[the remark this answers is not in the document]
JonathanHenson: will do

source/ring_buffer.c
```diff
@@ -0,0 +1,206 @@
...
+
+    ring_buf->allocation = aws_mem_acquire(allocator, size);
+
+    if (!ring_buf->allocation) {
+        return AWS_OP_ERR;
+    }
+
+    ring_buf->allocator = allocator;
+    aws_atomic_store_ptr(&ring_buf->head, ring_buf->allocation);
+    aws_atomic_store_ptr(&ring_buf->tail, ring_buf->allocation);
```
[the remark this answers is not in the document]
JonathanHenson: will do

source/ring_buffer.c
```diff
@@ -0,0 +1,206 @@
...
+    return aws_raise_error(AWS_ERROR_NO_AVAILABLE_BUFFERS);
+}
+
+static inline bool s_buf_belongs_to_pool(const struct aws_ring_buffer *ring_buffer, const struct aws_byte_buf *buf) {
+    return buf->buffer >= ring_buffer->allocation && buf->buffer + buf->capacity <= ring_buffer->allocation_end;
+}
+
+void aws_ring_buffer_release(struct aws_ring_buffer *ring_buffer, const struct aws_byte_buf *buf) {
+    AWS_ASSERT(s_buf_belongs_to_pool(ring_buffer, buf));
+    aws_atomic_store_ptr(&ring_buffer->tail, buf->buffer + buf->capacity);
```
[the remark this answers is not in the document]
JonathanHenson: good point, and sure.

include/aws/common/ring_buffer.h
```diff
@@ -0,0 +1,86 @@
...
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/atomics.h>
+#include <aws/common/byte_buf.h>
```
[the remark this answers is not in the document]
JonathanHenson: okay

source/ring_buffer.c
JonathanHenson: allocation_end is one past the end of the allocation

graebm: say the ring has 2 bytes, and we've vended 1, so:
allocation: 0
allocation_end: 2
tail: 0
head: 1
request: 1

We *could* vend that last byte, but with the code as written: `if ((2 - 1) > 1)` we won't vend that last byte

source/ring_buffer.c
```diff
@@ -0,0 +1,205 @@
...
+    if (requested_size == 0) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+
+    uint8_t *tail_cpy = aws_atomic_load_ptr(&ring_buf->tail);
+    uint8_t *head_cpy = aws_atomic_load_ptr(&ring_buf->head);
+
+    /* this branch is, we don't have any vended buffers. */
+    if (head_cpy == tail_cpy) {
+        size_t ring_space = ring_buf->allocation_end - ring_buf->allocation;
```
angelhof: This should be `size_t ring_space = ring_buf->allocation_end - ring_buf->allocation - 1;`. PR #410 makes this change and adds harnesses for the ring buffer functions.

source/ring_buffer.c
```diff
@@ -0,0 +1,205 @@
...
+    /* after N + 1 wraps */
+    if (tail_cpy > head_cpy) {
+        size_t space = tail_cpy - head_cpy;
+        /* this shouldn't be possible. */
+        AWS_ASSERT(space);
+        space -= 1;
+
+        size_t returnable_size = space > requested_size ? requested_size : space;
+
+        if (returnable_size >= minimum_size) {
```
angelhof: If `minimum size == 0` and `tail_cpy - head_cpy == 1`, then `dest` will point to a buffer that points to NULL and has 0 capacity. Is this behaviour expected? PR #410 mentions this and adds a proof harness.

source/ring_buffer.c
JonathanHenson: we updated this with a change to init

source/ring_buffer.c
JonathanHenson: updated

source/ring_buffer.c
JonathanHenson: will update in a future PR.

PR #411, opened 2019-06-08
Implement aws_hash_table_remove_element

ColdenCullen:
Helpful if you're conditionally removing an element from a table based on it's value.

Use case: In HPACK, I need to remove an element from the dynamic table only if it's value pointer is exactly what I want it to be. Therefore, I call find, check the value, and remove.

### DIFF
--- a/tests/hash_table_test.c
+++ b/tests/hash_table_test.c
@@ -99,6 +99,11 @@ static int s_test_hash_table_create_find_fn(struct aws_allocator *allocator, voi
         TEST_VAL_STR_2);
 
     ASSERT_HASH_TABLE_ENTRY_COUNT(&hash_table, 2);
+
+    err_code = aws_hash_table_remove_element(&hash_table, pElem);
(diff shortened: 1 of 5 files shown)


tests/hash_table_test.c
```diff
@@ -211,6 +216,13 @@ static int s_test_hash_table_string_create_find_fn(struct aws_allocator *allocat
         ((struct aws_string *)pElem->value)->len,
         "Returned value for binary bytes should have been %s",
         "hunter2");
+
+    aws_string_destroy((struct aws_string *)pElem->key);
+    aws_string_destroy(pElem->value);
+    ret = aws_hash_table_remove_element(&hash_table, pElem);
+    ASSERT_SUCCESS(ret, "Hash Map remove element should have succeeded.");
+    ASSERT_HASH_TABLE_ENTRY_COUNT(&hash_table, 2);
+
```
marcomagdy: prolly late to the party, but I'd also assert that looking up that item yields not-found

PR #415, opened 2019-06-11
Extend the harness for priority queue s_swap with fc assertions

angelhof:
Extend the harness for priority queue s_swap with functional correctness assertions.


source/priority_queue.c
```diff
@@ -206,6 +210,14 @@ bool aws_priority_queue_is_valid(const struct aws_priority_queue *const queue) {
     return pred_is_valid && container_is_valid && backpointer_list_is_valid && backpointer_struct_is_valid;
 }
 
+bool aws_backpointer_index_valid(const struct aws_priority_queue *const queue, size_t index) {
+    if (queue->backpointers.data && index < queue->backpointers.length) {
+        struct aws_priority_queue_node *node = ((struct aws_priority_queue_node **)queue->backpointers.data)[index];
```
karkhaz: Is the cast to `aws_priority_queue_node **` (rather than just single `*`) correct?

angelhof: Thanks for the comment. From my understanding the backpointers array list holds pointers to `struct aws_priority_queue_node`. So the backpointers.data should be a double pointer. Does this sound reasonable?

karkhaz:
@angelhof please rebase so that this is mergeable. I've assigned c-common people for review.

angelhof:
There seems to be a problem with the s_sift_either proof (even though I didn't directly change it). It seems to be stuck on CBMC batch. I have to debug before this is merged.

angelhof:
The downstream build error seems to be unrelated to these changes as it happens in #421 too. I also tried to run the downstream build on master locally and it fails at the same place.

source/priority_queue.c
```diff
@@ -198,12 +222,33 @@ bool aws_priority_queue_is_valid(const struct aws_priority_queue *const queue) {
     bool backpointer_list_item_size = queue->backpointers.item_size == sizeof(struct aws_priority_queue_node *);
     bool lists_equal_lengths = queue->backpointers.length == queue->container.length;
     bool backpointers_non_zero_current_size = queue->backpointers.current_size > 0;
+
+    /* This check must be guarded, as it is not efficient, neither
+     * when running tests nor CBMC */
+#ifdef PQ_BACKPOINTERS_CHECK
```
danielsn: Is this the best name here?

source/priority_queue.c
```diff
@@ -180,15 +184,35 @@ void aws_priority_queue_init_static(
...
 
-bool aws_priority_queue_is_valid(const struct aws_priority_queue *const queue) {
-    /* Pointer validity checks */
+bool aws_priority_queue_backpointer_index_valid(const struct aws_priority_queue *const queue, size_t index) {
+    if (queue->backpointers.data == NULL) {
+        return true;
+    }
+    if (index < queue->backpointers.length) {
+        struct aws_priority_queue_node *node = ((struct aws_priority_queue_node **)queue->backpointers.data)[index];
+        return node == NULL || AWS_MEM_IS_WRITABLE(node, sizeof(struct aws_priority_queue_node));
```
danielsn: Use brackets

source/priority_queue.c
```diff
@@ -198,12 +222,33 @@ bool aws_priority_queue_is_valid(const struct aws_priority_queue *const queue) {
     bool backpointer_list_item_size = queue->backpointers.item_size == sizeof(struct aws_priority_queue_node *);
     bool lists_equal_lengths = queue->backpointers.length == queue->container.length;
     bool backpointers_non_zero_current_size = queue->backpointers.current_size > 0;
+
+    /* This check must be guarded, as it is not efficient, neither
+     * when running tests nor CBMC */
+#ifdef PQ_BACKPOINTERS_CHECK
+    bool backpointers_allocated = aws_priority_queue_backpointers_valid_deep(queue);
```
danielsn: maybe call the function call _allocated? Or change the variable name.

source/priority_queue.c
angelhof: I changed that name to `AWS_DEEP_CHECKS`, which should be used to guard any validity check that is not O(1).

PR #419, opened 2019-06-12
Task log

bretambrose:
Adds a type string tag to the task struct and logs task scheduling and execution


source/task_scheduler.c
```diff
@@ -103,6 +138,13 @@ void aws_task_scheduler_schedule_future(
     AWS_ASSERT(task);
     AWS_ASSERT(task->fn);
 
+    AWS_LOGF_DEBUG(
```
graebm: TRACE

source/task_scheduler.c
```diff
@@ -15,8 +15,37 @@
 
 #include <aws/common/task_scheduler.h>
 
+#include <aws/common/logging.h>
+
+#include <inttypes.h>
+
 static const size_t DEFAULT_QUEUE_SIZE = 7;
 
+const char *aws_task_status_to_c_string(enum aws_task_status status) {
```
graebm: we already use the term "c_str" in byte_buf.h and string.h
aws_task_status_to_c_str()
or
aws_task_status_c_str()

source/task_scheduler.c
```diff
@@ -15,8 +15,37 @@
...
+            return "<Canceled>";
+
+        default:
+            return "<Unknown>";
+    }
+}
+
+void aws_task_run(struct aws_task *task, enum aws_task_status status) {
+    AWS_ASSERT(task->fn);
+    AWS_LOGF_DEBUG(
```
graebm: TRACE

source/task_scheduler.c
```diff
@@ -87,6 +116,12 @@ void aws_task_scheduler_schedule_now(struct aws_task_scheduler *scheduler, struc
     AWS_ASSERT(task);
     AWS_ASSERT(task->fn);
 
+    AWS_LOGF_DEBUG(
```
graebm: TRACE

source/task_scheduler.c
bretambrose: My current rule of thumb is to use TRACE only for debugging information that's too costly (spammy) to log under normal circumstances.  This doesn't feel like it meets that, so I'd prefer to keep it DEBUG for now.  Same for the others.

PR #427, opened 2019-06-21
Timebomb allocator

graebm:
Timebomb allocator fakes running out of memory after then Nth allocation.

This is for use in tests.

### DIFF
--- a/include/aws/testing/aws_test_allocators.h
+++ b/include/aws/testing/aws_test_allocators.h
@@ -34,7 +34,7 @@ struct aws_timebomb_impl {
 };
 
 static void *s_timebomb_mem_acquire(struct aws_allocator *timebomb_alloc, size_t size) {
-    struct aws_timebomb_impl *timebomb_impl = timebomb_alloc->impl;
+    struct aws_timebomb_impl *timebomb_impl = (struct aws_timebomb_impl *)timebomb_alloc->impl;
     void *ptr = NULL;
 
     aws_mutex_lock(&timebomb_impl->mutex);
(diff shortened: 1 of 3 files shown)


include/aws/testing/aws_test_allocators.h
```diff
@@ -0,0 +1,97 @@
...
+
+static void aws_timebomb_allocator_clean_up(struct aws_allocator *timebomb_alloc) {
+    if (timebomb_alloc) {
+        struct aws_timebomb_impl *timebomb_impl = timebomb_alloc->impl;
+        aws_mutex_clean_up(&timebomb_impl->mutex);
+        aws_mem_release(timebomb_impl->wrapped_allocator, timebomb_impl);
+    }
+}
+
+static void aws_timebomb_allocator_reset(struct aws_allocator *timebomb_alloc, size_t fail_after_n_allocations) {
```
danielsn: How we we make sure that nobody else is still using it - should there be a barrier here?

include/aws/testing/aws_test_allocators.h
```diff
@@ -0,0 +1,97 @@
...
+/** \file
+ * Alternate allocators for use in testing.
+ */
+
+/**
+ * Timebomb allocator fakes running out of memory after then Nth allocation.
+ * Once this allocator starts failing, it never succeeds, even if memory is released.
+ * Wraps an existing allocator.
+ */
+struct aws_timebomb_impl {
```
danielsn: Would it be interresting to potentially track how many bytes were requested, and give a max # of bytes?

include/aws/testing/aws_test_allocators.h
graebm: It just resets the timer, it's not for resetting the whole allocator. I'll rename this

include/aws/testing/aws_test_allocators.h
graebm: That would be interesting.
But that would be a job for a different allocator.

This is the timebomb allocator.

include/aws/testing/aws_test_allocators.h
```diff
@@ -0,0 +1,97 @@
...
+    struct aws_allocator *timebomb_allocator,
+    struct aws_allocator *wrapped_allocator,
+    size_t fail_after_n_allocations) {
+
+    AWS_ZERO_STRUCT(*timebomb_allocator);
+
+    struct aws_timebomb_impl *timebomb_impl = aws_mem_calloc(wrapped_allocator, 1, sizeof(struct aws_timebomb_impl));
+    ASSERT_NOT_NULL(timebomb_impl);
+
+    timebomb_allocator->mem_acquire = s_timebomb_mem_acquire;
```
danielsn: I guess since you don't define ```calloc``` or ```realloc```, you just rely on them falling back to use the standard allocator?

graebm: Yes.

PR #430, opened 2019-06-24
Replace AWS_FATAL_ASSERT with explicit AWS_FATAL_PRECONDITION and AWS…

tegansiobhan:
…_FATAL_POSTCONDITION macros

*Issue #, if available:*

*Description of changes:*

Created new AWS_FATAL_PRECONDITION_1, AWS_FATAL_PRECONDITION_2, AWS_FATAL_POSTCONDITION_1, and AWS_FATAL_POSTCONDITION_2 macros. Replaced calls to AWS_FATAL_ERROR with these macros. 


include/aws/common/string.h
```diff
@@ -68,7 +68,7 @@ AWS_STATIC_IMPL const uint8_t *aws_string_bytes(const struct aws_string *str) {
  * Returns true if bytes of string are the same, false otherwise.
  */
 AWS_STATIC_IMPL bool aws_string_eq(const struct aws_string *a, const struct aws_string *b) {
-    AWS_FATAL_ASSERT(a != NULL && b != NULL);
+    AWS_FATAL_PRECONDITION_1(a != NULL && b != NULL);
```
angelhof: I would suggest using `AWS_FATAL_PRECONDITION` instead of `AWS_FATAL_PRECONDITION_1`. Similarly for postconditions.

tegansiobhan: Thanks! Have made the changes!

danielsn:
We should consider whether we want to rename ```AWS_PRECONDITION``` to ```AWS_ASSERT_PRECONDITION``` for clarity.

include/aws/common/common.h
```diff
@@ -115,30 +115,41 @@
...
-#define AWS_MEM_IS_WRITABLE(base, len) (((len) == 0) || (base))
+#    define AWS_MEM_IS_READABLE(base, len) (((len) == 0) || (base))
+#    define AWS_MEM_IS_WRITABLE(base, len) (((len) == 0) || (base))
 #endif
 
 // The UNUSED is used to silence the complains of GCC for zero arguments in variadic macro
 #define AWS_PRECONDITION(...) GET_MACRO(__VA_ARGS__, AWS_PRECONDITION_2, AWS_PRECONDITION_1, UNUSED)(__VA_ARGS__)
+#define AWS_FATAL_PRECONDITION(...) GET_MACRO(__VA_ARGS__, AWS_FATAL_PRECONDITION_2, AWS_FATAL_PRECONDITION_1, UNUSED)(__VA_ARGS__)
 #define AWS_POSTCONDITION(...) GET_MACRO(__VA_ARGS__, AWS_POSTCONDITION_2, AWS_POSTCONDITION_1, UNUSED)(__VA_ARGS__)
+#define AWS_FATAL_POSTCONDITION(...) GET_MACRO(__VA_ARGS__, AWS_FATAL_POSTCONDITION_2, AWS_FATAL_POSTCONDITION_1, UNUSED)(__VA_ARGS__)
```
feliperodri: Since you're changing these macros, could you also add `AWS_ERROR_PRECONDITION` and `AWS_ERROR_POSTCONDITION` macros that would raise error codes when the condition is violated? Something like...

```
#define AWS_ERROR_PRECONDITION(cond, err)       \
    do {                                        \
        if (cond) {                             \
            return aws_raise_error(err);        \
        }                                       \
    } while (0)
```
@danielsn @ColdenCullen what do you think?

ColdenCullen: THIS IS COOL. I think it can be done in a separate PR tho, let's get just this in first.

danielsn: Glad you like it.  Working on a PR now.

Also, for consistency, should I rename the current one that asserts ```AWS_ASSERT_PRECONDITION```? Or just leave it as is?

ColdenCullen: I don't feel strongly either way? I don't mind the idea of all 3 being different, but I also like that ASSERT is the default. @JonathanHenson @graebm @justinboswell any thoughts?

include/aws/common/array_list.inl
```diff
@@ -86,7 +86,26 @@ bool aws_array_list_is_valid(const struct aws_array_list *AWS_RESTRICT list) {
     bool required_size_is_valid = (aws_mul_size_checked(list->length, list->item_size, &required_size) == AWS_OP_SUCCESS);
     bool current_size_is_valid = (list->current_size >= required_size);
     bool data_is_valid = ((list->current_size == 0 && list->data == NULL) || AWS_MEM_IS_WRITABLE(list->data, list->current_size));
-    return required_size_is_valid && current_size_is_valid && data_is_valid;
+    bool item_size_is_valid = (list->item_size!=0);
+    return required_size_is_valid && current_size_is_valid && data_is_valid && item_size_is_valid;}
```
angelhof: The brace should be in a new line? Did clang-format miss that?

include/aws/common/array_list.inl
```diff
@@ -86,7 +86,26 @@ bool aws_array_list_is_valid(const struct aws_array_list *AWS_RESTRICT list) {
...
+    bool current_size_is_wiped = (list->current_size == 0);
+    bool item_size_is_wiped = (list->item_size == 0);
+    bool length_is_wiped = (list->length == 0);
+    bool data_is_wiped = (list->data == NULL);
+    bool alloc_is_wiped = (list->alloc == NULL);
+    return current_size_is_wiped && item_size_is_wiped && length_is_wiped && data_is_wiped && alloc_is_wiped;
+}
+
+AWS_STATIC_IMPL
+void aws_array_list_debug_print(const struct aws_array_list * list) {
```
angelhof: Maybe this (and other debug print functions) should go in a different file debug.c or something.

source/priority_queue.c
```diff
@@ -337,13 +340,19 @@ static int s_remove_node(struct aws_priority_queue *queue, void *item, size_t it
...
+        aws_array_list_get_at(&queue->backpointers, &backpointer, swap_with);
+    }
+
     if (backpointer) {
         backpointer->current_index = SIZE_MAX;
     }
 
     aws_array_list_pop_back(&queue->container);
-    aws_array_list_pop_back(&queue->backpointers);
+
```
angelhof: It seems possible to merge get_at and pop_back to happen behind one is_wiped check. 
```
    if (!aws_array_list_is_wiped(&queue->backpointers)) {
        aws_array_list_get_at(&queue->backpointers, &backpointer, swap_with);
        if (backpointer) {
            backpointer->current_index = SIZE_MAX;
         }
         aws_array_list_pop_back(&queue->backpointers);
    }
```

tegansiobhan:
To fix broken proofs in aws_priority_queue_push_ref and aws_priority_queue_push, added the assumption that item_size must be non-zero into aws_array_list_is_valid(...). In order to account for backpointers (which can be an all zero struct), added a new function aws_array_list_is_wiped(...). When priority queue functions are called on a backpointer, this criteria is now checked first. @dsn

include/aws/common/array_list.inl
```diff
@@ -86,9 +91,34 @@ bool aws_array_list_is_valid(const struct aws_array_list *AWS_RESTRICT list) {
...
+        ((list->current_size == 0 && list->data == NULL) || AWS_MEM_IS_WRITABLE(list->data, list->current_size));
+    bool item_size_is_valid = (list->item_size!=0);
+    return required_size_is_valid && current_size_is_valid && data_is_valid && item_size_is_valid;
+}
+
+AWS_STATIC_IMPL
+bool aws_array_list_is_wiped(const struct aws_array_list *AWS_RESTRICT list) {
+    if(!list) {
+        return false;
+    }
```
angelhof: Should we return `false` if list is NULL? Or should this be checked in an assert?

angelhof: @danielsn what do you think?

tegansiobhan:
@JonathanHenson The build fail for aws-c-common-downstream is due to a bug in aws-c-io that is unrelated to this PR.

tegansiobhan:
@danielsn @ColdenCullen Do you think this PR can be merged now?

source/priority_queue.c
tegansiobhan: Done and pushed.

include/aws/common/array_list.inl
tegansiobhan: Yeah if there are other debug print functions we could do that, but I think that would be separate from this PR.

danielsn:
Downstream failure is 
```
         ^~~~~~~~~~~~~~~~~~~~~ 
1 error generated. 
```

PR #434, opened 2019-06-25
add AWS_ERROR_PRECONDTION macro, and use it.

danielsn:
*Issue #, if available:*

*Description of changes:*
Added ```AWS_ERROR_PRECONDITION()``` macro, and use it whenever appropiate.

### DIFF
--- a/.cbmc-batch/jobs/aws_byte_buf_init/aws_byte_buf_init_harness.c
+++ b/.cbmc-batch/jobs/aws_byte_buf_init/aws_byte_buf_init_harness.c
@@ -26,11 +26,8 @@ void aws_byte_buf_init_harness() {
     size_t capacity;
 
     /* assumptions */
-    if (nondet_bool()) {
-        ASSUME_CAN_FAIL_ALLOCATOR(allocator);
-    } else {
-        allocator = NULL;
-    }
+    /* Precondition: allocator is non-null */
+    ASSUME_CAN_FAIL_ALLOCATOR(allocator);
     __CPROVER_assume(capacity <= MAX_BUFFER_SIZE);
 
     if (!aws_byte_buf_init(&buf, allocator, capacity)) {
--- a/.cbmc-batch/jobs/aws_byte_buf_init/aws_byte_buf_init_harness.c
+++ b/.cbmc-batch/jobs/aws_byte_buf_init/aws_byte_buf_init_harness.c
@@ -19,18 +19,13 @@
 
 void aws_byte_buf_init_harness() {
     /* data structure */
-    struct aws_byte_buf buf;
+    struct aws_byte_buf buf; /* Precondition: buf is non-null */
 
     /* parameters */
-    struct aws_allocator *allocator;
+    struct aws_allocator *allocator = can_fail_allocator(); /* Precondition: allocator is non-null */
     size_t capacity;
 
-    /* assumptions */
-    /* Precondition: allocator is non-null */
-    ASSUME_CAN_FAIL_ALLOCATOR(allocator);
-    __CPROVER_assume(capacity <= MAX_BUFFER_SIZE);
-
-    if (!aws_byte_buf_init(&buf, allocator, capacity)) {
+    if (aws_byte_buf_init(&buf, allocator, capacity) == AWS_OP_SUCCESS) {
         /* assertions */
         assert(aws_byte_buf_is_valid(&buf));
         assert(buf.allocator == allocator);
--- a/source/byte_buf.c
+++ b/source/byte_buf.c
@@ -24,9 +24,8 @@
 #endif
 
 int aws_byte_buf_init(struct aws_byte_buf *buf, struct aws_allocator *allocator, size_t capacity) {
-    if (!buf || !allocator) {
-        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
-    }
+    AWS_ERROR_PRECONDITION(buf);
+    AWS_ERROR_PRECONDITION(allocator);
 
     buf->buffer = (capacity == 0) ? NULL : aws_mem_acquire(allocator, capacity);
     if (capacity != 0 && buf->buffer == NULL) {
@@ -41,9 +40,9 @@ int aws_byte_buf_init(struct aws_byte_buf *buf, struct aws_allocator *allocator,
 }
 
 int aws_byte_buf_init_copy(struct aws_byte_buf *dest, struct aws_allocator *allocator, const struct aws_byte_buf *src) {
-    if (!allocator || !dest || !aws_byte_buf_is_valid(src)) {
-        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
-    }
+    AWS_ERROR_PRECONDITION(allocator);
+    AWS_ERROR_PRECONDITION(dest);
+    AWS_ERROR_PRECONDITION(aws_byte_buf_is_valid(src));
 
     if (!src->buffer) {
         AWS_ZERO_STRUCT(*dest);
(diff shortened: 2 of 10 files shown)


angelhof:
This is really nice :)

include/aws/common/common.h
```diff
@@ -119,26 +120,46 @@
...
+#define AWS_ERROR_POSTCONDITION_2(cond, err)	\
+  do {						\
+    if(!(cond)) {				\
+      return aws_raise_error(err);		\
+    }						\
+  } while(0)
+#define AWS_ERROR_POSTCONDITION_3(cond, err, explanation) AWS_ERROR_POSTCONDITION_2(cond, err)
+
+/* The UNUSED is used to silence the complains of GCC for zero arguments in variadic macro */
+#define AWS_PRECONDITION(...) GET_MACRO(__VA_ARGS__, INVALID, AWS_PRECONDITION_2, AWS_PRECONDITION_1, UNUSED)(__VA_ARGS__)
```
ColdenCullen: Minor nit, use `UNUSED` both places instead of `INVALID`?

include/aws/common/common.h
```diff
@@ -119,26 +120,46 @@
...
+#    define AWS_MEM_IS_WRITABLE(base, len) (((len) == 0) || (base))
+#endif /* CBMC */
+
+#define AWS_ERROR_PRECONDITION_2(cond, err)	\
+  do {						\
+    if(!(cond)) {				\
+      return aws_raise_error(err);		\
+    }						\
+  } while(0)
+#define AWS_ERROR_PRECONDITION_3(cond, err, explanation) AWS_ERROR_PRECONDITION_2(cond, err)
```
ColdenCullen: It would be really cool if we could make this log explanation, but I'm not sure how to do that without adding a million parameters. Anyone have any ideas?

source/byte_buf.c
```diff
@@ -24,9 +24,8 @@
 #endif
 
 int aws_byte_buf_init(struct aws_byte_buf *buf, struct aws_allocator *allocator, size_t capacity) {
-    if (!buf || !allocator) {
-        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
-    }
+    AWS_ERROR_PRECONDITION(buf);
+    AWS_ERROR_PRECONDITION(allocator);
```
ColdenCullen: These two I think can be not-the-ERROR variant. If you pass `NULL` for `buf` or `allocator` you should probably eat the assert.

source/byte_buf.c
```diff
@@ -41,9 +40,9 @@ int aws_byte_buf_init(struct aws_byte_buf *buf, struct aws_allocator *allocator,
 }
 
 int aws_byte_buf_init_copy(struct aws_byte_buf *dest, struct aws_allocator *allocator, const struct aws_byte_buf *src) {
-    if (!allocator || !dest || !aws_byte_buf_is_valid(src)) {
-        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
-    }
+    AWS_ERROR_PRECONDITION(allocator);
+    AWS_ERROR_PRECONDITION(dest);
```
ColdenCullen: the IS_VALID below can probably stay, but these guys probably should assert too.

source/byte_buf.c
```diff
@@ -146,9 +145,9 @@ int aws_byte_buf_init_copy_from_cursor(
     struct aws_byte_buf *dest,
     struct aws_allocator *allocator,
     struct aws_byte_cursor src) {
-    if (!allocator || !dest || !aws_byte_cursor_is_valid(&src)) {
-        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
-    }
+    AWS_ERROR_PRECONDITION(allocator);
+    AWS_ERROR_PRECONDITION(dest);
```
ColdenCullen: Same, `allocator` and `dest` should probably assert.

source/ring_buffer.c
```diff
@@ -49,11 +49,7 @@ void aws_ring_buffer_clean_up(struct aws_ring_buffer *ring_buf) {
 int aws_ring_buffer_acquire(struct aws_ring_buffer *ring_buf, size_t requested_size, struct aws_byte_buf *dest) {
     AWS_PRECONDITION(aws_ring_buffer_is_valid(ring_buf));
     AWS_PRECONDITION(aws_byte_buf_is_valid(dest));
-    if (requested_size == 0) {
-        AWS_POSTCONDITION(aws_ring_buffer_is_valid(ring_buf));
-        AWS_POSTCONDITION(aws_byte_buf_is_valid(dest));
-        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
-    }
+    AWS_ERROR_PRECONDITION(requested_size != 0, AWS_ERROR_INVALID_ARGUMENT);
```
ColdenCullen: You can use the 1-param overload here right?

include/aws/common/common.h
danielsn: I have an idea.  I'll try and let you know if it works.

source/byte_buf.c
danielsn: ```FATAL_ASSERT``` or ```assert```?

source/ring_buffer.c
danielsn: Yep.  Willdo.

source/byte_buf.c
ColdenCullen: I don't feel strongly either way, but I think vanilla `AWS_PRECONDITION` is more prevalent, maybe stick with that?

include/aws/common/common.h
danielsn: Done.

include/aws/common/common.h
danielsn: My plan was to add a new function ```aws_raise_error_and_log(errorcode, log_format_string, ...)```, and then call it from here (and anywhere else that might be useful). But that seems like a big enough change it should be its own followup PR.

source/byte_buf.c
danielsn: Done

source/byte_buf.c
danielsn: Done

source/byte_buf.c
danielsn: Done

include/aws/common/common.h
ColdenCullen: I like it! Looking forward to it.

ColdenCullen:
Looks like lots of tests are failing with `Expected error but no error occurred; rv=0, aws_last_error=0000` on Windows

.cbmc-batch/jobs/aws_byte_buf_init/aws_byte_buf_init_harness.c
```diff
@@ -26,11 +26,8 @@ void aws_byte_buf_init_harness() {
     size_t capacity;
 
     /* assumptions */
-    if (nondet_bool()) {
-        ASSUME_CAN_FAIL_ALLOCATOR(allocator);
-    } else {
-        allocator = NULL;
-    }
+    /* Precondition: allocator is non-null */
+    ASSUME_CAN_FAIL_ALLOCATOR(allocator);
```
feliperodri: Should we stop using `ASSUME_CAN_FAIL_ALLOCATOR` and replace it with `struct aws_allocator *allocator = can_fail_allocator();`? In addition, no need for `/* assumptions */` in this case anymore.

danielsn: Done

danielsn:
I have debugged the windows build failures - it turns out that they are due to the way windows handles varidic macros. https://stackoverflow.com/questions/5134523/msvc-doesnt-expand-va-args-correctly

I am experimenting with fixes.

danielsn:
I fixed the macros, and added some ```static_asserts``` to ensure they stay correct.

PR #435, opened 2019-06-26
Fix of null ptr in aws_array_list and aws_priority_queue init functions 

tegansiobhan:
*Issue #, if available:*

*Description of changes:*
Fix of null ptr bug in aws_array_list_init_dynamic, aws_array_list_init_static, aws_priority_queue_init_dynamic, and aws_priority_queue_init_static. 

Updated harnesses of each these functions. For the static init functions, the assumption that the list or queue cannot be null is made explicit. This is now consistent with their implementation. For the dynamic init functions, we explicitly test the case where the list or queue is null which is now handled in the implementation. 

Additionally, the harness for aws_array_list_init_static now checks that the raw_array remains unchanged. This is also consistent with the harness for aws_priority_queue_init_static. 


.cbmc-batch/jobs/aws_array_list_init_dynamic/aws_array_list_init_dynamic_harness.c
```diff
@@ -21,26 +21,25 @@
...
 
     /* assumptions */
-    ASSUME_VALID_MEMORY(list);
     ASSUME_CAN_FAIL_ALLOCATOR(allocator);
     __CPROVER_assume(initial_item_allocation <= MAX_INITIAL_ITEM_ALLOCATION);
     __CPROVER_assume(item_size <= MAX_ITEM_SIZE);
 
     /* perform operation under verification */
-    if (!aws_array_list_init_dynamic(list, allocator, initial_item_allocation, item_size)) {
+    if (!aws_array_list_init_dynamic(nondet_bool() ? &list : NULL, allocator, initial_item_allocation, item_size)) {
```
feliperodri: For readability, replace `!aws_array_list_init_dynamic(...)` with `aws_array_list_init_dynamic(...) == AWS_OP_SUCCESS`

include/aws/common/array_list.inl
```diff
@@ -62,14 +62,15 @@ void aws_array_list_init_static(
     void *raw_array,
     size_t item_count,
     size_t item_size) {
-    AWS_FATAL_ASSERT(raw_array);
-    AWS_FATAL_ASSERT(item_count);
-    AWS_FATAL_ASSERT(item_size);
+    AWS_FATAL_PRECONDITION(list);
+    AWS_FATAL_PRECONDITION(raw_array);
+    AWS_FATAL_PRECONDITION(item_count);
+    AWS_FATAL_PRECONDITION(item_size);
```
feliperodri: Should we raise an `AWS_ERROR_INVALID_ARGUMENT` error or keep the `AWS_FATAL_PRECONDITION`? I'd suggest to pick one and use for both `init_static` and `init_dynamic`

feliperodri: @ColdenCullen

.cbmc-batch/jobs/aws_array_list_init_dynamic/aws_array_list_init_dynamic_harness.c
tegansiobhan: Thanks - done!

include/aws/common/array_list.inl
tegansiobhan: The static init functions are void so I could not return the AWS_ERROR_INVALID_ARGUMENT error as in the dynamic versions. However, I could raise a AWS_FATAL_PRECONDITION in the dynamic functions.

source/priority_queue.c
```diff
@@ -159,6 +159,10 @@ int aws_priority_queue_init_dynamic(
     size_t item_size,
     aws_priority_queue_compare_fn *pred) {
 
+    if (queue == NULL) {
```
ColdenCullen: You can use AWS_ERROR_PRECONDITION here soon (as soon as #434 is merged)

.cbmc-batch/jobs/aws_array_list_init_dynamic/aws_array_list_init_dynamic_harness.c
```diff
@@ -21,26 +21,26 @@
...
+    struct aws_array_list list;
 
     /* parameters */
     struct aws_allocator *allocator;
     size_t item_size;
     size_t initial_item_allocation;
 
     /* assumptions */
-    ASSUME_VALID_MEMORY(list);
     ASSUME_CAN_FAIL_ALLOCATOR(allocator);
```
danielsn: Are we still using these assumes?

.cbmc-batch/jobs/aws_array_list_init_static/aws_array_list_init_static_harness.c
```diff
@@ -21,28 +21,33 @@
...
-    assert(list->current_size == initial_item_allocation * item_size);
-    assert_bytes_match((uint8_t *)list->data, raw_array, len);
+    assert(aws_array_list_is_valid(&list));
+    assert(list.alloc == NULL);
+    assert(list.item_size == item_size);
+    assert(list.length == 0);
+    assert(list.current_size == initial_item_allocation * item_size);
+    assert_bytes_match((uint8_t *)list.data, raw_array, len);
+
+    assert_byte_from_buffer_matches(raw_array, &old_byte);
```
danielsn: nice

.cbmc-batch/jobs/aws_priority_queue_init_dynamic/aws_priority_queue_init_dynamic_harness.c
```diff
@@ -31,22 +31,22 @@ void aws_priority_queue_init_dynamic_harness() {
...
     ASSUME_CAN_FAIL_ALLOCATOR(allocator);
     __CPROVER_assume(initial_item_allocation <= MAX_INITIAL_ITEM_ALLOCATION);
     __CPROVER_assume(item_size <= MAX_ITEM_SIZE);
     __CPROVER_assume(!aws_mul_size_checked(initial_item_allocation, item_size, &len));
 
     /* perform operation under verification */
     uint8_t *raw_array = bounded_malloc(len);
-    if (!aws_priority_queue_init_dynamic(queue, allocator, initial_item_allocation, item_size, nondet_compare)) {
+    if (!aws_priority_queue_init_dynamic(
+            nondet_bool() ? &queue : NULL, allocator, initial_item_allocation, item_size, nondet_compare)) {
```
danielsn: What if allocator is null?

source/priority_queue.c
tegansiobhan: Good idea! I've gone ahead and made those changes.

.cbmc-batch/jobs/aws_array_list_init_dynamic/aws_array_list_init_dynamic_harness.c
tegansiobhan: Looking at more recent proofs, seems we have replaced that assume with can_fail_allocator(). I have updated these proofs to match.

.cbmc-batch/jobs/aws_priority_queue_init_dynamic/aws_priority_queue_init_dynamic_harness.c
tegansiobhan: Then we have many failing proofs. I added an explicit check for this in both init_dynamic harnesses and fixed the bug by checking for a null alloc in aws_array_list_init_dynamic(). Since the init_dynamic of priority_queue calls this function, there was not a need to check for it there though we could.

tegansiobhan:
Downstream failure is 

```
```

tegansiobhan:
@danielsn @ColdenCullen I think this is ready to be merged? Could you take a look?

source/priority_queue.c
```diff
@@ -176,6 +178,8 @@ void aws_priority_queue_init_static(
     size_t item_size,
     aws_priority_queue_compare_fn *pred) {
 
+    AWS_FATAL_PRECONDITION(queue);
```
ColdenCullen: Why are the preconditions different? Why is one an ERROR and one a FATAL? Why does one do `!= NULL` but the other doesn't?

tegansiobhan: The static methods are void so cannot return an error. I changed the dynamic inits to use FATAL  asserts instead of ERROR and modified the harnesses to explicitly make the same assumptions. This means we are not checking for the null ptrs directly anymore, but keeping the assumptions between the harness and the c code in agreement. 

In the case of priority_queue, some of the assumptions are redundant since it calls the array_list function but I thought it better to make everything explicit.

PR #436, opened 2019-06-26
Split common.h into multiple .h files, named after what is actually in them.

danielsn:
*Description of changes:*
```common.h``` had become very large, full of many different macros and function declarations.  As a side note, because it redefined ```bool``` it confused ```clang-format``` and hence couldn't have an enforced standard code format.  This PR splits it into several ```.h``` files, with descriptive names, all of which can have ```clang-format``` run on them.  This also breaks the recursive import where ```error``` imported ```common```, and ```common``` imported ```error```.

### CODE REVIEW

**Comment by danielsn** [comment]
@justinboswell I can rename them to ```.inl```

**Comment by danielsn** [comment]
@justinboswell to clarify, it was my intention that each .h file include its dependencies, and be able to be included independently.  Given that, should I rename them to .inl?

File: include/aws/common/error_codes.h [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+enum aws_common_error {
```
angelhof: Maybe we could merge error_codes.h and error.h?

justinboswell:
> @justinboswell to clarify, it was my intention that each .h file include its dependencies, and be able to be included independently. Given that, should I rename them to .inl?

If that's the case, I'm happy to just go with it. We should probably fix assert.inl then and make it consistent, eh?

include/aws/common/allocator.h
```diff
@@ -0,0 +1,110 @@
...
+    void *(*mem_realloc)(struct aws_allocator *allocator, void *oldptr, size_t oldsize, size_t newsize);
+    /* Optional method; if not supported, this pointer must be NULL */
+    void *(*mem_calloc)(struct aws_allocator *allocator, size_t num, size_t size);
+    void *impl;
+};
+
+/* Avoid pulling in CoreFoundation headers in a header file. */
+#ifdef __MACH__
+struct __CFAllocator;
+typedef const struct __CFAllocator *CFAllocatorRef;
```
ColdenCullen: I think you can move this into the #ifdef block below, which should make it clearer to read.

include/aws/common/error.h
```diff
@@ -16,8 +16,13 @@
  * permissions and limitations under the License.
  */
 
-#include <aws/common/common.h>
+#include <aws/common/assert.inl>
```
ColdenCullen: why include this .inl directly? I believe the intent is to include the .h file, and have this come along for the ride.

include/aws/common/invariants.h
```diff
@@ -0,0 +1,53 @@
...
+#include <aws/common/macros.h>
+
+/**
+ * Define function contracts.
+ * When the code is being verified using CBMC these contracts are formally verified;
+ * When the code is built in debug mode, they are checked as much as possible using assertions
+ * When the code is built in production mode, they are not checked.
+ * Violations of the function contracts are undefined behaviour.
+ */
+#ifdef CBMC
```
ColdenCullen: I'd vote to merge this file into assert.h

danielsn: Done

include/aws/common/error.h
danielsn: Done

include/aws/common/allocator.h
danielsn: Done

include/aws/common/error_codes.h
danielsn: Done

PR #444, opened 2019-07-01
Replace abort() with assert(0) so that CBMC catches it

angelhof:
Currently abort() is not caught as an error by CBMC, and this could lead to failed assertions in `AWS_FATAL_ASSERT` to be missed. 

I created this as a draft PR to see which proofs fail after this change.


danielsn:
you could also make ```AWS_FATAL_ASSERT``` turninto ```assert``` in CBMC mode.

angelhof:
> you could also make `AWS_FATAL_ASSERT` turninto `assert` in CBMC mode.

That would certainly make sense together with replacing `abort()` with `assert(0)`. I will do that.

angelhof:
Note: The CBMC proof error here seems to be the same as in #430, so I will not debug this before #430 is merged.

angelhof:
The error in downstream seem to be unrelated to this PR.
```
         ^~~~~~~~~~~~~~~~~~~~~ 
1 error generated. 
```

ColdenCullen:
PLEASE DON'T MERGE THIS YET. @graebm is running into similar issues and may need to glue more changes onto this.

angelhof:
@ColdenCullen @graebm any update on that? Can we merge this?

PR #445, opened 2019-07-02
Move standard logging implementations into common, move io error tran…

bretambrose:
Standard logger implementations to common per request.
Downstream build will fail until io and auth are updated


include/aws/common/log_channel.h
```diff
@@ -0,0 +1,82 @@
+
+#ifndef AWS_COMMON_LOG_CHANNEL_H
+#define AWS_COMMON_LOG_CHANNEL_H
+
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
danielsn: Is this really 2010-2019?

bretambrose: Well the same question could be asked of pretty much every .c and .h file at this point

PR #447, opened 2019-07-03
Add macro for AWS_RETURN_WITH_POSTCONDITION and use it

danielsn:
*Description of changes:*
The current way we handle postconditions ends up with a series of ```AWS_POSTCONDITION``` calls, followed by a ```return foo``` statement.  This can make the code feel cluttered, and can make it hard to tell at a glance whether 
1. all the postconditions are in the proper place, directly before a return
2. Whether all the returns have similar postconditions.

Since postconditions should be tightly tied to returns, it makes sense to have a macro that combines the two.  This PR proposes such a macro, and trials it on one module (hash-table)


include/aws/common/assert.h
```diff
@@ -119,6 +119,22 @@ AWS_EXTERN_C_END
...
+
+#define AWS_RETURN_WITH_POSTCONDITION2(_rval, postcondition)                                                           \
+    do {                                                                                                               \
+        AWS_POSTCONDITION1(postcondition);                                                                             \
+        return _rval;                                                                                                  \
+    } while (0)
+
+#define AWS_SUCCEED_WITH_POSTCONDITION(postcondition) AWS_RETURN_WITH_POSTCONDITION2(AWS_OP_SUCCESS, postcondition)
+
+#define AWS_RETURN_WITH_POSTCONDITION(...) CALL_OVERLOAD(AWS_RETURN_WITH_POSTCONDITION, __VA_ARGS__)
```
ColdenCullen: Does this work? It'd be less code:
```c
#define AWS_RETURN_WITH_POSTCONDITION(_rval, ...)                                                           \
    do {                                                                                                               \
        AWS_POSTCONDITION(__VA_ARGS__);                                                                             \
        return _rval;                                                                                                  \
    } while (0)

#define AWS_SUCCEED_WITH_POSTCONDITION(...) AWS_RETURN_WITH_POSTCONDITION(AWS_OP_SUCCESS, __VA_ARGS__)
```

danielsn: Works on OSX.  Lets see how it does on other compilers.

danielsn:
The downstream issue is due to the bug fixed in https://github.com/awslabs/aws-c-common/pull/451

graebm:
Personally, I think this makes the code less readable.
but looks like this was merged a second ago...

PR #453, opened 2019-07-10
aws_is_zeroed()

graebm:


include/aws/common/zero.h
```diff
@@ -0,0 +1,88 @@
...
+            return false;
+        }
+    }
+
+    /* Update buf to where u64 checks left off */
+    buf = buf_u64 + num_u64_checks;
+    bufsize = bufsize % 8;
+
+    /* Check 8 bits at a time */
+    const uint8_t *buf_u8 = (const uint8_t *)buf;
```
danielsn: You can replace this loop with a ```switch(bufsize & 0x7) { ...``` unrolling.

graebm: I ran this through Godbolt and the compiler did unroll it
I didn't want to get *too* hung up on theoretical efficiency improvements, since the obvious one is to use vectorized instructions, but I don't want to spend a week getting that working across all compilers right now. I just want to get a good-enough version in.

graebm: https://godbolt.org/z/5UG8je

graebm: I also thought about a different version for 32bit processors that took 32bit steps instead of 64bit steps, but it compiled down to essentially the same thing, so I just kept it simple.

angelhof:
I ran CBMC locally on `aws_priority_queue_top` and I see the problem (which should probably be the same for every failing harness). The loop in `aws_is_zeroed` is not unwinded and so CBMC complains that it cannot explore the whole execution. 

You can solve this by  adding this line in the Makefile in the harness directory:
```
UNWINDSET += __CPROVER_file_local_zero_h_aws_is_zeroed.0:3
```
This line instructs CBMC to unwind the 0 loop (first) of `aws_is_zeroed` 3 times for this proof. I think that  3 unwindings are enough because in this proof `aws_is_zeroed` is called for an array list struct (`queue->backpointers`). If you add this line it would be nice to write a comment next to it that explains why the number of unwindings is adequate for this proof.

Note that the `__CPROVER_file_local_zero_h_` needs to be added because of the way that static functions are handled by our process.

Suggestion: You could remove the following two lines from the Makefile in `.cbmc-batch/jobs/aws_priority_queue_top/` as they don't offer any benefit for the proof. The CBMC internal `memset` implementation is good enough.
```
UNWINDSET += memset_impl.0:41
...
DEPENDENCIES += $(HELPERDIR)/stubs/memset_override.c
```

include/aws/common/zero.h
```diff
@@ -0,0 +1,98 @@
+#ifndef AWS_COMMON_ZERO_H
+#define AWS_COMMON_ZERO_H
+
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
danielsn: Should this just be 2019?

include/aws/common/zero.h
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Set each byte in the struct to zero.
+ */
+#define AWS_ZERO_STRUCT(object)                                                                                        \
+    do {                                                                                                               \
+        memset(&(object), 0, sizeof(object));                                                                          \
+    } while (0)
+
+/**
```
danielsn: Might want a note that this only words for arrays declared as such. This won't work for arrays allocated with malloc.

include/aws/common/zero.h
graebm: I genuinely have no idea. I copy/pasted "2010-2018" from another header and thought I was real cool for updating 18 to 19

include/aws/common/zero.h
graebm: done

include/aws/common/zero.h
```diff
@@ -0,0 +1,99 @@
...
+ * Set each byte in the array to zero.
+ * Does not work with arrays of unknown bound.
+ */
+#define AWS_ZERO_ARRAY(array)                                                                                          \
+    do {                                                                                                               \
+        memset((void *)(array), 0, sizeof(array));                                                                     \
+    } while (0)
```
ColdenCullen: I don't think you need the do { while since it's just a one liner. Just do:
```c
#define AWS_ZERO_ARRAY(array) memset((void *)(array), 0, sizeof(array))
```

include/aws/common/zero.h
```diff
@@ -0,0 +1,99 @@
...
+        }
+/* clang-format on */
+#else
+#    define AWS_IS_ZEROED(object) aws_is_zeroed(&(object), sizeof(object))
+#endif
+
+/**
+ * Returns whether each byte is zero.
+ */
+AWS_STATIC_IMPL
```
ColdenCullen: make sure codegen.c includes this file plz.

include/aws/common/zero.h
```diff
@@ -0,0 +1,99 @@
...
+#define AWS_ZERO_STRUCT(object)                                                                                        \
+    do {                                                                                                               \
+        memset(&(object), 0, sizeof(object));                                                                          \
+    } while (0)
+
+/**
+ * Set each byte in the array to zero.
+ * Does not work with arrays of unknown bound.
+ */
+#define AWS_ZERO_ARRAY(array)                                                                                          \
```
ColdenCullen: I worry about having a macro and function with the same name. Should we call the function _impl or something so people don't call it directly?

include/aws/common/zero.h
graebm: so much feedback on this code I cut/paste from macros.h

include/aws/common/zero.h
graebm: it does. common.h includes zero.h

include/aws/common/zero.h
graebm: went with: aws_is_mem_zeroed(buf, bufsize)

include/aws/common/zero.h
```diff
@@ -0,0 +1,96 @@
...
+
+#include <string.h>
+
+/**
+ * Set each byte in the struct to zero.
+ */
+#define AWS_ZERO_STRUCT(object)                                                                                        \
+    do {                                                                                                               \
+        memset(&(object), 0, sizeof(object));                                                                          \
+    } while (0)
```
ColdenCullen: LOL same here but I won't make you do it right this moment.

graebm: I did it

PR #454, opened 2019-07-13
Ensure error list stays in sync

graebm:
- Abort with helpful message if **any** library registers an error list that's out of sync.
- Add unit test that loads aws-c-common's error list, so we can freak out if it's out of sync.
- Modify MACRO used to defin aws-c-common's error list, so it can't actually get out of sync anymore.

### DIFF
--- a/source/error.c
+++ b/source/error.c
@@ -167,8 +167,13 @@ void aws_register_error_info(const struct aws_error_info_list *error_info) {
     /* Assert that error info entries are in the right order. */
     for (int i = 1; i < error_info->count; ++i) {
         int expected_code = min_range + i;
-        if (error_info->error_list[i].error_code != expected_code) {
-            fprintf(stderr, "Error %s is at wrong index of error info list.\n", error_info->error_list[i].literal_name);
+        const struct aws_error_info *info = &error_info->error_list[i];
+        if (info->error_code != expected_code) {
+            if (info->error_code) {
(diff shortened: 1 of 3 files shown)


source/error.c
```diff
@@ -163,12 +163,23 @@ void aws_register_error_info(const struct aws_error_info_list *error_info) {
         abort();
     }
 
+#if DEBUG_BUILD
+    /* Assert that error info entries are in the right order. */
+    for (int i = 1; i < error_info->count; ++i) {
+        int expected_code = min_range + i;
+        if (error_info->error_list[i].error_code != expected_code) {
+            fprintf(stderr, "Error %s is at wrong index of error info list.\n", error_info->error_list[i].literal_name);
+            abort();
```
JonathanHenson: Don’t we have an AWS_ABORT macro or something?

source/error.c
```diff
@@ -163,12 +163,28 @@ void aws_register_error_info(const struct aws_error_info_list *error_info) {
         abort();
     }
 
+#if DEBUG_BUILD
+    /* Assert that error info entries are in the right order. */
+    for (int i = 1; i < error_info->count; ++i) {
+        int expected_code = min_range + i;
```
ColdenCullen: `const`

source/error.c
justinboswell: AWS_FATAL_ASSERT() will abort.

PR #455, opened 2019-07-15
Enable -Wunreachable-code

ColdenCullen:
MSVC checks this, so it's probably worth turning on for clang/gcc.


ColdenCullen:
This was a bad idea and I'm filled with regerts, plz no one look at this.

PR #458, opened 2019-07-22
Linked hash table

JonathanHenson:
Preserves insertion order and provides iterator access.

### DIFF
--- a/include/aws/common/linked_hash_table.h
+++ b/include/aws/common/linked_hash_table.h
@@ -1,0 +1,113 @@
+#ifndef AWS_COMMON_LINKED_HASH_TABLE_H
+#define AWS_COMMON_LINKED_HASH_TABLE_H
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+#include <aws/common/hash_table.h>
+#include <aws/common/linked_list.h>
+
+/**
+ * Simple linked hash table. Preserves insertion order, and can be iterated in insertion order.
+ *
+ * You can also change the order safely without altering the shape of the underlying hash table.
+ */
+struct aws_linked_hash_table {
+    struct aws_allocator *allocator;
+    struct aws_linked_list list;
+    struct aws_hash_table table;
+    aws_hash_callback_destroy_fn *user_on_value_destroy;
+};
+
+/**
+ * Linked-List node stored in the table. This is the node type that will be returned in
+ * aws_linked_hash_table_get_iteration_list().
+ */
+struct aws_linked_hash_table_node {
+    struct aws_linked_list_node node;
+    struct aws_linked_hash_table *table;
+    const void *key;
+    void *value;
+};
+
+AWS_EXTERN_C_BEGIN
+
+/**
+ * Initializes the table. Sets up the underlying hash table and linked list.
+ * For the other parameters, see aws/common/hash_table.h. Hash table
+ * semantics of these arguments are preserved.
+ */
+AWS_COMMON_API
+int aws_linked_hash_table_init(
+    struct aws_linked_hash_table *table,
+    struct aws_allocator *allocator,
+    aws_hash_fn *hash_fn,
+    aws_hash_callback_eq_fn *equals_fn,
+    aws_hash_callback_destroy_fn *destroy_key_fn,
+    aws_hash_callback_destroy_fn *destroy_value_fn,
+    size_t initial_item_count);
+
+/**
+ * Cleans up the table. Elements in the table will be evicted and cleanup
+ * callbacks will be invoked.
+ */
+AWS_COMMON_API
+void aws_linked_hash_table_clean_up(struct aws_linked_hash_table *table);
+
+/**
+ * Finds element in the table by key. If found, AWS_OP_SUCCESS will be
+ * returned. If not found, AWS_OP_SUCCESS will be returned and *p_value will be
+ * NULL.
+ *
+ * If any errors occur AWS_OP_ERR will be returned.
+ */
+AWS_COMMON_API
+int aws_linked_hash_table_find(struct aws_linked_hash_table *table, const void *key, void **p_value);
+
+/**
+ * Puts `p_value` at `key`. If an element is already stored at `key` it will be replaced.
+ */
+AWS_COMMON_API
+int aws_linked_hash_table_put(struct aws_linked_hash_table *table, const void *key, void *p_value);
+
+/**
+ * Removes item at `key` from the table.
+ */
+AWS_COMMON_API
+int aws_linked_hash_table_remove(struct aws_linked_hash_table *table, const void *key);
+
+/**
+ * Clears all items from the table.
+ */
+AWS_COMMON_API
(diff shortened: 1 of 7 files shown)


include/aws/common/linked_hash_table.h
```diff
@@ -0,0 +1,111 @@
...
+void aws_linked_hash_table_clear(struct aws_linked_hash_table *table);
+
+/**
+ * returns number of elements in the table.
+ */
+AWS_COMMON_API
+size_t aws_linked_hash_table_get_element_count(const struct aws_linked_hash_table *table);
+
+/**
+ * returns the underlying linked list for iteration. You must not do this in a multi-threaded context.
```
bretambrose: Seems a little strange to make this comment here.  Presumably all of the mutable apis are not usable concurrently and read-only iteration would be ok wouldn't it?

include/aws/common/linked_list.h
```diff
@@ -212,6 +212,34 @@ AWS_STATIC_IMPL void aws_linked_list_insert_after(
     AWS_POSTCONDITION(after->next == to_add);
 }
 
+/**
+ * Swaps the order two nodes in the linked list.
```
bretambrose: Something seems weird here.  The postconditions imply a and b must be adjacent with a before b, but the comments don't mention anything about such requirements.   The preconditions don't verify any adjacency.

include/aws/common/linked_list.h
```diff
@@ -212,6 +212,34 @@ AWS_STATIC_IMPL void aws_linked_list_insert_after(
...
+ */
+AWS_STATIC_IMPL void aws_linked_list_swap_nodes(struct aws_linked_list_node *a, struct aws_linked_list_node *b) {
```
bretambrose: It would be comforting to have unit tests for happy case,  a == b, head (of list, not the .head member) tail (same) cases, as well as separated (non-adjacent) if applicable

include/aws/common/linked_list.h
```diff
@@ -212,6 +212,34 @@ AWS_STATIC_IMPL void aws_linked_list_insert_after(
...
+    AWS_PRECONDITION(aws_linked_list_node_next_is_valid(b));
+    AWS_PRECONDITION(a != NULL);
+    AWS_PRECONDITION(b != NULL);
+
+    struct aws_linked_list_node *a_prev_cpy = a->prev;
+    struct aws_linked_list_node *a_prev_next = a->next;
+
+    a->prev->next = b;
+    a->prev = b->prev;
+    a->next = b->next;
```
bretambrose: Are we missing a->next->prev = b; ?

include/aws/common/linked_list.h
```diff
@@ -212,6 +212,34 @@ AWS_STATIC_IMPL void aws_linked_list_insert_after(
     AWS_POSTCONDITION(after->next == to_add);
 }
 
+/**
+ * Swaps the order two nodes in the linked list.
+ */
+AWS_STATIC_IMPL void aws_linked_list_swap_nodes(struct aws_linked_list_node *a, struct aws_linked_list_node *b) {
+    AWS_PRECONDITION(aws_linked_list_node_next_is_valid(a));
+    AWS_PRECONDITION(aws_linked_list_node_next_is_valid(b));
+    AWS_PRECONDITION(a != NULL);
```
bretambrose: After looking at the linked list impl, it seems like prev should be valid too

include/aws/common/linked_list.h
```diff
@@ -212,6 +212,34 @@ AWS_STATIC_IMPL void aws_linked_list_insert_after(
...
+    a->prev = b->prev;
+    a->next = b->next;
+
+    b->prev->next = a;
+    b->prev = a_prev_cpy;
+    b->next->prev = a;
+    b->next = a_prev_next;
+
+    AWS_POSTCONDITION(aws_linked_list_node_next_is_valid(a));
+    AWS_POSTCONDITION(aws_linked_list_node_prev_is_valid(b));
```
bretambrose: This seems odd to just assert b->prev valid

include/aws/common/linked_hash_table.h
JonathanHenson: that was largely when I was allowing mutation. and I wanted to make clear not to screw around with the linked list while insertions etc... were happening. Can remove the thread safety part.

include/aws/common/linked_list.h
JonathanHenson: that's left over from previous iteration, willupdate.

include/aws/common/linked_list.h
```diff
@@ -212,6 +212,34 @@ AWS_STATIC_IMPL void aws_linked_list_insert_after(
...
 }
 
+/**
+ * Swaps the order two nodes in the linked list.
+ */
+AWS_STATIC_IMPL void aws_linked_list_swap_nodes(struct aws_linked_list_node *a, struct aws_linked_list_node *b) {
+    AWS_PRECONDITION(aws_linked_list_node_next_is_valid(a));
+    AWS_PRECONDITION(aws_linked_list_node_next_is_valid(b));
+    AWS_PRECONDITION(a != NULL);
+    AWS_PRECONDITION(b != NULL);
```
graebm: no need to check against null. the is_valid() checks already do that

include/aws/common/linked_list.h
```diff
@@ -212,6 +212,34 @@ AWS_STATIC_IMPL void aws_linked_list_insert_after(
...
+    b->prev->next = a;
+    b->prev = a_prev_cpy;
+    b->next->prev = a;
+    b->next = a_prev_next;
+
+    AWS_POSTCONDITION(aws_linked_list_node_next_is_valid(a));
+    AWS_POSTCONDITION(aws_linked_list_node_prev_is_valid(b));
+    AWS_POSTCONDITION(aws_linked_list_node_next_is_valid(b));
+    AWS_POSTCONDITION(b->next == a);
+    AWS_POSTCONDITION(a->prev == b);
```
graebm: `b->next == a` is only true if A and B were right next to each other when you swapped them.

include/aws/common/linked_list.h
```diff
@@ -212,6 +212,34 @@ AWS_STATIC_IMPL void aws_linked_list_insert_after(
...
+ */
+AWS_STATIC_IMPL void aws_linked_list_swap_nodes(struct aws_linked_list_node *a, struct aws_linked_list_node *b) {
+    AWS_PRECONDITION(aws_linked_list_node_next_is_valid(a));
+    AWS_PRECONDITION(aws_linked_list_node_next_is_valid(b));
+    AWS_PRECONDITION(a != NULL);
+    AWS_PRECONDITION(b != NULL);
+
+    struct aws_linked_list_node *a_prev_cpy = a->prev;
+    struct aws_linked_list_node *a_prev_next = a->next;
+
```
graebm: Function fails to ever update the value of `a->prev`

Can I be an overstepping douchebag and suggest this implementation instead?

```
// swap A and B
tmp = *a
*a = *b
*b = tmp

// fix up adjacent nodes
a->prev->next = a
a->next->prev = a

b->prev->next = b
b->next->prev = b
```

graebm: I'm a dummy. This doesn't work if A and B are adjacent.

graebm: ok, here's my new proposal. different branches for adjacent swap vs non-adjacent swap
```
if (a == b)
    return

// test if A and B are adjacent
first_adjacent = NULL
if (a->next == b)
    first_adjacent = a
else if (b->next == a)
    first_adjacent = b

if (first_adjacent)
    // number nodes by the order we WANT them to be
    node1 = first_adjacent->prev
    node2 = first_adjacent->next
    node3 = first_adjacent
    node4 = first_adjacent->next->next

    // now put them in that order
    node1->next = node2

    node2->prev = node1
    node2->next = node3

    node3->prev = node2
    node3->next = node4

    node4->prev = node3

else
    // swap non-adjacent A and B
    tmp = *a
    *a = *b
    *b = tmp

    // fix up nodes adjacent to A and B
    a->prev->next = a
    a->next->prev = a

    b->prev->next = b
    b->next->prev = b
```

PR #464, opened 2019-07-25
Update ring buffer proofs

nchong-at-aws:
* Fix proofs for aws_ring_buffer_buf_belongs_to_pool_harness() and
aws_ring_buffer_release()
* More precise aws_ring_buffer_is_valid() test
* Head/tail properties for acquire


nchong-at-aws:
@danielsn @tautschnig

source/ring_buffer.c
```diff
@@ -214,7 +214,16 @@ int aws_ring_buffer_acquire_up_to(
 }
 
 static inline bool s_buf_belongs_to_pool(const struct aws_ring_buffer *ring_buffer, const struct aws_byte_buf *buf) {
-    return buf->buffer >= ring_buffer->allocation && buf->buffer + buf->capacity <= ring_buffer->allocation_end;
+/* comparison of pointers to different objects is undefined (C11 6.5.8) */
+#ifdef CBMC
+#    pragma CPROVER check push
+#    pragma CPROVER check disable "pointer"
```
angelhof: Is there any way to ensure that all CBMC proofs that call this function call it with pointers to the same object, instead of disabling pointer checks?

tautschnig: 1) I would expect a fairly extensive comment as to why we are disabling pointer checks.
2) It seems @angelhof is rather calling for even stricter pointer checks? I believe we could use `__CPROVER_POINTER_OBJECT(foo) == __CPROVER_POINTER_OBJECT(bar)` to do the suggested comparison.

nchong-at-aws: We can't guarantee `__CPROVER_POINTER_OBJECT(buf->buffer) == __CPROVER_POINTER_OBJECT(ring_buffer->allocation)` because external code can call this function via `aws_ring_buffer_buf_belongs_to_pool`. We could make this a requirement at the API level, but then I think this function will always return true.

It is possible we could rewrite this function using pointer-to-integer casts. Although the guarantees for integers converted from pointers are implementation-defined.

Or we can scratch this up as being okay in practice and disable the pointer-check.

tautschnig: So maybe do `if(__CPROVER_POINTER_OBJECT(buf->buffer) == __CPROVER_POINTER_OBJECT(ring_buffer->allocation)) { ...` instead of disabling the pointer checks completely?

include/aws/common/ring_buffer.h
```diff
@@ -74,6 +65,30 @@ AWS_STATIC_IMPL bool aws_ring_buffer_check_atomic_ptr(
...
+ * Evaluates the set of properties that define the shape of all valid aws_ring_buffer structures.
+ * It is also a cheap check, in the sense it run in constant time (i.e., no loops or recursion).
+ */
+AWS_STATIC_IMPL bool aws_ring_buffer_is_valid(const struct aws_ring_buffer *ring_buf) {
+    uint8_t *head = (uint8_t *)aws_atomic_load_ptr(&ring_buf->head);
+    uint8_t *tail = (uint8_t *)aws_atomic_load_ptr(&ring_buf->tail);
+    bool head_in_range = aws_ring_buffer_check_atomic_ptr(ring_buf, head);
+    bool tail_in_range = aws_ring_buffer_check_atomic_ptr(ring_buf, tail);
+    /* if head points-to the first element of the buffer then tail must too */
+    bool valid_head_tail = head != ring_buf->allocation || tail == ring_buf->allocation;
```
danielsn: Brackets here please.

.cbmc-batch/source/make_common_data_structures.c
```diff
@@ -38,13 +38,43 @@ void ensure_ring_buffer_has_allocated_members(struct aws_ring_buffer *ring_buf,
...
+    size_t space = hi - lo;
+    size_t pos;
+    __CPROVER_assume(pos < space);
+    buf->buffer = lo + pos;
+    size_t max_capacity = hi - buf->buffer;
+    assert(0 < max_capacity);
+    __CPROVER_assume(0 < buf->capacity && buf->capacity <= max_capacity);
+}
+
+void ensure_byte_buf_has_allocated_buffer_member_in_ring_buf(
```
danielsn: Function comment.

.cbmc-batch/source/make_common_data_structures.c
```diff
@@ -38,13 +38,43 @@ void ensure_ring_buffer_has_allocated_members(struct aws_ring_buffer *ring_buf,
...
-    __CPROVER_assume(position_head < size);
-    __CPROVER_assume(position_tail < size);
+    __CPROVER_assume(position_head <= size);
+    __CPROVER_assume(position_tail <= size);
     aws_atomic_store_ptr(&ring_buf->head, (ring_buf->allocation + position_head));
     aws_atomic_store_ptr(&ring_buf->tail, (ring_buf->allocation + position_tail));
     ring_buf->allocation_end = ring_buf->allocation + size;
 }
 
+void ensure_byte_buf_has_allocated_buffer_member_in_range(struct aws_byte_buf *buf, uint8_t *lo, uint8_t *hi) {
```
danielsn: Function comment

.cbmc-batch/source/make_common_data_structures.c
```diff
@@ -38,13 +38,43 @@ void ensure_ring_buffer_has_allocated_members(struct aws_ring_buffer *ring_buf,
...
+    assert(lo < hi);
+    size_t space = hi - lo;
```
danielsn: Brackets

.cbmc-batch/jobs/aws_ring_buffer_buf_belongs_to_pool/aws_ring_buffer_buf_belongs_to_pool_harness.c
```diff
@@ -26,14 +26,26 @@ void aws_ring_buffer_buf_belongs_to_pool_harness() {
 
     /* assumptions */
     ensure_ring_buffer_has_allocated_members(&ring_buf, ring_buf_size);
+    bool is_member = nondet_bool(); /* nondet assignment required to force true/false */
```
danielsn: I don't understand this comment.  Shouldn't it nondet to bool already?

nchong-at-aws: Yes, I thought so too. But, I get a counterexample without the nondet assignment. In the following, CBMC will assign a truthy value (like 2) to x:

```
#include <stdbool.h>
int main() {
   bool x;
   assert(x == true || x == false); //< fails
}
```

tautschnig: Since C doesn't have single-bit types, `bool` (which typically is defined to be `_Bool`) is at least eight bits (on common platforms), and CBMC will happily pick a value that is neither `0` (false) nor `1` (true). `nondet_bool` in CBMC is special (see https://github.com/diffblue/cbmc/blob/306697d69605bb9b3c7f2e60a9a71cc9dff65419/src/goto-programs/builtin_functions.cpp#L875) in that it forces the return value to be 0 or 1.

source/ring_buffer.c
nchong-at-aws: Ok, now we short-circuit if the pointer comparison is undefined due to different pointer objects.

nchong-at-aws:
Failing check is downstream aws-c-http build. Looks unrelated.

.cbmc-batch/include/proof_helpers/ring_buffer_abstract_states.h
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Ring buffer is empty
+ */
+bool is_empty_state(struct aws_ring_buffer *ring_buf) {
+    uint8_t *head = aws_atomic_load_ptr(&ring_buf->head);
+    uint8_t *tail = aws_atomic_load_ptr(&ring_buf->tail);
+    return tail == head;
+}
+
+/**
```
danielsn: Could these be promoted to the mainline code?

include/aws/common/ring_buffer.h
```diff
@@ -74,6 +65,30 @@ AWS_STATIC_IMPL bool aws_ring_buffer_check_atomic_ptr(
...
+ * It is also a cheap check, in the sense it run in constant time (i.e., no loops or recursion).
+ */
+AWS_STATIC_IMPL bool aws_ring_buffer_is_valid(const struct aws_ring_buffer *ring_buf) {
+    uint8_t *head = (uint8_t *)aws_atomic_load_ptr(&ring_buf->head);
+    uint8_t *tail = (uint8_t *)aws_atomic_load_ptr(&ring_buf->tail);
+    bool head_in_range = aws_ring_buffer_check_atomic_ptr(ring_buf, head);
+    bool tail_in_range = aws_ring_buffer_check_atomic_ptr(ring_buf, tail);
+    /* if head points-to the first element of the buffer then tail must too */
+    bool valid_head_tail = (head != ring_buf->allocation) || (tail == ring_buf->allocation);
+    return ring_buf && AWS_MEM_IS_READABLE(ring_buf->allocation, ring_buf->allocation_end - ring_buf->allocation) &&
```
danielsn: Can this subtraction underflow?

nchong-at-aws: I don't believe so since we could prove the invariant `allocation < allocation_size`. These variables are set once by a successful ring_buffer init() and then never modified. There is a precondition (that I think you added?) that the size of the requested size of the ring_buffer is non-zero and `allocation_end` is set to `allocation+size`.

Issue #471, opened 2019-07-30
Make the presence of assertions testable


danielsn:
This was a request from https://github.com/awslabs/aws-c-common/pull/469

PR #472, opened 2019-07-30
Test for assertions

danielsn:
*Issue #, if available:*
https://github.com/awslabs/aws-c-common/issues/471

*Description of changes:*
Proposal of how to address this issue

### DIFF
--- a/include/aws/common/assert.h
+++ b/include/aws/common/assert.h
@@ -23,8 +23,7 @@
 AWS_EXTERN_C_BEGIN
 
 AWS_COMMON_API
-AWS_DECLSPEC_NORETURN
-void aws_fatal_assert(const char *cond_str, const char *file, int line) AWS_ATTRIBUTE_NORETURN;
+void aws_fatal_assert(const char *cond_str, const char *file, int line);
 
 /**
  * Print a backtrace from either the current stack, or (if provided) the current exception/signal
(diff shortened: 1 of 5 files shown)


danielsn:
@ColdenCullen @justinboswell This is still a draft, but how's the general idea look?

source/assert.c
```diff
@@ -18,11 +18,24 @@
...
 
 void aws_fatal_assert(const char *cond_str, const char *file, int line) {
     aws_debug_break();
-    fprintf(stderr, "Fatal error condition occurred in %s:%d: %s\nExiting Application\n", file, line, cond_str);
-    aws_backtrace_print(stderr, NULL);
-    abort();
+
+    if (at_assert) {
+        at_assert(cond_str, file, line);
+    } else {
```
danielsn: I could also just make this a default handler, and make it always call whatever is set as the handler.

include/aws/common/assert.h
```diff
@@ -33,6 +32,21 @@ void aws_fatal_assert(const char *cond_str, const char *file, int line) AWS_ATTR
 AWS_COMMON_API
 void aws_backtrace_print(FILE *fp, void *call_site_data);
 
+typedef void (*assert_handler_fn_t)(const char *cond_str, const char *file, int line);
```
ColdenCullen: 2 things here:
* prefix the name with aws_
* Make the typedef not a pointer type, and put the pointer on the parameter.

include/aws/common/assert.h
```diff
@@ -33,6 +32,21 @@ void aws_fatal_assert(const char *cond_str, const char *file, int line) AWS_ATTR
...
+ * This is useful in testing, where you might wish to test that a given assert has triggered.
+ * It may also be useful in cases where you may want to add addional debug information to the log.
+ * Note that this will NOT call abort(); you are expected to do this yourself if you require.
+ * This is not guaraneed to be thread
+ *
+ * Returns the existing handler: this is useful in case you want to temporarily swap handlers, and then restore the old
+ * one.
+ */
+AWS_COMMON_API
+assert_handler_fn_t register_assert_handler(assert_handler_fn_t the_fn);
```
ColdenCullen: prefix with aws_ pls

source/assert.c
ColdenCullen: That will reduce branches, maybe worth doing? I would:
* Make the default assert handler a publicly exported function, in case someone else wants to call it.
* Finally add that `aws_common_library_init()` function we've been meaning to add in forever, and install it there.

PR #475, opened 2019-08-05
Extend the postconditions of aws_array_list_front with bytes_eq check

angelhof:
This PR introduces a new assertion predicate that checks whether the bytes in two byte arrays match. It extends the postcondition of `aws_array_list_front` with that check.

### DIFF
--- a/include/aws/common/array_list.inl
+++ b/include/aws/common/array_list.inl
@@ -147,6 +147,7 @@ int aws_array_list_front(const struct aws_array_list *AWS_RESTRICT list, void *v
         "Input pointer [val] must point writable memory of [list->item_size] bytes.");
     if (aws_array_list_length(list) > 0) {
         memcpy(val, list->data, list->item_size);
+        AWS_POSTCONDITION(AWS_BYTES_EQ(val, list->data, list->item_size));
         AWS_POSTCONDITION(aws_array_list_is_valid(list));
         return AWS_OP_SUCCESS;
     }
--- a/include/aws/common/predicates.h
+++ b/include/aws/common/predicates.h
@@ -1,0 +1,33 @@
+#ifndef AWS_COMMON_PREDICATES_H
+#define AWS_COMMON_PREDICATES_H
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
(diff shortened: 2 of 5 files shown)


include/aws/common/array_list.inl
```diff
@@ -147,6 +147,7 @@ int aws_array_list_front(const struct aws_array_list *AWS_RESTRICT list, void *v
         "Input pointer [val] must point writable memory of [list->item_size] bytes.");
     if (aws_array_list_length(list) > 0) {
         memcpy(val, list->data, list->item_size);
+        AWS_POSTCONDITION(AWS_BYTES_EQ(val, list->data, list->item_size));
```
ColdenCullen: Is this really necessary? This seems like it would add O(n) cost to each call to this function in debug builds. Are we really concerned that memcpy could fail?

include/aws/common/predicates.h
```diff
@@ -0,0 +1,33 @@
...
+        __CPROVER_forall {                                                                                             \
+            int i;                                                                                                     \
+            (i >= 0 && i < len) ==> ((const uint8_t *)&arr1)[i] == ((const uint8_t *)&arr2)[i]                         \
+        }
+/* clang-format on */
+#else
+#    define AWS_BYTES_EQ(arr1, arr2, len) (memcmp(arr1, arr2, len) == 0)
+#endif
+
+#endif /* AWS_PREDICATES_MACROS_H */
```
ColdenCullen: Comment doesn't match include guard

angelhof: Thanks

include/aws/common/array_list.inl
angelhof: Thanks for the comment! 

The reason why I added this, is to extend the function contract (what we know about a function by just reading its pre and postconditions). It is true that the `memcmp` would have to be executed on debug builds, but we are already doing a `memcpy` so, it shouldn't be a big issue. 

If the cost of a `memcmp` there is prohibitive in debug builds, then I could wrap it in `AWS_DEEP_CHECKS` macro, to only be enabled then.

In any case, I have made a PR in CBMC (https://github.com/diffblue/cbmc/pull/4976) that will allow us to remove all `AWS_POSTCONDITION`s from the code, and just have them as contracts -- which means that they will not be executed in debug builds.

ColdenCullen: I would prefer it go in a DEEP_CHECKS check if you don't mind.

angelhof: I wrapped it in a deep check.

PR #477, opened 2019-08-08
Add a aws_add_size_checked_varargs function

danielsn:
*Description of changes:*
Sometimes, we just want to add more than 2 things.


source/math.c
```diff
@@ -0,0 +1,34 @@
...
+
+#include <aws/common/math.h>
+#include <stdarg.h>
+
+AWS_COMMON_API int aws_add_size_checked_varargs(size_t num, size_t *r, ...) {
+    va_list argp;
+    va_start(argp, r);
+
+    size_t accum = 0;
+    for (unsigned i = 0; i < num; ++i) {
```
ColdenCullen: looks like clang-tidy wants `i` to be a `size_t`, which makes sense to me

PR #480, opened 2019-08-14
moved byte buf functions to the .c where they belong

danielsn:
*Description of changes:*
Right now, a lot of stuff is "static inline" in the header files.  This makes the header files hard to read, gives little performance benefit in most cases, and makes my CBMC proof life much harder.  This is the first of several refactorings to move the code the .c files where it belongs.

### DIFF
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -66,6 +66,12 @@ struct aws_byte_cursor {
 #define AWS_BYTE_BUF_PRI(B) ((int)(B).len < 0 ? 0 : (int)(B).len), (const char *)(B).buffer
 
 /**
+ * Helper Macro for inititilizing a byte cursor from a string literal
+ */
+#define AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL(literal)                                                                 \
+    { .ptr = (uint8_t *)(const char *)(literal), .len = sizeof(literal) - 1 }
+
+/**
  * Signature for function argument to trim APIs
  */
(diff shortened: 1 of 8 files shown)


danielsn:
I’m seeing a clang-tidy error in one of the tests.  

```
/codebuild/output/src425878266/src/github.com/awslabs/aws-c-common/tests/device_random_test.c:135:24: error: 2nd function call argument is an uninitialized value [clang-analyzer-core.CallAndMessage,-warnings-as-errors] 
        ASSERT_SUCCESS(s_distribution_tester_put(&tester, array[i])); 
```
It looks like the problem is that array[i] is set by the device random but clang-tidy doesn’t know that

include/aws/common/byte_buf.h
```diff
@@ -464,146 +470,28 @@ int aws_byte_cursor_compare_lookup(
...
-     * an arithmetic right shift operation on x86.
-     */
-    combined_mask = combined_mask * UINTPTR_MAX;
-
-    return combined_mask;
-}
-#ifdef CBMC
-#    pragma CPROVER check pop
-#endif
+AWS_COMMON_API size_t aws_nospec_mask(size_t index, size_t bound);
```
justinboswell: Do we actually need to export this? Pretty sure we just use this internally.

danielsn: Move it to a private header, or leave it in the public header but without the ```AWS_COMMON_API```?

danielsn: As per offline discussion, move to private header.

PR #486, opened 2019-08-20
added aws_string_clone_or_reuse() function, and test for it

danielsn:
*Description of changes:*
Add `aws_string_clone_or_reuse()` functionality, and a test for it.

### DIFF
--- a/source/string.c
+++ b/source/string.c
@@ -243,3 +243,18 @@ struct aws_byte_cursor aws_byte_cursor_from_string(const struct aws_string *src)
     AWS_PRECONDITION(aws_string_is_valid(src));
     return aws_byte_cursor_from_array(aws_string_bytes(src), src->len);
 }
+
+struct aws_string *aws_string_clone_or_reuse(struct aws_allocator *allocator, const struct aws_string *str) {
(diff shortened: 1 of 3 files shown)


source/string.c
```diff
@@ -243,3 +243,18 @@ struct aws_byte_cursor aws_byte_cursor_from_string(const struct aws_string *src)
...
 }
+
+struct aws_string *aws_string_clone_or_reuse(struct aws_allocator *allocator, const struct aws_string *str) {
+    AWS_PRECONDITION(allocator);
+    AWS_PRECONDITION(aws_string_is_valid(str));
+
+    if (str->allocator == NULL) {
+        /* Since the string cannot be deallocated, we assume that it will remain valid for the lifetime of the
+         * application */
+        AWS_POSTCONDITION(aws_string_is_valid(str));
```
justinboswell: Is the post condition necessary? No mutation has occurred?

danielsn: It's not strictly necessary.  But its useful documentation (tho I guess you could argue the `const` qualifier on the input is all the documentation you need).

PR #487, opened 2019-08-20
Implement common library-init functions

ColdenCullen:

### DIFF
--- a/include/aws/common/common.h
+++ b/include/aws/common/common.h
@@ -39,7 +39,7 @@ AWS_EXTERN_C_BEGIN
  * Must be called before using any functionality in aws-c-common.
  */
 AWS_COMMON_API
-void aws_common_library_init(void);
+void aws_common_library_init(struct aws_allocator *allocator);
 
 /**
  * Shuts down the internal datastructures used by aws-c-common.
--- a/include/aws/common/common.h
+++ b/include/aws/common/common.h
@@ -32,4 +32,21 @@
 #include <stdlib.h> /* for abort() */
 #include <string.h>
 
+AWS_EXTERN_C_BEGIN
+
+/**
+ * Initializes internal datastructures used by aws-c-common.
+ * Must be called before using any functionality in aws-c-common.
+ */
+AWS_COMMON_API
+void aws_common_library_init(void);
+
+/**
+ * Shuts down the internal datastructures used by aws-c-common.
+ */
+AWS_COMMON_API
+void aws_common_library_clean_up(void);
+
+AWS_EXTERN_C_END
+
 #endif /* AWS_COMMON_COMMON_H */
--- a/include/aws/common/common.h
+++ b/include/aws/common/common.h
@@ -47,6 +47,9 @@ void aws_common_library_init(struct aws_allocator *allocator);
 AWS_COMMON_API
 void aws_common_library_clean_up(void);
 
+AWS_COMMON_API
+void aws_common_fatal_assert_library_initialized(void);
+
 AWS_EXTERN_C_END
 
 #endif /* AWS_COMMON_COMMON_H */
--- a/source/common.c
+++ b/source/common.c
@@ -549,9 +549,8 @@ void aws_common_library_clean_up(void) {
 
 void aws_common_fatal_assert_library_initialized(void) {
     if (!s_common_library_initialized) {
-        AWS_LOGF_FATAL(
-            AWS_LS_COMMON_GENERAL,
-            "aws_common_library_init() must be called before using any functionality in aws-c-common.");
+        fprintf(
+            stderr, "%s", "aws_common_library_init() must be called before using any functionality in aws-c-common.");
 
         AWS_FATAL_ASSERT(s_common_library_initialized);
     }
--- a/source/common.c
+++ b/source/common.c
@@ -14,8 +14,8 @@
  */
 
 #include <aws/common/common.h>
-#include <aws/common/math.h>
 #include <aws/common/logging.h>
+#include <aws/common/math.h>
 
 #include <stdarg.h>
 #include <stdlib.h>
--- a/source/common.c
+++ b/source/common.c
@@ -529,7 +529,9 @@ static struct aws_log_subject_info_list s_common_log_subject_list = {
 
 static bool s_common_library_initialized = false;
 
-void aws_common_library_init(void) {
+void aws_common_library_init(struct aws_allocator *allocator) {
+    (void)allocator;
+
     if (!s_common_library_initialized) {
         s_common_library_initialized = true;
         aws_register_error_info(&s_list);
--- a/source/common.c
+++ b/source/common.c
@@ -547,6 +547,16 @@ void aws_common_library_clean_up(void) {
     }
 }
 
+void aws_common_fatal_assert_library_initialized(void) {
+    if (!s_common_library_initialized) {
+        AWS_LOGF_FATAL(
+            AWS_LS_COMMON_GENERAL,
+            "aws_common_library_init() must be called before using any functionality in aws-c-common.");
+
+        AWS_FATAL_ASSERT(s_common_library_initialized);
+    }
+}
+
 #ifdef _MSC_VER
 #    pragma warning(pop)
 #endif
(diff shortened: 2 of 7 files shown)


source/common.c
```diff
@@ -352,6 +353,198 @@ void aws_secure_zero(void *pBuf, size_t bufsize) {
...
+        aws_register_error_info(&s_list);
+        aws_register_log_subject_info_list(&s_common_log_subject_list);
+    }
+}
+
+void aws_common_library_clean_up(void) {
+    if (s_common_library_initialized) {
+        s_common_library_initialized = false;
+        aws_unregister_error_info(&s_list);
+        aws_unregister_log_subject_info_list(&s_common_log_subject_list);
```
graebm: Just curious, why bother unregistering these? We never did so before

ColdenCullen: Step 1: User hand-loads DLL
Step 2: User inits library. aws-c-common now has pointers to strings in DLL static memory
Step 3: User hand-unloads DLL
Step 4: User attempts to print error info. BANG DING OW

include/aws/common/common.h
```diff
@@ -32,4 +32,24 @@
...
+void aws_common_library_init(struct aws_allocator *allocator);
+
+/**
+ * Shuts down the internal datastructures used by aws-c-common.
+ */
+AWS_COMMON_API
+void aws_common_library_clean_up(void);
+
+AWS_COMMON_API
+void aws_common_fatal_assert_library_initialized(void);
```
graebm: Should probably call this from somewhere. Maybe from `get_error_by_code()`?

source/common.c
```diff
@@ -352,6 +353,210 @@ void aws_secure_zero(void *pBuf, size_t bufsize) {
...
+    }
+}
+
+void aws_common_fatal_assert_library_initialized(void) {
+    if (!s_common_library_initialized) {
+        AWS_LOGF_FATAL(
+            AWS_LS_COMMON_GENERAL,
```
graebm: Can't log if the library was never initialized 😭
`fprintf(stderr,` ?

source/common.c
```diff
@@ -352,6 +353,210 @@ void aws_secure_zero(void *pBuf, size_t bufsize) {
...
+
+    if (!s_common_library_initialized) {
+        s_common_library_initialized = true;
+        aws_register_error_info(&s_list);
+        aws_register_log_subject_info_list(&s_common_log_subject_list);
+    }
+}
+
+void aws_common_library_clean_up(void) {
+    if (s_common_library_initialized) {
```
danielsn: Should it be an error (or at least a warning) to call clean_up on a library that was never init?

source/error.c
```diff
@@ -150,9 +150,8 @@ void aws_register_error_info(const struct aws_error_info_list *error_info) {
     AWS_ASSERT(error_info->error_list);
     AWS_ASSERT(error_info->count);
 
-    int min_range = error_info->error_list[0].error_code;
-
-    int slot_index = min_range >> SLOT_DIV_SHIFT;
+    const int min_range = error_info->error_list[0].error_code;
+    const int slot_index = min_range >> SLOT_DIV_SHIFT;
```
danielsn: Any reasonable compiler will do strength reduction if you do division by a constant power of 2. Would be easier to read.

source/common.c
```diff
@@ -352,6 +353,210 @@ void aws_secure_zero(void *pBuf, size_t bufsize) {
 #endif     // #else not windows
 }
 
+#define AWS_DEFINE_ERROR_INFO_COMMON(C, ES) [(C)-0x0000] = AWS_DEFINE_ERROR_INFO(C, ES, "aws-c-common")
+/* clang-format off */
+static struct aws_error_info errors[] = {
```
danielsn: Should there be a way to internationalize this?

source/error.c
```diff
@@ -150,9 +150,8 @@ void aws_register_error_info(const struct aws_error_info_list *error_info) {
...
 
     AWS_ASSERT(slot_index < AWS_MAX_ERROR_SLOTS && slot_index >= 0);
```
danielsn: If this is in prod mode, we'll return random data here. Should this either be a fatal assert, or a check that slot_index is in range, followed by returning a "this was a bad slot" error message?

source/error.c
```diff
@@ -182,172 +181,24 @@ void aws_register_error_info(const struct aws_error_info_list *error_info) {
...
-        aws_register_error_info(&s_list);
+void aws_unregister_error_info(const struct aws_error_info_list *error_info) {
+    AWS_ASSERT(error_info);
+    AWS_ASSERT(error_info->error_list);
+    AWS_ASSERT(error_info->count);
+
+    const int min_range = error_info->error_list[0].error_code;
+    const int slot_index = min_range >> SLOT_DIV_SHIFT;
+
+    AWS_ASSERT(slot_index < AWS_MAX_ERROR_SLOTS && slot_index >= 0);
```
danielsn: If `AWS_FATAL_ASSERT()` could take an error message string, these could both be combined into a single `AWS_FATAL_ASSERT`

source/logging.c
```diff
@@ -335,9 +333,8 @@ void aws_register_log_subject_info_list(struct aws_log_subject_info_list *log_su
     AWS_ASSERT(log_subject_list->subject_list);
     AWS_ASSERT(log_subject_list->count);
 
-    uint32_t min_range = log_subject_list->subject_list[0].subject_id;
-
-    uint32_t slot_index = min_range >> AWS_LOG_SUBJECT_BIT_SPACE;
+    const uint32_t min_range = log_subject_list->subject_list[0].subject_id;
+    const uint32_t slot_index = min_range >> AWS_LOG_SUBJECT_BIT_SPACE;
 
     AWS_ASSERT(slot_index < AWS_MAX_LOG_SUBJECT_SLOTS);
```
danielsn: Same as before: should this be `AWS_FATAL_ASSERT`?

source/logging.c
```diff
@@ -335,9 +333,8 @@ void aws_register_log_subject_info_list(struct aws_log_subject_info_list *log_su
     AWS_ASSERT(log_subject_list->subject_list);
     AWS_ASSERT(log_subject_list->count);
 
-    uint32_t min_range = log_subject_list->subject_list[0].subject_id;
-
-    uint32_t slot_index = min_range >> AWS_LOG_SUBJECT_BIT_SPACE;
+    const uint32_t min_range = log_subject_list->subject_list[0].subject_id;
+    const uint32_t slot_index = min_range >> AWS_LOG_SUBJECT_BIT_SPACE;
```
danielsn: Same as before - should this be division?

source/logging.c
```diff
@@ -351,22 +348,27 @@ void aws_register_log_subject_info_list(struct aws_log_subject_info_list *log_su
...
-        "task-scheduler",
-        "Subject for task scheduler or task specific logging."),
-};
+void aws_unregister_log_subject_info_list(struct aws_log_subject_info_list *log_subject_list) {
+    /*
+     * We're not so worried about these asserts being removed in an NDEBUG build
+     * - we'll either segfault immediately (for the first two) or for the count
+     * assert, the registration will be ineffective.
+     */
+    AWS_ASSERT(log_subject_list);
```
danielsn: If you expect to segfault immediatly, why not make it a `FATAL_ASSERT` anyway? Its not like this function is on the critical path of a tight loop.

source/common.c
ColdenCullen: No, the goal is for these functions to be completely idempotent.

source/common.c
ColdenCullen: Probably, but I believe that to be out of scope for this PR, and likely for GA.

source/error.c
ColdenCullen: I'm not sure I understand, what would be your proposed implementation?

source/error.c
ColdenCullen: I can definitely make it a fatal assert. I think this code predates that macro.

ColdenCullen: WAit no, it's fatal asserted right below. I can remove this check.

source/error.c
ColdenCullen: We're logging right below if this condition is bad, so I'm just gonna nuke this.

PR #491, opened 2019-08-23
CI: Support up and downstream dependencies

ColdenCullen:
With this PR, each repo may define it's own upstream dependencies, and downstream projects to test (when running a downstream build).


.codebuild.json
```diff
@@ -0,0 +1,12 @@
+{
```
justinboswell: My only concern with this file is the name, who's to say codebuild doesn't eventually have a .codebuild.json default file. Maybe just name it something different?

codebuild/builder.py
```diff
@@ -490,24 +497,150 @@ def run_build(build_spec, is_dryrun):
...
+        _log_command(*command)
         if not is_dryrun:
-            subprocess.check_call(_flatten_command(command), stdout=sys.stdout, stderr=sys.stderr)
+            subprocess.check_call(_flatten_command(*command), stdout=sys.stdout, stderr=sys.stderr)
+
+    # Helper to run makedirs regardless of dry run status
+    def _mkdir(directory):
+        if not is_dryrun:
+            os.makedirs(directory, exist_ok=True)
+        _log_command("mkdir", "-p", directory)
```
justinboswell: Can we log before attempting the makedirs, so if an exception is thrown, we have some info?

codebuild/builder.py
```diff
@@ -490,24 +497,150 @@ def run_build(build_spec, is_dryrun):
...
+    def _cd(directory):
+        if is_dryrun:
+            global cwd
+            if os.path.isabs(directory) or directory.startswith('$'):
+                cwd = directory
+            else:
+                cwd = os.path.join(cwd, directory)
+        else:
+            os.chdir(directory)
+        _log_command("cd", directory)
```
justinboswell: Same here, log first.

codebuild/builder.py
```diff
@@ -516,23 +649,27 @@ def _run_command(command):
         build_dir = tempfile.mkdtemp()
     _log_command(['mkdir', build_dir])
```
justinboswell: Log first

codebuild/builder.py
ColdenCullen: Sure

codebuild/builder.py
ColdenCullen: Yup!

codebuild/builder.py
ColdenCullen: Can't here, don't have the directory name until mkdtemp() returns.

PR #506, opened 2019-09-13
Handle static/lib side-by-side builds.

JonathanHenson:

### DIFF
--- a/cmake/aws-c-common-config.cmake
+++ b/cmake/aws-c-common-config.cmake
@@ -1,5 +1,9 @@
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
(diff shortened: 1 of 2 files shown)


cmake/aws-c-common-config.cmake
```diff
@@ -1,5 +1,9 @@
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
-include(${CMAKE_CURRENT_LIST_DIR}/@CMAKE_PROJECT_NAME@-targets.cmake)
+if (BUILD_SHARED_LIBS)
```
singku: When building customer executables, it doesn't make sense to specify `BUILD_SHARED_LIBS`, hence we probably won't get `BUILD_SHARED_LIBS` from CMake command line. Then CMake will try to find a static targets by default. But if we specify BUILD_SHARED_LIBS when building this library, then only shared targets CMake file will be generated. I think this is a problem for customer application.

PR #507, opened 2019-09-18
Fixed support for gcc 4.1.2.

JonathanHenson:


cmake/AwsCFlags.cmake
```diff
@@ -16,6 +16,7 @@ include(CheckIncludeFile)
 include(CMakeParseArguments) # needed for CMake v3.4 and lower
 
 option(AWS_ENABLE_LTO "Enables LTO on libraries. Ensure this is set on all consumed targets, or linking will fail" OFF)
+option(LEGACY_COMPILER_SUPPORT "Do not use this unless you're trying to build this library on legacy compilers that you shouldn't be using in the first place" OFF)
```
bretambrose: Maybe add a more specific comment here about the actual compiler and version this fixes up: "this enables compilation for <blah blah> but versions prior to <blah blah> remain unsupported."

JonathanHenson: fair enough

include/aws/common/assert.h
```diff
@@ -45,8 +45,10 @@ AWS_EXTERN_C_END
             bool _result = (cond);                                                                                     \
             __builtin_assume(_result);                                                                                 \
         } while (false)
-#elif defined(__GNUC__)
+#elif defined(__GNUC__) && __GNUC__ >= 4 && __GNUC_MINOR__ >= 5
```
ColdenCullen: uuuuuh so gcc 8.1 for example doesn't get this branch?

JonathanHenson: Well crap! I’ll have to do another fix.

PR #513, opened 2019-10-01
Fixed long function name buffer overflow

justinboswell:

### DIFF
--- a/source/posix/system_info.c
+++ b/source/posix/system_info.c
@@ -149,6 +149,11 @@ int s_parse_symbol(const char *symbol, void *addr, struct aws_stack_frame_info *
     /* parse function */
     const char *function_start = strstr(addr_end, " ") + 1;
     const char *function_end = strstr(function_start, " ");
+    /* truncate function name if needed */
+    size_t function_len = function_end - function_start;
+    if (function_len >= (sizeof(frame->function) - 1)) {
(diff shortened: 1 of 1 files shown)


source/posix/system_info.c
```diff
@@ -109,6 +109,38 @@ struct aws_stack_frame_info {
     char function[128];
 };
 
+char* s_strip_shell_chars(char *path) {
```
graebm: add comment why we're doing this

justinboswell: Done!

source/posix/system_info.c
```diff
@@ -109,6 +109,41 @@ struct aws_stack_frame_info {
...
+            case '(':
+            case ')':
+            case '&':
+            case '#':
+            case '@':
+            case '!':
+            case '^':
+            case '=':
+            case '?':
+                *cur = '_';
```
ColdenCullen: Why not `if (!(isalpha(*cur) || isspace(*cur))) { *cur = '_'; }`

justinboswell: /path/to/do-the-thing.exe

No really, I did think about doing this as a whitelist, because that's the correct thing to do, but it's restrictive because it does limit what a customer can put in their binary name. I'm happy to do it either way.

PR #521, opened 2019-10-11
Stats

bretambrose:
* Base statistics gathering types
* Add package id and per-library enum stride and range calculation pattern


include/aws/common/error.h
```diff
@@ -192,7 +196,7 @@ enum aws_common_error {
     AWS_ERROR_SYS_CALL_FAILURE,
     AWS_ERROR_C_STRING_BUFFER_NOT_NULL_TERMINATED,
 
-    AWS_ERROR_END_COMMON_RANGE = 0x03FF
+    AWS_ERROR_END_COMMON_RANGE = (AWS_C_COMMON_PACKAGE_ID + 1) * AWS_ERROR_ENUM_STRIDE - 1
```
ColdenCullen: Everyone is going to have to implement this in every package. How about macros like `AWS_ERROR_START(package_id)` and `AWS_ERROR_END(package_id)`?

bretambrose: Seems reasonable

PR #526, opened 2019-10-15
Stat

bretambrose:
Attempt no. 2

* Base types for stat gathering
* Update cross-library enum range calculations


source/statistics.c
```diff
@@ -0,0 +1,25 @@
...
+ * permissions and limitations under the License.
+ */
+#include <aws/common/statistics.h>
+
+void aws_crt_statistics_handler_destroy(struct aws_crt_statistics_handler *handler) {
+    if (handler == NULL) {
+        return;
+    }
+
+    handler->vtable->cleanup(handler);
```
JonathanHenson: This isn't how the pattern is supposed to work, it should be the module that allocates destroys. The handler vtable should be destroy, and it should be responsible for freeing its own memory.

source/error.c
```diff
@@ -29,21 +29,15 @@ static void *s_global_error_context = NULL;
...
- * define those constants */
-#define AWS_ERROR_SLOT_SIZE 0x0400
-#define SLOT_DIV_SHIFT 0x0A
-#define SLOT_MASK 0x03FF
+ * 00000011 11111111 to find the index in that slot.
+ */
+#define SLOT_MASK (AWS_ERROR_ENUM_STRIDE - 1)
 
-static const int MAX_ERROR_CODE = AWS_ERROR_SLOT_SIZE * AWS_MAX_ERROR_SLOTS;
+static const int MAX_ERROR_CODE = AWS_ERROR_ENUM_STRIDE * AWS_PACKAGE_SLOTS;
```
JonathanHenson: why unsigned below but signed here? I don't care either way, I'd just push for consistency.

bretambrose: We're using signed integers for errors and unsigned integers for the other enums (log subject, statistics category).  I don't think it's necessarily inconsistent to have the bounds constants match the associated integer type rather than be the same across all.

JonathanHenson: Fair enough. Ship it.

include/aws/common/error.h
```diff
@@ -19,11 +19,17 @@
 #include <aws/common/assert.h>
 #include <aws/common/exports.h>
 #include <aws/common/macros.h>
+#include <aws/common/package.h>
 #include <aws/common/stdint.h>
 
 #define AWS_OP_SUCCESS (0)
 #define AWS_OP_ERR (-1)
 
+#define AWS_ERROR_ENUM_STRIDE_BITS 10
```
justinboswell: Maybe just a quick comment here so that the next reader doesn't have to math out what is happening here.

/* Create 1K ranges */ would even be enough

include/aws/common/logging.h
```diff
@@ -63,9 +63,10 @@ enum aws_log_level {
  */
 typedef uint32_t aws_log_subject_t;
 
-#define AWS_LOG_SUBJECT_BIT_SPACE 10
-#define AWS_LOG_SUBJECT_SPACE_SIZE (1 << AWS_LOG_SUBJECT_BIT_SPACE)
-#define AWS_LOG_SUBJECT_SPACE_MASK (AWS_LOG_SUBJECT_SPACE_SIZE - 1)
+#define AWS_LOG_SUBJECT_STRIDE_BITS 10
```
justinboswell: Same as with AWS_ERROR_ENUM_XXX

include/aws/common/package.h
```diff
@@ -0,0 +1,23 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#define AWS_PACKAGE_SLOTS 16
```
justinboswell: What does this do? I'll probably find out later in the PR, but it should probably have a comment.

include/aws/common/statistics.h
```diff
@@ -0,0 +1,149 @@
...
+#include <aws/common/common.h>
+#include <aws/common/package.h>
+
+#include <aws/common/stdint.h>
+
+struct aws_array_list;
+
+typedef uint32_t aws_crt_statistics_category_t;
+
+#define AWS_CRT_STATISTICS_CATEGORY_STRIDE_BITS 8
```
justinboswell: Same as AWS_ERROR_ENUM_XXX

include/aws/common/statistics.h
```diff
@@ -0,0 +1,149 @@
...
+/*
+ * Statistics intake function.  The array_list is a list of pointers to aws_crt_statistics_base "derived" (via
+ * pattern) objects.  The handler should iterate the list and downcast elements whose RTTI category it understands,
+ * while skipping those it does not understand.
+ */
+typedef void(aws_crt_statistics_handler_process_statistics_fn)(
+    struct aws_crt_statistics_handler *,
+    struct aws_crt_statistics_sample_interval *,
+    struct aws_array_list *,
+    void *);
```
justinboswell: It would be self-documenting to name these params

source/error.c
```diff
@@ -54,8 +48,8 @@ static const struct aws_error_info *get_error_by_code(int err) {
         return NULL;
     }
 
-    int slot_index = err >> SLOT_DIV_SHIFT;
-    int error_index = err & SLOT_MASK;
+    unsigned int slot_index = (unsigned int)err >> AWS_ERROR_ENUM_STRIDE_BITS;
```
justinboswell: ```uint32_t ```

PR #527, opened 2019-10-17
Imported ring buffer allocator from node CRT

justinboswell:


source/ring_buffer.c
```diff
@@ -268,3 +268,55 @@ bool aws_ring_buffer_buf_belongs_to_pool(const struct aws_ring_buffer *ring_buff
...
+
+int aws_ring_buffer_allocator_init(struct aws_allocator *allocator, struct aws_ring_buffer *ring_buffer) {
+    if (allocator == NULL || ring_buffer == NULL) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+
+    allocator->impl = ring_buffer;
+    allocator->mem_acquire = s_ring_buffer_mem_acquire;
+    allocator->mem_release = s_ring_buffer_mem_release;
+    allocator->mem_calloc = s_ring_buffer_mem_calloc;
```
JonathanHenson: I worry here that realloc will accidentally cause an out of order allocation/free since we didn’t implement it the fallback will be used. Maybe we should assert in realloc?

justinboswell: We don't define realloc, (it's null), so aws_mem_realloc() will assert for us.

JonathanHenson: https://github.com/awslabs/aws-c-common/blob/master/source/allocator.c#L184

justinboswell: NOPE, you were right. Fixed.

PR #532, opened 2019-10-28
Allow out-of-source builds for CBMC proofs

danielsn:
*Description of changes:*
What it says in the title.

### DIFF
--- a/.cbmc-batch/jobs/Makefile.common
+++ b/.cbmc-batch/jobs/Makefile.common
@@ -211,6 +211,12 @@ $(ENTRY_GOTO)7.goto: $(ENTRY_GOTO)6.goto
 $(ENTRY_GOTO).goto: $(ENTRY_GOTO)7.goto
 	cp $< $@
 
+# For now, CBMC batch expects the goto in the same folder.  I could update the
+# .yaml file for all 160 proofs, or I could copy it, and have one goto file
+# in the main directory.  
+$(ENTRY).goto: $(ENTRY_GOTO).goto
+	cp $< $@
+
 # Catch-all used for building goto-binaries of the individual
 # dependencies, which are then linked in the $(ENTRY_GOTO)0.goto rule above
 # DSN TODO Make this depend on the correct thing.
@@ -222,16 +228,16 @@ $(ENTRY_GOTO).goto: $(ENTRY_GOTO)7.goto
 	  $(INC) $(DEFINES) -o $@ \
 	  2>&1 | tee $(call LOG_FROM_GOTO, $@) ; exit $${PIPESTATUS[0]}
 
-goto: $(ENTRY_GOTO).goto
+goto: $(ENTRY).goto
 	echo $(DEPENDENT_GOTOS)
 
-cbmc.log: $(ENTRY_GOTO).goto
+cbmc.log: $(ENTRY).goto
 	cbmc $(CBMCFLAGS) --trace $< 2>&1 | tee $@
 
-property.xml: $(ENTRY_GOTO).goto
+property.xml: $(ENTRY).goto
 	cbmc $(CBMCFLAGS) --show-properties --xml-ui $< 2>&1 > $@
 
-coverage.xml: $(ENTRY_GOTO).goto
+coverage.xml: $(ENTRY).goto
 	cbmc $(filter-out --unwinding-assertions,$(CBMCFLAGS)) --cover location --xml-ui $< 2>&1 > $@
 
 cbmc: cbmc.log
@@ -243,7 +249,7 @@ coverage: coverage.xml
 report: cbmc.log property.xml coverage.xml
 	$(VIEWER) \
 	--block coverage.xml \
-	--goto $(ENTRY_GOTO).goto \
+	--goto $(ENTRY).goto \
 	--htmldir html \
 	--property property.xml \
 	--result cbmc.log \
--- a/.cbmc-batch/jobs/Makefile.common
+++ b/.cbmc-batch/jobs/Makefile.common
@@ -135,9 +135,6 @@ setup_dependencies:
 # sure changes in the headers files will also trigger make,
 # the user must run make clean first.
 $(ENTRY_GOTO)0.goto: setup_dependencies $(ENTRY).c $(DEPENDENT_GOTOS)
-	echo "FOOOO"
-	echo $(call LOG_FROM_ENTRY, $@)
-	echo $(dir $(call LOG_FROM_ENTRY, $@))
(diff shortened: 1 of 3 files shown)


.cbmc-batch/jobs/Makefile.common
```diff
@@ -23,6 +23,12 @@ sinclude Makefile.local
 SPACE :=$() $()
 COMMA :=,
 
+################################################################
+# Useful macros
+C_FROM_GOTO = $(patsubst $(GOTODIR)%, $(SRCDIR)%, $(1:.goto=.c))
+GOTO_FROM_C = $(patsubst $(SRCDIR)%, $(GOTODIR)%, $(1:.c=.goto))
+LOG_FROM_GOTO = $(patsubst $(GOTODIR)%, $(LOGDIR)%, $(1:.goto=.log))
```
karkhaz: I always thought that argument separators have to be just a comma (with no space afterwards) in Make, and the GNU Make manual itself says that. However, the actual behaviour seems to be that Make strips all whitespace before/after a comma, so this code does work.

I would have written this as
```Make
C_FROM_GOTO = $(patsubst $(GOTODIR)%,$(SRCDIR)%,$(1:.goto=.c))
```
which is less readable. But I don't have strong feelings about this, since your version does work.

.cbmc-batch/jobs/Makefile.common
```diff
@@ -124,90 +134,103 @@ setup_dependencies:
...
+		2>&1 | tee $(call LOG_FROM_ENTRY, $@) ; exit $${PIPESTATUS[0]}
+
+$(ENTRY_GOTO).goto: $(ENTRY_GOTO)7.goto
+	cp $< $@
 
-$(ENTRY).goto: $(ENTRY)7.goto
+# For now, CBMC batch expects the goto in the same folder.  I could update the
+# .yaml file for all 160 proofs, or I could copy it, and have one goto file
+# in the main directory.  
+$(ENTRY).goto: $(ENTRY_GOTO).goto
```
karkhaz: Just to confirm what is going on here: you're building all of the dependent goto object files out-of-tree, linking them to the harness goto file (also out-of-tree), then copying that final linked goto back into the source tree so that the proofs continue to just work?

I didn't initially know what `it` was referring to, I recommend clarifying the comment:
```make
# CBMC batch expects the goto binary to be in the same folder as the
# Makefile. Therefore copy the goto binary back to this directory
# from the out-of-tree build.
```

.cbmc-batch/jobs/aws_array_list_clear/cbmc-batch.yaml
```diff
@@ -1,4 +1,4 @@
-jobos: ubuntu16
-cbmcflags: "--bounds-check;--pointer-check;--div-by-zero-check;--signed-overflow-check;--unsigned-overflow-check;--pointer-overflow-check;--undefined-shift-check;--float-overflow-check;--nan-check;--unwinding-assertions;--unwind;1"
-goto: aws_array_list_clear_harness.goto
+cbmcflags: "--bounds-check;--div-by-zero-check;--float-overflow-check;--nan-check;--pointer-check;--pointer-overflow-check;--signed-overflow-check;--undefined-shift-check;--unsigned-overflow-check;--unwind;1;--unwinding-assertions;--object-bits;8"
```
karkhaz: The change to this file does not seem related to the PR. This change adds `--object-bits 8` to the command line flags and reorders the keys in this file, I would prefer to keep such changes separate (if they're necessary at all, i.e. the reordering)

.cbmc-batch/jobs/aws_is_power_of_two/cbmc-batch.yaml
```diff
@@ -1,4 +1,4 @@
-jobos: ubuntu16
-cbmcflags: "--bounds-check;--div-by-zero-check;--float-overflow-check;--nan-check;--pointer-check;--pointer-overflow-check;--signed-overflow-check;--undefined-shift-check;--unsigned-overflow-check;--unwind;1;--unwinding-assertions"
-goto: aws_is_power_of_two_harness.goto
+cbmcflags: "--bounds-check;--div-by-zero-check;--float-overflow-check;--nan-check;--pointer-check;--pointer-overflow-check;--signed-overflow-check;--undefined-shift-check;--unsigned-overflow-check;--unwind;1;--unwinding-assertions;--object-bits;8"
```
karkhaz: Again, unrelated change.

.cbmc-batch/jobs/Makefile.common
danielsn: Not quite.  We do have the out-of-source builds, in that all `.goto` files are now build under the proof directory.  However, I had tried to do a second level of cleanup, and move it so that all files would go in subdirectories of the proof itself called `logs` and `gotos`.

The `.yaml` file for the CI system expects the final `.goto` to be in the same folder as the proof itself.  So I can either update all `.yaml` files, or just copy the one `.goto` file.  But we're still building a separate build per proof.  I'll update the comment.

.cbmc-batch/jobs/aws_array_list_clear/cbmc-batch.yaml
danielsn: I'll remove this.  At some point it might be nice to go and remake all the `.yaml` files to use the latest template, so these changes don't keep slipping into PRs.

.cbmc-batch/jobs/aws_is_power_of_two/cbmc-batch.yaml
danielsn: Ack

.cbmc-batch/jobs/Makefile.common
danielsn: Done

.cbmc-batch/jobs/Makefile.common
karkhaz: Makes sense, thanks for the clarification

.cbmc-batch/jobs/aws_array_list_clear/cbmc-batch.yaml
karkhaz: Indeed, why do you actually commit the .yaml files? I think we stopped doing that in FreeRTOS. At least for us, those files get generated at CI time.

In fact, I remember that committing .yaml files caused a problem: somebody committed a .yaml file, then changed the Makefile that generated the yaml file. But because the .yaml file was already in the repo and therefore 'up-to-date', Make did not rebuild it under CI, so the CI used old flags.

I'd recommend removing all .yaml files from the repo if possible.

PR #536, opened 2019-11-01
Generalized Mem tracing

justinboswell:


include/aws/common/allocator.h
```diff
@@ -110,6 +110,50 @@ int aws_mem_realloc(struct aws_allocator *allocator, void **ptr, size_t oldsize,
...
+
+/*
+ * Wraps an allocator and tracks all external allocations. If aws_mem_trace_dump() is called
+ * and there are still allocations active, they will be reported to stderr.
+ * frames_per_stack is how many frames to store per callstack if AWS_MEMTRACE_STACKS is in use,
+ * otherwise it is ignored. 8 tends to be a pretty good number balancing storage space vs useful stacks.
+ * Returns the tracer allocator, which should be used for all allocations that should be tracked.
+ */
+AWS_COMMON_API
+struct aws_allocator *aws_mem_tracer_new(
```
graebm: comment here says `stderr`, but actually it goes through logger at TRACE level.
Update comment, or actually printf to stderr

justinboswell: Fixed.

source/memtrace.c
```diff
@@ -0,0 +1,488 @@
...
+    return ptr;
+}
+
+struct aws_allocator *aws_mem_tracer_new(
+    struct aws_allocator *allocator,
+    enum aws_mem_trace_level level,
+    size_t frames_per_stack) {
+    struct aws_allocator *trace_allocator = aws_mem_calloc(aws_default_allocator(), 1, sizeof(struct aws_allocator));
+    AWS_FATAL_ASSERT(trace_allocator);
+    struct alloc_tracer *tracer = aws_mem_calloc(aws_default_allocator(), 1, sizeof(struct alloc_tracer));
```
ColdenCullen: `allocator`

And for bonus points, use acuire_many for the allocator and it's impl.

source/posix/system_info.c
```diff
@@ -236,6 +239,63 @@ void s_resolve_cmd(char *cmd, size_t len, struct aws_stack_frame_info *frame) {
...
+    aws_byte_buf_init(&lines, aws_default_allocator(), stack_depth * 256);
+
+    /* insert pointers for each stack entry */
+    memset(lines.buffer, 0, stack_depth * sizeof(void *));
+    lines.len += stack_depth * sizeof(void *);
+
+    /* symbols look like: <exe-or-shared-lib>(<function>+<addr>) [0x<addr>]
+     *                or: <exe-or-shared-lib> [0x<addr>]
+     * start at 1 to skip the current frame (this function) */
+    for (int frame_idx = 0; frame_idx < stack_depth; ++frame_idx) {
```
ColdenCullen: `size_t frame_idx`

source/posix/system_info.c
```diff
@@ -236,6 +239,63 @@ void s_resolve_cmd(char *cmd, size_t len, struct aws_stack_frame_info *frame) {
...
+
+    parse_failed:
+        /* record the pointer to where the symbol will be */
+        *((char **)&lines.buffer[frame_idx * sizeof(void *)]) = (char *)lines.buffer + lines.len;
+        struct aws_byte_cursor line_cursor = aws_byte_cursor_from_c_str(symbol);
+        line_cursor.len += 1; /* strings must be null terminated, make sure we copy the null */
+        aws_byte_buf_append_dynamic(&lines, &line_cursor);
+    }
+    free(symbols);
+    return (char **)lines.buffer; /* caller is responsible for freeing */
```
ColdenCullen: This makes me really nervous, it goes against all of the safety mechanisms we have in place. Can we not just return the byte_buf instead? Then the user just calls `aws_byte_buf_clean_up`

source/windows/system_info.c
```diff
@@ -80,58 +86,131 @@ void aws_backtrace_print(FILE *fp, void *call_site_data) {
...
+    }
+    return false;
+}
+
+int aws_backtrace(void **frames, size_t size) {
+    return (int)CaptureStackBackTrace(0, (ULONG)size, frames, NULL);
+}
+
+char **aws_backtrace_symbols(void *const *stack, size_t num_frames) {
+    if (!s_init_dbghelp()) {
```
ColdenCullen: Should we do this as a `aws_thread_call_once`?

include/aws/common/allocator.h
```diff
@@ -110,6 +110,50 @@ int aws_mem_realloc(struct aws_allocator *allocator, void **ptr, size_t oldsize,
...
+ * Should be passed the tracer allocator returned from aws_mem_trace().
+ */
+AWS_COMMON_API
+void aws_mem_trace_dump(struct aws_allocator *trace_allocator);
+
+/*
+ * Returns the current number of bytes in outstanding allocations
+ */
+AWS_COMMON_API
+size_t aws_mem_tracer_count(struct aws_allocator *trace_allocator);
```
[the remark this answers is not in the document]
justinboswell: Added ```aws_mem_tracer_bytes``` and implemented both

include/aws/common/system_info.h
```diff
@@ -32,6 +32,48 @@ bool aws_is_debugger_present(void);
...
+#    define AWS_BACKTRACE_STACKS_AVAILABLE
+#endif
+
+/*
+ * Records a stack trace from the call site.
+ * Returns the number of stack entries/stack depth captured, or 0 if the operation
+ * is not supported on this platform
+ */
+AWS_COMMON_API
+int aws_backtrace(void **frames, size_t num_frames);
```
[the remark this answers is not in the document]
justinboswell: Good call.

source/windows/system_info.c
justinboswell: It can fail, so we would crash trying to use the resolved functions. Otherwise, yeah, that'd be the best way.

source/posix/system_info.c
justinboswell: I considered that. The reason I didn't is that it's an array of strings. The user would have to unpack it from the byte_buf manually, by grabbing ```buf.buffer``` and casting it to a ```const char**```. Also, the calls we are wrapping/emulating/replacing work this way. Thoughts?

ColdenCullen: I suppose that makes sense.

The only thing I can think to do that would make the byte_buf version work better would be to add a `aws_byte_buf_write_ptr()` and `aws_byte_cursor_read_ptr()`, but that seems like a bad pattern to enable. Let's leave as is.

source/windows/system_info.c
ColdenCullen: Does that mean that if it does fail, we're going to retry every time one of these functions get called? Could those ever succeed? It might be cleaner to add a have something like this:
```c
if s_try_init_dbghelp() {
    if (one_of_the_functions) {
        return true;
    } else {
        aws_thread_call_once(do_init);
        return one_of_the_functions != NULL;
    }
}
```

source/memtrace.c
```diff
@@ -240,23 +244,23 @@ static int s_stack_info_compare_count(const void *a, const void *b) {
 }
 
 static void s_stack_info_destroy(void *data) {
-    struct aws_allocator *allocator = aws_default_allocator();
     struct stack_info_t *stack = data;
+    struct aws_allocator *allocator = stack->trace->allocator;
```
ColdenCullen: clever girl

source/memtrace.c
```diff
@@ -0,0 +1,488 @@
...
+/* describes a single live allocation */
+struct alloc_t {
+    size_t size;
+    time_t time;
+    uint64_t stack; /* hash of stack frame pointers */
+};
+
+/* one of these is stored per unique stack */
+struct stacktrace_t {
+    void *const frames[1]; /* rest of frames are allocated after */
```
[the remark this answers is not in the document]
justinboswell: Old habit

PR #538, opened 2019-11-06
Use PROJECT_NAME instead of CMAKE_PROJECT_NAME

madebr:
*Description of changes:*
Use `PROJECT_NAME` instead of `CMAKE_PROJECT_NAME` to refer to the project name in the cmake scripts.

`CMAKE_PROJECT_NAME` is the name of the top project.
https://cmake.org/cmake/help/latest/variable/CMAKE_PROJECT_NAME.html

`PROJECT_NAME` is the name of the most recent `project()` command.
https://cmake.org/cmake/help/latest/variable/PROJECT_NAME.html

This fixes including the aws-c-common project as a subproject.


ColdenCullen:
Looks like this broke the header test, try configuring with `-DPERFORM_HEADER_CHECK=ON`, hopefully that helps! But it appears as though include paths aren't being set properly.

madebr:
@ColdenCullen Since neither `CMAKE_PROJECT_NAME` or `PROJECT_NAME` was appropriate in the `header-check` project, I've just used `aws-c-common`.

An alternative would be to remove the `project` in `cmake/header-tester/CMakeLists.txt` and keep using `PROJECT_NAME`.

ColdenCullen:
Thanks for the quick response!

> An alternative would be to remove the `project` in `cmake/header-tester/CMakeLists.txt` and keep using `PROJECT_NAME`.

Sorry to be a super both, but could you try this? I would eventually like to get the header-tester generalized so we can use it in the other libs, and I think this would be the best way to work towards that goal.

madebr:
@ColdenCullen The header-tester now does not include any `project` call anymore.

I think this cmake file can now be copied to other libs without change.

PR #540, opened 2019-11-07
Always print both raw and resolved stack trace, helps in release builds

justinboswell:
Also fixes a couple crashes that are possible when the symbol resolves to only an address of the form ```[0x<addr>]```

### DIFF
--- a/source/posix/system_info.c
+++ b/source/posix/system_info.c
@@ -191,6 +191,7 @@ void s_resolve_cmd(char *cmd, size_t len, struct aws_stack_frame_info *frame) {
 int s_parse_symbol(const char *symbol, void *addr, struct aws_stack_frame_info *frame) {
     /* symbols look like: <exe-or-shared-lib>(<function>+<addr>) [0x<addr>]
      *                or: <exe-or-shared-lib> [0x<addr>]
+     *                or: [0x<addr>]
      */
     (void)addr;
     const char *open_paren = strstr(symbol, "(");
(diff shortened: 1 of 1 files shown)


source/posix/system_info.c
```diff
@@ -285,7 +293,7 @@ char **aws_backtrace_addr2line(void *const *stack_frames, size_t stack_depth) {
         }
         pclose(out);
 
-    parse_failed:
+        parse_failed:
```
ColdenCullen: this may not be intentional?

source/posix/system_info.c
```diff
@@ -341,7 +361,7 @@ void aws_backtrace_print(FILE *fp, void *call_site_data) {
...
```
ColdenCullen: samslies

source/posix/system_info.c
```diff
@@ -313,8 +321,20 @@ void aws_backtrace_print(FILE *fp, void *call_site_data) {
         return;
     }
 
+    fprintf(fp, "################################################################################");
+    fprintf(fp, "Raw stacktrace:\n");
+    fprintf(fp, "################################################################################");
+    for (size_t frame_idx = 1; frame_idx < stack_depth; ++frame_idx) {
+        const char *symbol = symbols[frame_idx];
+        fprintf(fp, "%s\n", symbol);
+    }
```
ColdenCullen: Can we do this one second? the resolved stack trace is likely to be more helpful, and should probably be printed first.

PR #543, opened 2019-11-10
added naive string find function.

JonathanHenson:


source/byte_buf.c
```diff
@@ -257,6 +257,44 @@ int aws_byte_cursor_split_on_char(
...
+    }
+
+    if (to_find->len < 1) {
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+    }
+
+    struct aws_byte_cursor working_cur = *input_str;
+
+    while (working_cur.len) {
+        uint8_t *first_char_location = memchr(working_cur.ptr, (char)*working_cur.ptr, working_cur.len);
```
bretambrose: I think this does nothing.  I'm guessing you meant "(char)*to_find->ptr" which could be cached outside the loop for clarity (compiler probably would no matter what)

justinboswell: I had the same thought...

marcomagdy: now I'm curious how the positive and the negative unit tests both passed!!!!

source/byte_buf.c
```diff
@@ -257,6 +257,44 @@ int aws_byte_cursor_split_on_char(
     return aws_byte_cursor_split_on_char_n(input_str, split_on, 0, output);
 }
 
+int aws_byte_cursor_find_exact(
+    const struct aws_byte_cursor *AWS_RESTRICT input_str,
+    const struct aws_byte_cursor *AWS_RESTRICT to_find,
```
marcomagdy: nitpick: I don't think those inputs need to be marked restrict, because you don't read from one after you write to the other. So, there's no performance penalty if they alias.

That being said, I don't think it hurts.

source/byte_buf.c
JonathanHenson: oh woops! I was watching Lawrence of Arabia in 4k while writing this, must have been distracted by Peter O'Toole's brilliant over-acting.

JonathanHenson: Marco, it worked because it's technically correct, just VERY slow.

source/byte_buf.c
JonathanHenson: I can't promise it won't be when I finally come back and implement Boyer-Moore

Issue #562, opened 2019-12-19
Non portable use of posix pthread_t


nchong-at-aws:
posix pthread_t is opaque so comparison with `==` and printing as a `uint64_t` is not portable.

http://man7.org/linux/man-pages/man3/pthread_self.3.html

PR #563, opened 2019-12-19
Remove non portable use of pthread_t

nchong-at-aws:
*Issue #, if available:* #562

*Description of changes:*

Add platform-specific typedef `aws_thread_id` and portable printing/comparison functions.

### DIFF
--- a/include/aws/common/thread.h
+++ b/include/aws/common/thread.h
@@ -37,20 +37,22 @@ typedef union {
 } aws_thread_once;
 #    define AWS_THREAD_ONCE_STATIC_INIT                                                                                \
         { NULL }
+typedef unsigned long aws_thread_id;
 #else
 typedef pthread_once_t aws_thread_once;
 #    define AWS_THREAD_ONCE_STATIC_INIT PTHREAD_ONCE_INIT
+typedef pthread_t aws_thread_id;
 #endif
+
+#define AWS_THREAD_ID_REPR_LEN (sizeof(aws_thread_id) * 2 + 1)
 
 struct aws_thread {
     struct aws_allocator *allocator;
     enum aws_thread_detach_state detach_state;
 #ifdef _WIN32
     void *thread_handle;
-    unsigned long thread_id;
-#else
-    pthread_t thread_id;
 #endif
+    aws_thread_id thread_id;
 };
 
 AWS_EXTERN_C_BEGIN
@@ -86,7 +88,7 @@ int aws_thread_launch(
  * Gets the id of thread
  */
 AWS_COMMON_API
-uint64_t aws_thread_get_id(struct aws_thread *thread);
+aws_thread_id aws_thread_get_id(struct aws_thread *thread);
 
 /**
  * Gets the detach state of the thread. For example, is it safe to call join on
@@ -110,10 +112,16 @@ AWS_COMMON_API
 void aws_thread_clean_up(struct aws_thread *thread);
 
 /**
- * returns the thread id of the calling thread.
+ * Returns the thread id of the calling thread.
  */
 AWS_COMMON_API
-uint64_t aws_thread_current_thread_id(void);
+aws_thread_id aws_thread_current_thread_id(void);
+
+/**
+ * Compare thread ids.
+ */
+AWS_COMMON_API
+bool aws_thread_thread_id_equal(aws_thread_id t1, aws_thread_id t2);
 
 /**
  * Sleeps the current thread by nanos.
--- a/include/aws/common/thread.h
+++ b/include/aws/common/thread.h
@@ -37,19 +37,19 @@ typedef union {
 } aws_thread_once;
 #    define AWS_THREAD_ONCE_STATIC_INIT                                                                                \
         { NULL }
-typedef unsigned long aws_thread_id;
+typedef unsigned long aws_thread_id_t;
 #else
 typedef pthread_once_t aws_thread_once;
 #    define AWS_THREAD_ONCE_STATIC_INIT PTHREAD_ONCE_INIT
-typedef pthread_t aws_thread_id;
+typedef pthread_t aws_thread_id_t;
 #endif
 
 /*
- * Number of chars to represent aws_thread_id as a string (2 hex chars per byte
+ * Buffer size needed to represent aws_thread_id_t as a string (2 hex chars per byte
  * plus '\0' terminator). Needed for portable printing because pthread_t is
  * opaque.
  */
-#define AWS_THREAD_ID_REPR_LEN (sizeof(aws_thread_id) * 2 + 1)
+#define AWS_THREAD_ID_T_REPR_BUFSZ (sizeof(aws_thread_id_t) * 2 + 1)
 
 struct aws_thread {
     struct aws_allocator *allocator;
@@ -57,7 +57,7 @@ struct aws_thread {
 #ifdef _WIN32
     void *thread_handle;
 #endif
-    aws_thread_id thread_id;
+    aws_thread_id_t thread_id;
 };
 
 AWS_EXTERN_C_BEGIN
@@ -93,7 +93,7 @@ int aws_thread_launch(
  * Gets the id of thread
  */
 AWS_COMMON_API
-aws_thread_id aws_thread_get_id(struct aws_thread *thread);
+aws_thread_id_t aws_thread_get_id(struct aws_thread *thread);
 
 /**
  * Gets the detach state of the thread. For example, is it safe to call join on
@@ -120,13 +120,13 @@ void aws_thread_clean_up(struct aws_thread *thread);
  * Returns the thread id of the calling thread.
  */
 AWS_COMMON_API
-aws_thread_id aws_thread_current_thread_id(void);
+aws_thread_id_t aws_thread_current_thread_id(void);
 
 /**
  * Compare thread ids.
  */
 AWS_COMMON_API
-bool aws_thread_thread_id_equal(aws_thread_id t1, aws_thread_id t2);
+bool aws_thread_thread_id_equal(aws_thread_id_t t1, aws_thread_id_t t2);
 
 /**
  * Sleeps the current thread by nanos.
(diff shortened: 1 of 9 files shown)


tests/thread_test.c
```diff
@@ -38,9 +37,8 @@ static int s_test_thread_creation_join_fn(struct aws_allocator *allocator, void
     ASSERT_INT_EQUALS(
         AWS_THREAD_JOINABLE, aws_thread_get_detach_state(&thread), "thread state should have returned JOINABLE");
     ASSERT_SUCCESS(aws_thread_join(&thread), "thread join failed");
-    ASSERT_INT_EQUALS(
-        test_data.thread_id,
-        aws_thread_get_id(&thread),
+    ASSERT_TRUE(
```
danielsn: Maybe have an assert false test here too?

tests/error_test.c
```diff
@@ -359,20 +359,20 @@ static int s_unknown_error_code_range_too_large_test_fn(struct aws_allocator *al
...
     int thread_2_encountered_count;
-    uint64_t thread_2_id;
+    aws_thread_id thread_2_id;
 };
 
 static void s_error_thread_test_thread_local_cb(int err, void *ctx) {
     struct error_thread_test_data *cb_data = (struct error_thread_test_data *)ctx;
 
-    uint64_t thread_id = aws_thread_current_thread_id();
+    aws_thread_id thread_id = aws_thread_current_thread_id();
```
danielsn: Is this used again? If not, maybe move it inside the test?

source/log_formatter.c
```diff
@@ -109,16 +109,16 @@ int aws_format_standard_log_line(struct aws_logging_standard_formatting_data *fo
         /*
          * Add thread id and user content separator (" - ")
          */
-        uint64_t current_thread_id = aws_thread_current_thread_id();
+        aws_thread_id current_thread_id = aws_thread_current_thread_id();
+        char repr[AWS_THREAD_ID_REPR_LEN];
+        if (aws_thread_id_to_string(current_thread_id, repr, AWS_THREAD_ID_REPR_LEN)) {
```
bretambrose: The logger is pretty slow naturally, but I'm still a little worried about potential overhead here for Trace situations.  The value is const, so what about adding the thread_id's string value as a member of aws_thread and a getter for that value rather than doing the to-string loop with every log call?

include/aws/common/thread.h
```diff
@@ -110,10 +112,16 @@ AWS_COMMON_API
 void aws_thread_clean_up(struct aws_thread *thread);
 
 /**
- * returns the thread id of the calling thread.
+ * Returns the thread id of the calling thread.
+ */
+AWS_COMMON_API
+aws_thread_id aws_thread_current_thread_id(void);
```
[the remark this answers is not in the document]
nchong-at-aws: This can't fail. On posix platforms it maps to `pthread_self()`. On windows to `GetCurrentThreadId()`.

http://man7.org/linux/man-pages/man3/pthread_self.3.html
https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentthreadid

source/posix/thread.c
```diff
@@ -98,7 +98,7 @@ void aws_thread_call_once(aws_thread_once *flag, void (*call_once)(void *), void
 
 int aws_thread_init(struct aws_thread *thread, struct aws_allocator *allocator) {
     thread->allocator = allocator;
-    thread->thread_id = 0;
+    // do not init opaque thread->thread_id
```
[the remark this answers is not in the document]
nchong-at-aws: I changed the init as you suggested, but the real point is that there is no designated invalid phread_t value.

tests/thread_test.c
nchong-at-aws: To do that, we would need to manufacture an aws_thread_id value that we know is not the current thread. One way would be to create a new thread and then check against that.

source/log_formatter.c
nchong-at-aws: Yes, I see... except, I think this wouldn't work for logging in the main procedure since that isn't tied to an aws_thread instance. Instead, I made a thread-local copy of the string representation so the cost is at most once per thread.

source/log_formatter.c
```diff
@@ -51,6 +51,13 @@ static size_t s_advance_and_clamp_index(size_t current_index, int amount, size_t
 
     return next_index;
 }
+
+/* Thread-local string representation of current thread id */
+AWS_THREAD_LOCAL struct {
+    bool is_valid;
+    char repr[AWS_THREAD_ID_REPR_LEN];
```
danielsn: If you also store the actual `thread_id` here, then this becomes the thing you take the address of in aws-c-io.

nchong-at-aws: That could work, but would couple together c-io event loops and c common logging. Instead, how about changing the API for getting the current thread id to return a pointer to a thread-local `aws_thread_id` variable? The string repr can still be kept in logging, which is the only place it is needed.

```c
// in posix/thread.c
AWS_THREAD_LOCAL struct {
    bool is_valid;
    aws_thread_id thread_id;
} tl_current_thread = {.is_valid = false};

const aws_thread_id *aws_thread_current_thread_id(void) {
    if (!tl_current_thread.is_valid) {
        tl_current_thread.thread_id = pthread_self();
        tl_current_thread.is_valid = true;
    }
    return &tl_current_thread.thread_id;
}
```

danielsn:
What if we moved the logging code to c-common. Seems like it would be
useful for other libraries as well

source/log_formatter.c
graebm: Two things:
1) "Many systems impose restrictions on the size of the thread-local memory block, in fact often rather tight limits." [wikipedia](https://en.wikipedia.org/wiki/Thread-local_storage). So it's not a great idea to make a bunch of thread-local storage variables, each of which caches one tiny thing.

2) Is there a chance of the us trying to query the tl_current_thread after the thread is gone? I know the logger usually uses one logging thread. Other threads add their statements to a queue, and the logging thread drains the queue. If a thread logs something right before it exits, then the logging thread is going to try to get its name a little bit later and it will already be gone.

source/logging.c
```diff
@@ -291,6 +291,26 @@ int aws_log_level_to_string(enum aws_log_level log_level, const char **level_str
...
 }
 
+int aws_thread_id_to_string(aws_thread_id thread_id, char *str, size_t length) {
+    AWS_ERROR_PRECONDITION(AWS_THREAD_ID_REPR_LEN == length);
+    AWS_ERROR_PRECONDITION(str && AWS_MEM_IS_WRITABLE(str, length));
+    size_t current_index = 0;
+    unsigned char *bytes = (unsigned char *)&thread_id;
+    for (size_t i = sizeof(aws_thread_id); i != 0; --i) {
+        unsigned char c = bytes[i - 1];
+        int written = snprintf(str + current_index, length - current_index, "%02x", c);
```
graebm: very debatable: less flexible, but more efficient, to just call snprintf once
```
#if sizeof(aws_thread_id_t) == 4
    snprintf(str, length, "%08x", (uint32_t)thread_id);
#elif sizeof(aws_thread_id_t) == 8
    snprintf(str, length, "%016x", (uint64_t)thread_id);
#else
#    error
#end
```

source/logging.c
```diff
@@ -291,6 +291,26 @@ int aws_log_level_to_string(enum aws_log_level log_level, const char **level_str
     return AWS_OP_SUCCESS;
 }
 
+int aws_thread_id_to_string(aws_thread_id thread_id, char *str, size_t length) {
+    AWS_ERROR_PRECONDITION(AWS_THREAD_ID_REPR_LEN == length);
```
graebm: debatable: it's a little weird to make the user pass length, but there is only one correct length. Though I understand that we're concerned about overflow, so this is kind of like making the user sign a waiver each time. I could go either way...

However, I do care about this: "length" when you're talking about c-strings usually doesn't include the null terminator. In the [snprintf docs](https://en.cppreference.com/w/c/io/fprintf) they call their args `(char *buffer, size_t bufsz, ...)` maybe we should do the same?

include/aws/common/thread.h
```diff
@@ -37,20 +37,27 @@ typedef union {
...
 #    define AWS_THREAD_ONCE_STATIC_INIT PTHREAD_ONCE_INIT
+typedef pthread_t aws_thread_id;
 #endif
 
+/*
+ * Number of chars to represent aws_thread_id as a string (2 hex chars per byte
+ * plus '\0' terminator). Needed for portable printing because pthread_t is
+ * opaque.
+ */
+#define AWS_THREAD_ID_REPR_LEN (sizeof(aws_thread_id) * 2 + 1)
```
graebm: "length" when talking about c-strings doesn't usually include the null-terminator.
For example `AWS_DATE_TIME_STR_MAX_LEN` does not account for the null terminator.
Maybe `AWS_THREAD_ID_REPR_BUFSZ`? or keep it as `_LEN` but remove the +1?

source/logging.c
nchong-at-aws: 1. TLS limits. Yes, this could be a problem if a platform has limited size. In this case, we would not be able to cache the string repr and we'd have to recompute it each time into a stack-local variable.
2. I think this isn't a concern because the logging thread deals with formatted strings that have already had their thread_id resolved. I.e., the logging thread does not query `tl_current_thread` itself.
3. This would be nice except `sizeof` in macros isn't supported

bretambrose: RE: 1
I don't think it should hold up this PR but we should consider adding a thread-local storage system in common that everything else hangs off of so that the CRT's usage is always a single slot.  It could be a mixture of static data (like thread id's stringified value) and key-value data.

nchong-at-aws:
Thanks all. Changes to downstream dependencies: I think only `aws-c-io` needs updating. I did a quick check (grep for thread) of `aws-checksums`, `aws-c-event-stream`, `aws-c-compression`, `aws-c-cal`, `aws-c-mqtt`, `aws-c-http` and `aws-c-auth`.

source/logging.c
nchong-at-aws: Thanks Bret. I made your comment into an issue (#565).

PR #566, opened 2020-01-02
support in source builds of aws-c-* libs.

JonathanHenson:


cmake/AwsFindPackage.cmake
```diff
@@ -0,0 +1,32 @@
...
+# This is largely because CMake was not well planned out, and as a result, in-source and modules don't play well
+# together. Only use this on CRT libraries (including S2N), libcrypto will stay as an assumed external dependency.
+#
+# package_name: is the name of the package to find
+# DEP_CRT_LIBS: output variable will be appended after each call to this function. You don't have to use it,
+#    but it can be passed directly target_link_libraries and it will be the properly qualified library
+#    name and namespace based on configuration.
+function(aws_use_package package_name)
+    if (IN_SOURCE_BUILD)
+        set(DEP_CRT_LIBS ${DEP_CRT_LIBS} ${package_name} PARENT_SCOPE)
```
graebm: trivia: aws_use_package is a good name, but not sure about use of the term "CRT" down here in aws-c-common, and the use of variables like "IN_SOURCE_BUILD" that aren't prefixed with "AWS"

JonathanHenson: none of our other cmake options are prefixed with AWS.

There's multiple deps list that users may maintain and I wanted it namespaced, what would you recommend? DEP_AWS_LIBS ? AWS_LIBS? CRT_LIBS?

PR #570, opened 2020-01-03
Cmake is the worst.

JonathanHenson:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -232,7 +232,7 @@ configure_file(${CONFIG_HEADER_TEMPLATE}
         ESCAPE_QUOTES)
 
 if (NOT CMAKE_CROSSCOMPILING)
-    if (BUILD_TESTING)
+    if (NOT DEFINED BUILD_TESTING OR BUILD_TESTING)
         include(CTest)
         add_subdirectory(tests)
     endif()


CMakeLists.txt
```diff
@@ -232,7 +232,7 @@ configure_file(${CONFIG_HEADER_TEMPLATE}
         ESCAPE_QUOTES)
 
 if (NOT CMAKE_CROSSCOMPILING)
-    if (BUILD_TESTING)
+    if (NOT DEFINED BUILD_TESTING OR BUILD_TESTING)
         include(CTest)
         add_subdirectory(tests)
     endif()
```
graebm: CTest defines the `BUILD_TESTING` option, should this be:
```
if (NOT CMAKE_CROSSCOMPILING)
    include(CTest)
    if (BUILD_TESTING)
        add_subdirectory(tests)
    endif()
endif()
```
???

JonathanHenson: no it shouldn't, otherwise this PR wouldn't exist

graebm: https://cmake.org/cmake/help/v3.1/module/CTest.html

JonathanHenson: I don't care what the docs say.... they;re wrong

JonathanHenson: we could move include(CTest) to the top of the project if you'd like

PR #572, opened 2020-01-08
Builder work

bretambrose:
Nested build path overrides were getting stomped by the top-level values due to the order of inputs to the hash map construction expression.


codebuild/builder.py
```diff
@@ -720,12 +720,12 @@ def _test_project_ctest():
...
 
             command_variables = {
+                **config,
+                **config['variables'],
                 'source_dir': project_source_dir,
                 'build_dir': project_build_dir,
                 'install_dir': install_dir,
                 'build_config': build_config,
-                **config,
-                **config['variables'],
```
graebm: WOOOOOOOOOOOOOOOOW

PR #573, opened 2020-01-14
Split alignment tests 

bretambrose:
Split alignment tests into 16 and 32 bytes; 32 enabled only if avx2 support is active

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -241,7 +241,13 @@ add_test_case(test_byte_cursor_compare_lexical)
 add_test_case(test_byte_cursor_compare_lookup)
 
 add_test_case(byte_swap_test)
-add_test_case(alignment_test)
+
+if (HAVE_AVX2_INTRINSICS AND HAVE_SIMD_CPUID)
+    add_test_case(alignment32_test)
+else()
+    add_test_case(alignment16_test)
+endif()
+
 
 add_test_case(test_cpu_count_at_least_works_superficially)
 add_test_case(test_stack_trace_decoding)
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -241,7 +241,13 @@ add_test_case(test_byte_cursor_compare_lexical)
 add_test_case(test_byte_cursor_compare_lookup)
 
 add_test_case(byte_swap_test)
-add_test_case(alignment_test)
+
+if (HAVE_AVX2_INTRINSICS AND HAVE_SIMD_CPUID)
+    add_test_case(alignment32_test)
+else()
+    add_test_case(alignment16_test)
```
justinboswell: Will 16 always work? Isn't the guarantee only 8?

bretambrose: Henso suggested 16 which is what we've seen consistently on the failed internal builds.  Maybe dial back to 8 only if we have to.

tests/byte_order_test.c
```diff
@@ -66,23 +66,42 @@ static int s_byte_swap_test_fn(struct aws_allocator *allocator, void *ctx) {
...
+static int s_alignment32_test_fn(struct aws_allocator *allocator, void *ctx) {
     (void)allocator;
     (void)ctx;
 
-    struct padding_disaster padded;
+    struct padding32_disaster padded;
 
-    ASSERT_UINT_EQUALS(0, ((intptr_t)&padded.a) % 32);
-    ASSERT_UINT_EQUALS(0, ((intptr_t)&padded.b) % 32);
+    ptrdiff_t spacing = (intptr_t)&padded.b - (intptr_t)&padded.dumb;
```
justinboswell: this could also be ```offsetof(padding32_disaster, b) - offsetof(padding32_disaster, dumb)```

bretambrose: Sure, updated.

PR #574, opened 2020-01-17
thread-sanitizer fix spiritual journey

graebm:
- Fixed missing thread_join() in a unit test
- memtrace.c fixes
    - don't clean up stuff we never initialized
    - fix places where tracer wasn't using system allocator for bookkeeping.
- turned off clang-format on a file full of ASM. it was failing clang-format on my machine, not sure why build machines didn't care

### DIFF
--- a/source/memtrace.c
+++ b/source/memtrace.c
@@ -25,6 +25,7 @@
 
 /* describes a single live allocation */
 struct alloc_info {
+    struct aws_allocator *allocator; /* Allocator of this object */
     size_t size;
     time_t time;
     uint64_t stack; /* hash of stack frame pointers */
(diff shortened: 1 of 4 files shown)


source/memtrace.c
```diff
@@ -483,35 +498,37 @@ struct aws_allocator *aws_mem_tracer_new(
...
+    }
 
-    struct aws_allocator *system_allocator = tracer->system_allocator;
-    aws_mem_release(system_allocator, tracer);
+    aws_mem_release(tracer->system_allocator, tracer);
     /* trace_allocator is freed as part of the block tracer was allocated in */
-    return allocator;
-}
 
-void aws_mem_tracer_dump(struct aws_allocator *trace_allocator) {
```
bretambrose: Was this delete intended?

source/memtrace.c
```diff
@@ -104,7 +103,7 @@ static void s_alloc_tracer_init(
         level = level > AWS_MEMTRACE_BYTES ? AWS_MEMTRACE_BYTES : level;
     }
 
-    tracer->allocator = allocator;
+    tracer->traced_allocator = traced_allocator;
```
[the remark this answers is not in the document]
graebm: It's valid if there the same, they could both just call into `malloc()` under the hood.

source/memtrace.c
graebm: yeah, that was intentional.

`aws_mem_tracer_dump()` was just calling `s_alloc_tracer_dump()`.
I renamed `s_alloc_tracer_dump()` to `aws_mem_tracer_dump()` (higher up in the file) and removed useless indirection down here.

PR #575, opened 2020-01-21
idempotent clean_up() for mutex & condition_variable

graebm:
Makes error-handling simpler for datastructures that have mutexes/cond-vars within them.
the `error:` label can clean_up() the mutex whether or not it was ever initialized.

### DIFF
--- a/include/aws/common/condition_variable.h
+++ b/include/aws/common/condition_variable.h
@@ -33,6 +33,7 @@ struct aws_condition_variable {
 #else
     pthread_cond_t condition_handle;
 #endif
+    bool initialized;
 };
 
 /**
@@ -45,10 +46,10 @@ struct aws_condition_variable {
  */
 #ifdef _WIN32
 #    define AWS_CONDITION_VARIABLE_INIT                                                                                \
-        { .condition_handle = NULL }
+        { .condition_handle = NULL, .initialized = true }
 #else
 #    define AWS_CONDITION_VARIABLE_INIT                                                                                \
-        { .condition_handle = PTHREAD_COND_INITIALIZER }
+        { .condition_handle = PTHREAD_COND_INITIALIZER, .initialized = true }
 #endif
 
 AWS_EXTERN_C_BEGIN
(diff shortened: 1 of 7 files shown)


include/aws/common/condition_variable.h
```diff
@@ -33,6 +33,7 @@ struct aws_condition_variable {
 #else
     pthread_cond_t condition_handle;
 #endif
+    bool initialized;
```
justinboswell: Isn't condition_handle != 0/NULL enough to flag init?

include/aws/common/mutex.h
```diff
@@ -30,14 +30,15 @@ struct aws_mutex {
 #else
     pthread_mutex_t mutex_handle;
 #endif
+    bool initialized;
```
justinboswell: Isn't mutex_handle != 0/NULL enough to flag init?

include/aws/common/mutex.h
```diff
@@ -19,7 +19,7 @@
 #include <aws/common/common.h>
 #ifdef _WIN32
 /* NOTE: Do not use this macro before including Windows.h */
-#    define AWSMUTEX_TO_WINDOWS(pCV) (PSRWLOCK) pCV
+#    define AWSMUTEX_TO_WINDOWS(pCV) AWS_CONTAINER_OF((pCV), PSRWLOCK, mutex_handle)
```
justinboswell: ```AWS_MUTEX_TO_WINDOWS``` if this is public?

include/aws/common/condition_variable.h
graebm: On Windows, the AWS_CONDITION_VARIABLE_INIT macro zeros out the datastructure. So checking if it's zeroed will not tell you whether or not its been initialized.

For pthread, it's **probably** one of those "The spec doesn't say that 0 can't be valid" things. I'm just speculating there. I don't know where one goes to read the pthread standard. I dug into the headers on Mac and `PTHREAD_COND_INITIALIZER` makes it non-zero, but I can't say for sure on other platforms.

include/aws/common/mutex.h
graebm: Same, AWS_MUTEX_INIT zeroes out the handle on Windows.

include/aws/common/mutex.h
graebm: It's public because `windows/condition_variable.c` is using it

PR #581, opened 2020-01-24
More byte_buf.h functions for HTTP/2

graebm:
Better way to read/write an integer with 3 bytes (HTTP/2 encodes length this way).
- DELETE `aws_hton24()` and `aws_ntoh24()`.
  - These had bugs in the big-endian implementation, and it was hard to understand how to use them correctly.
- ADD `aws_byte_cursor_read_be24()` reads 3 bytes into a u32.
- ADD `aws_byte_buffer_write_be24()` writes "low" 3 bytes of u32.

Better way to write padding zeroes.
- ADD `aws_byte_buf_write_u8_n()` which is like memset()

### DIFF
--- a/source/byte_buf.c
+++ b/source/byte_buf.c
@@ -1107,6 +1107,35 @@ bool aws_byte_cursor_read_be16(struct aws_byte_cursor *cur, uint16_t *var) {
 }
 
 /**
+ * Reads an unsigned 24-bit value (3 bytes) in network byte order from cur,
+ * and places it in host byte order into 32-bit var.
+ * Ex: if cur's next 3 bytes are {0xAA, 0xBB, 0xCC}, then var becomes 0x00AABBCC.
+ *
+ * On success, returns true and updates the cursor pointer/length accordingly.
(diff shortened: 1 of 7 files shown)


graebm:
DO NOT MERGE until downstream fixes are ready in aws-c-http

source/byte_buf.c
```diff
@@ -1231,7 +1260,7 @@ bool aws_byte_buf_advance(
  */
 bool aws_byte_buf_write(struct aws_byte_buf *AWS_RESTRICT buf, const uint8_t *AWS_RESTRICT src, size_t len) {
     AWS_PRECONDITION(aws_byte_buf_is_valid(buf));
-    AWS_PRECONDITION(AWS_MEM_IS_WRITABLE(src, len), "Input array [src] must be readable up to [len] bytes.");
```
JonathanHenson: NIT: I see that you're just fixing a bug but should we check the destination is writable?

graebm: Assuming `AWS_PRECONDITION(aws_byte_buf_is_valid(buf));` is checking the validity of that argument

PR #585, opened 2020-02-04
Added return code to fixture functions so errors can be handled durin…

JonathanHenson:
…g setup and shutdown.

### DIFF
--- a/include/aws/testing/aws_test_harness.h
+++ b/include/aws/testing/aws_test_harness.h
@@ -311,9 +311,9 @@ static int total_failures;
         }                                                                                                              \
     } while (0)
 
-typedef void(aws_test_before_fn)(struct aws_allocator *allocator, void *ctx);
+typedef int(aws_test_before_fn)(struct aws_allocator *allocator, void *ctx);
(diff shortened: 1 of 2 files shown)


graebm:
> Just out of curiosity, what necessitated this?

No one was using the built in fixture setup/teardown functionality because they returned void, so you couldn't use `ASSERT_SUCCESS()` macros. You had to `AWS_FATAL_ASSERT()`, or just not bother checking errors.

We got into a debate just now where we couldn't decide whether the teardown function should:
- always run
- only run if setup and test were successful
- only run if setup was successful

ugh. Any testing experts out there?

graebm:
> ugh. Any testing experts out there?

Talking about this I think we all agreed that the teardown function should run even if the test failed. Not 100% clear whether it should run if the setup failed. Here's my proposal:



Teardown always runs, has function signature that lets it know if something went wrong in setup or test:
`int teardown_fn(int error_code)`

That lets people use teardown for stuff that always needs to happen, but also let's them put common checks in there that can be skipped if we already know something went wrong.

PR #586, opened 2020-02-04
Bigint

bretambrose:
* Bigint part 1 - init/cleanup/comparisons

### DIFF
--- a/include/aws/common/bigint.h
+++ b/include/aws/common/bigint.h
@@ -1,0 +1,97 @@
+#ifndef AWS_COMMON_BIGINT_H
+#define AWS_COMMON_BIGINT_H
+
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/common.h>
+
+#include <aws/common/array_list.h>
+#include <aws/common/byte_buf.h>
+
+/*
+ *
+ */
+struct aws_bigint {
+    /*
+     * A sequence of base 2^32 digits starting from the least significant
+     */
+    struct aws_array_list digits;
+
+    /*
+     * 1 = positive, -1 = negative
+     */
+    int sign;
+};
+
+AWS_EXTERN_C_BEGIN
+
+AWS_COMMON_API
+void aws_bigint_clean_up(struct aws_bigint *bigint);
+
+AWS_COMMON_API
+int aws_bigint_init(struct aws_bigint *bigint, struct aws_allocator *allocator);
+
+AWS_COMMON_API
+int aws_bigint_init_reserve(struct aws_bigint *bigint, struct aws_allocator *allocator, uint64_t reserve_bits);
+
+AWS_COMMON_API
+int aws_bigint_init_from_hex(
+    struct aws_bigint *bigint,
+    struct aws_allocator *allocator,
+    struct aws_byte_cursor hex_digits);
+
+AWS_COMMON_API
+int aws_bigint_init_from_int64(struct aws_bigint *bigint, struct aws_allocator *allocator, int64_t value);
+
+AWS_COMMON_API
+int aws_bigint_init_from_uint64(struct aws_bigint *bigint, struct aws_allocator *allocator, uint64_t value);
+
+AWS_COMMON_API
+int aws_bigint_bytebuf_append_as_hex(struct aws_bigint *bigint, struct aws_byte_buf *buffer);
+
+AWS_COMMON_API
+bool aws_bigint_is_negative(struct aws_bigint *bigint);
+
+AWS_COMMON_API
+bool aws_bigint_is_positive(struct aws_bigint *bigint);
+
+AWS_COMMON_API
+bool aws_bigint_is_zero(struct aws_bigint *bigint);
+
+AWS_COMMON_API
+bool aws_bigint_equals(struct aws_bigint *lhs, struct aws_bigint *rhs);
+
+AWS_COMMON_API
+bool aws_bigint_not_equals(struct aws_bigint *lhs, struct aws_bigint *rhs);
+
+AWS_COMMON_API
+bool aws_bigint_less_than(struct aws_bigint *lhs, struct aws_bigint *rhs);
+
+AWS_COMMON_API
+bool aws_bigint_less_than_or_equals(struct aws_bigint *lhs, struct aws_bigint *rhs);
+
+AWS_COMMON_API
+bool aws_bigint_greater_than(struct aws_bigint *lhs, struct aws_bigint *rhs);
+
+AWS_COMMON_API
+bool aws_bigint_greater_than_or_equals(struct aws_bigint *lhs, struct aws_bigint *rhs);
+
+AWS_COMMON_API
+void aws_bigint_negate(struct aws_bigint *bigint);
+
+AWS_EXTERN_C_END
+
+#endif /* AWS_COMMON_BIGINT_H */
--- a/include/aws/common/bigint.h
+++ b/include/aws/common/bigint.h
@@ -22,7 +22,9 @@
 #include <aws/common/byte_buf.h>
 
 /*
- *
+ * A basic big integer implementation using 2^32 as the base.  Algorithms used are formalizations of the basic
+ * grade school operations everyone knows and loves (as formalized in AoCP Vol 2, 4.3.1).  Current use case
+ * targets do not yet involve a domain large enough that its worth exploring more complex algorithms.
  */
 struct aws_bigint {
     /*
--- a/include/aws/common/bigint.h
+++ b/include/aws/common/bigint.h
@@ -26,6 +26,9 @@ struct aws_bigint;
 AWS_EXTERN_C_BEGIN
 
 AWS_COMMON_API
+bool aws_bigint_is_valid(const struct aws_bigint *bigint);
+
+AWS_COMMON_API
 void aws_bigint_destroy(struct aws_bigint *bigint);
 
 /**
--- a/include/aws/common/bigint.h
+++ b/include/aws/common/bigint.h
@@ -43,54 +43,99 @@ AWS_EXTERN_C_BEGIN
 AWS_COMMON_API
 void aws_bigint_clean_up(struct aws_bigint *bigint);
 
-AWS_COMMON_API
-int aws_bigint_init(struct aws_bigint *bigint, struct aws_allocator *allocator);
-
-AWS_COMMON_API
-int aws_bigint_init_reserve(struct aws_bigint *bigint, struct aws_allocator *allocator, uint64_t reserve_bits);
-
+/**
+ * Initializes a big int from a string of hex digits.  String may start with "0x".  Leading zeros are skipped.
+ * An empty string is considered an error.  A leading (-) symbol is not supported.  Use aws_bigint_negate() after
+ * calling aws_bigint_init_from_hex() to generate an arbitrary negative number.
+ */
 AWS_COMMON_API
 int aws_bigint_init_from_hex(
     struct aws_bigint *bigint,
     struct aws_allocator *allocator,
     struct aws_byte_cursor hex_digits);
 
+/**
+ * Initializes a big int from a 64 bit signed integer
+ */
 AWS_COMMON_API
 int aws_bigint_init_from_int64(struct aws_bigint *bigint, struct aws_allocator *allocator, int64_t value);
 
+/**
+ * Initializes a big int from a 64 bit unsigned integer
+ */
 AWS_COMMON_API
 int aws_bigint_init_from_uint64(struct aws_bigint *bigint, struct aws_allocator *allocator, uint64_t value);
 
+/**
+ * Initializes a big int as a copy of another big int
+ */
 AWS_COMMON_API
-int aws_bigint_bytebuf_append_as_hex(struct aws_bigint *bigint, struct aws_byte_buf *buffer);
+int aws_bigint_init_from_copy(struct aws_bigint *bigint, const struct aws_bigint *source);
 
+/**
+ * Writes a bigint to a buffer as a hexadecimal number.  Will prepend (-) in front of negative numbers for
+ * easier testing.  This API is primarily intended for testing.  Actual output (to various formats/bases) is TBD.
+ */
 AWS_COMMON_API
-bool aws_bigint_is_negative(struct aws_bigint *bigint);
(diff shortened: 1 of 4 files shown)


include/aws/common/bigint.h
```diff
@@ -0,0 +1,99 @@
...
+int aws_bigint_init_from_int64(struct aws_bigint *bigint, struct aws_allocator *allocator, int64_t value);
+
+AWS_COMMON_API
+int aws_bigint_init_from_uint64(struct aws_bigint *bigint, struct aws_allocator *allocator, uint64_t value);
+
+AWS_COMMON_API
+int aws_bigint_bytebuf_append_as_hex(struct aws_bigint *bigint, struct aws_byte_buf *buffer);
+
+AWS_COMMON_API
+bool aws_bigint_is_negative(struct aws_bigint *bigint);
```
justinboswell: According to the docs in the .c file, this should be const, and all functions that are just reading should be const, yes?

source/bigint.c
```diff
@@ -0,0 +1,353 @@
...
+    if (value >= 0) {
+        return aws_bigint_init_from_uint64(bigint, allocator, (uint64_t)value);
+    }
+
+    if (value == INT64_MIN) {
+        if (aws_bigint_init_from_uint64(bigint, allocator, INT64_MIN_AS_HEX)) {
+            return AWS_OP_ERR;
+        }
+    } else {
+        if (aws_bigint_init_from_uint64(bigint, allocator, (uint64_t)(-value))) {
```
justinboswell: I don't know that I understand why the negation is here, comment maybe?

include/aws/common/bigint.h
bretambrose: Yes.  I was thinking that there may be spots where I wanted to (temporarily) negate the inputs under certain conditions, but after doing addition and subtraction, that is unnecessary with a proper helper function, so I'm just going to tag all the operands with const.

source/bigint.c
bretambrose: Sure, just ensuring that negating the negative integer would be representable as a positive integer (pre uint64 cast).  Maybe that kind of care is not necessary if you understand the C standard and defined/undefined behavior better than I.

Will add comments.

include/aws/common/bigint.h
```diff
@@ -0,0 +1,99 @@
...
+/*
+ * A basic big integer implementation using 2^32 as the base.  Algorithms used are formalizations of the basic
+ * grade school operations everyone knows and loves (as formalized in AoCP Vol 2, 4.3.1).  Current use case
+ * targets do not yet involve a domain large enough that its worth exploring more complex algorithms.
+ */
+struct aws_bigint {
+    /*
+     * A sequence of base 2^32 digits starting from the least significant
+     */
+    struct aws_array_list digits;
```
graebm: why not 2^64?

include/aws/common/bigint.h
```diff
@@ -0,0 +1,99 @@
...
+int aws_bigint_init(struct aws_bigint *bigint, struct aws_allocator *allocator);
+
+AWS_COMMON_API
+int aws_bigint_init_reserve(struct aws_bigint *bigint, struct aws_allocator *allocator, uint64_t reserve_bits);
+
+AWS_COMMON_API
+int aws_bigint_init_from_hex(
+    struct aws_bigint *bigint,
+    struct aws_allocator *allocator,
+    struct aws_byte_cursor hex_digits);
```
graebm: Nice job accepting input that starts with "0x" but not requiring it.
Add a comment here explaining that.

include/aws/common/bigint.h
```diff
@@ -0,0 +1,99 @@
...
+
+AWS_COMMON_API
+int aws_bigint_init_from_int64(struct aws_bigint *bigint, struct aws_allocator *allocator, int64_t value);
+
+AWS_COMMON_API
+int aws_bigint_init_from_uint64(struct aws_bigint *bigint, struct aws_allocator *allocator, uint64_t value);
+
+AWS_COMMON_API
+int aws_bigint_bytebuf_append_as_hex(struct aws_bigint *bigint, struct aws_byte_buf *buffer);
```
graebm: Ought to be named `aws_byte_buf_append_as_hex(struct aws_byte_buf *buffer, const aws_bigint *bigint)`

Should there be a fn for querying the space required? like `size_t aws_bigint_length_as_hex()`?

source/bigint.c
```diff
@@ -0,0 +1,353 @@
...
+
+    while (digit_cursor.len > 0) {
+        char hex_digit = *digit_cursor.ptr;
+        uint32_t hex_value = 0;
+        if (hex_digit <= '9' && hex_digit >= '0') {
+            hex_value = hex_digit - '0';
+        } else if (hex_digit <= 'f' && hex_digit >= 'a') {
+            hex_value = hex_digit - 'a' + 10;
+        } else if (hex_digit <= 'F' && hex_digit >= 'A') {
+            hex_value = hex_digit - 'A' + 10;
```
graebm: Not sure if we care about the efficiency of this fn, or if it's just for testing, but I did a [crazy table-lookup](https://github.com/awslabs/aws-c-http/blob/31b5dd632e5b27c5d538bb7b1604322cdc6da6db/source/strutil.c#L17) for the `int aws_strutil_read_unsigned_hex(struct aws_byte_cursor cursor, uint64_t *dst)` function I built for aws-c-http

feel free to copy the table, or maybe we can move this fn to common, or this is just for testing and whooooo caaares

source/bigint.c
```diff
@@ -0,0 +1,353 @@
...
+        } else {
+            return AWS_OP_ERR;
```
graebm: return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);

source/bigint.c
```diff
@@ -0,0 +1,353 @@
...
+        return AWS_OP_ERR;
+    }
+
+    /*
+     * We don't support negative hex numbers from an initialization standpoint, but we still
+     * need to indicate the number's sign on output
+     */
+    if (bigint->sign < 0) {
+        buffer->buffer[buffer->len++] = '-';
+    }
```
graebm: Or we could let user write '-' themselves, just like they need to call negate() themselves when initializing.
Also gives flexibility for them to format something like "-0x01".

include/aws/common/bigint.h
bretambrose: 2^32 is the largest base where multiplying two digits can be done using primitive types available in current processors

include/aws/common/bigint.h
bretambrose: Going to temporarily change the function name to reflect the debug/test intent.  Will convert to a proper output function once I know the use cases it needs to support beyond testing.

source/bigint.c
bretambrose: I'd like to reuse it, but want to hold off for now on moving things across libraries

source/bigint.c
bretambrose: Deferring output functionality decisions by downgrading this to a (temporary) debug function.

include/aws/common/bigint.h
```diff
@@ -0,0 +1,144 @@
...
+struct aws_bigint {
+    /*
+     * A sequence of base 2^32 digits starting from the least significant
+     */
+    struct aws_array_list digits;
+
+    /*
+     * 1 = positive, -1 = negative
+     */
+    int sign;
```
JonathanHenson: why make this something so large? because packing makes it irrelevant? why not int8_t or something?

include/aws/common/bigint.h
```diff
@@ -0,0 +1,144 @@
...
+ */
+AWS_COMMON_API
+int aws_bigint_init_from_copy(struct aws_bigint *bigint, const struct aws_bigint *source);
+
+/**
+ * Writes a bigint to a buffer as a hexadecimal number.  Will prepend (-) in front of negative numbers for
+ * easier testing.  This API is primarily intended for testing.  Actual output (to various formats/bases) is TBD.
+ */
+AWS_COMMON_API
+int aws_bigint_bytebuf_debug_output(const struct aws_bigint *bigint, struct aws_byte_buf *buffer);
```
JonathanHenson: maybe just aws_bigint_bytebuf_to_string?

include/aws/common/bigint.h
```diff
@@ -0,0 +1,144 @@
...
+AWS_COMMON_API
+bool aws_bigint_greater_than_or_equals(const struct aws_bigint *lhs, const struct aws_bigint *rhs);
+
+/**
+ * Negates the supplied bigint.  Has no effect on a zero-valued integer.
+ */
+AWS_COMMON_API
+void aws_bigint_negate(struct aws_bigint *bigint);
+
+AWS_EXTERN_C_END
```
JonathanHenson: no arithmetic operations?

source/bigint.c
```diff
@@ -0,0 +1,362 @@
...
+ * (2) params may be temporarily altered during computation before being restored in order to prevent pointless copying
+ *   (most often around sign negation and operator switching)
+ *
+ * Dev ideas:
+ *
+ * (1) Internal arithmetic ops may use a "raw" interface to allow for subsequence operations used in multiply/divide.
+ */
+
+void aws_bigint_clean_up(struct aws_bigint *bigint) {
+    aws_array_list_clean_up(&bigint->digits);
```
JonathanHenson: make this idempotent?

source/bigint.c
```diff
@@ -0,0 +1,362 @@
...
+    if (hex_cursor->len >= 2) {
+        const char *raw_ptr = (char *)hex_cursor->ptr;
+        if (raw_ptr[0] == '0' && (raw_ptr[1] == 'x' || raw_ptr[1] == 'X')) {
+            aws_byte_cursor_advance(hex_cursor, 2);
+        }
+    }
+}
+
+static void s_advance_cursor_to_non_zero(struct aws_byte_cursor *hex_cursor) {
+    while (hex_cursor->len > 0 && *hex_cursor->ptr == '0') {
```
JonathanHenson: memchar?

source/bigint.c
```diff
@@ -0,0 +1,362 @@
...
+        aws_byte_cursor_advance(hex_cursor, 1);
+    }
+}
+
+static int s_uint32_from_hex(struct aws_byte_cursor digit_cursor, uint32_t *output_value) {
+    AWS_FATAL_ASSERT(digit_cursor.len <= NIBBLES_PER_DIGIT);
+
+    *output_value = 0;
+
+    while (digit_cursor.len > 0) {
```
JonathanHenson: moving this to an array would be an obvious speed improvement.

source/bigint.c
```diff
@@ -0,0 +1,362 @@
...
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+
+    /* skip past leading zeros */
+    s_advance_cursor_to_non_zero(&hex_digits);
+    if (hex_digits.len == 0) {
+        return aws_bigint_init_from_uint64(bigint, allocator, 0);
+    }
+
+    uint64_t digit_count = hex_digits.len / BASE_BITS + 1;
```
JonathanHenson: what aboutm 0x01ab?

wouldn't digit count end up being 2 instead of 3?

include/aws/common/bigint.h
bretambrose: could change

include/aws/common/bigint.h
bretambrose: That'd be a good name for the final Api, would take the output base as well.

include/aws/common/bigint.h
bretambrose: Separate reviews.  For example https://github.com/awslabs/aws-c-common/pull/588

source/bigint.c
bretambrose: Oops, should be "(hex_digits.len - 1) / NIBBLES_PER_DIGIT + 1" actually.  Every 8 hex digits is one of our digits, so 0-8 -> 1, 9-16->2, etc...

source/bigint.c
bretambrose: Yeah, Mike pointed out to since it already exists in aws-c-io.  Waiting to get everything completed before starting a cross-library change.

source/bigint.c
bretambrose: I don't know of a not-equal version, so don't see how that would help

source/bigint.c
bretambrose: Sure

include/aws/common/bigint.h
```diff
@@ -0,0 +1,144 @@
...
+
+#include <aws/common/array_list.h>
+#include <aws/common/byte_buf.h>
+
+/*
+ * A basic big integer implementation using 2^32 as the base.  Algorithms used are formalizations of the basic
+ * grade school operations everyone knows and loves (as formalized in AoCP Vol 2, 4.3.1).  Current use case
+ * targets do not yet involve a domain large enough that its worth exploring more complex algorithms.
+ */
+struct aws_bigint {
```
feliperodri: What about `aws_bigint_is_valid()`? What are the set of invariants for this data structure? For example,
```
bool aws_bigint_is_valid(const struct aws_bigint * bigint) {
/* Pointer validity checks */
    if (!bigint) {
        return false;
    }
    bool digits_is_valid = aws_array_list_is_valid(&bigint->digits);
    bool sign_valid = (sign == 1 || sign == -1);
    return digits_is_valid && sign_valid;
}
```
It could be check as pre- and postconditions in every function of this module.

source/bigint.c
```diff
@@ -0,0 +1,363 @@
...
+ * (1) Internal arithmetic ops may use a "raw" interface to allow for subsequence operations used in multiply/divide.
+ */
+
+void aws_bigint_clean_up(struct aws_bigint *bigint) {
+    aws_array_list_clean_up(&bigint->digits);
+    AWS_ZERO_STRUCT(bigint->digits);
+}
+
+static void s_advance_cursor_past_hex_prefix(struct aws_byte_cursor *hex_cursor) {
+    if (hex_cursor->len >= 2) {
```
feliperodri: if `hex_cursor == NULL`, then NULL pointer deference, which could be avoided here by adding at the beginning `AWS_PRECONDITION(aws_byte_cursor_is_valid(hex_cursor))`.

include/aws/common/bigint.h
bretambrose: Implementation now internal

include/aws/common/bigint.h
bretambrose: Added an is_valid function and pre condition checks.  Post condition could be added on demand whenever proofs get added.

source/bigint.c
bretambrose: Should the check be here or is it sufficient to put a precondition on the cursor at the entry point of the public function that uses this? (how much redundancy?)

include/aws/common/bigint.h
feliperodri: Nice! You didn't added ` bool sign_valid = (sign == 1 || sign == -1);`, isn't necessary anymore?

source/bigint.c
feliperodri: If you added to the entry point, there is no need to add it here as well; however, you must make sure that all entry points that call `s_advance_cursor_past_hex_prefix` contains the `PRECONDITION`. AFAIK, the asserts are only triggered in debug mode (all preconditions are treated as assertions), would it be so bad to add it in both places?

bretambrose: Should be fine, this isn't really performance sensitive.

include/aws/common/bigint.h
bretambrose: Actually I went and re-read the comment and updated is_valid to include it.

PR #587, opened 2020-02-05
CMake package should add the modules to CMAKE_MODULE_PATH

Lectem:
Issue: https://github.com/awslabs/aws-c-event-stream/issues/15 

Right now `aws-c-event-stream` needs a lot of boilerplate code to determine the modules installation path while the `aws-c-common` package should provide it.
This would fix https://github.com/awslabs/aws-c-event-stream/issues/15 simply by moving `find_package(aws-c-common REQUIRED)` before the various module `include`s
Using `configure_package_config_file` makes it easier to have a relocatable package.

A pull request will be opened in `aws-c-event-stream` too.


Lectem:
Any news on this ?
I'm waiting for this PR and https://github.com/awslabs/aws-c-event-stream/pull/33 to be merged so that we can set our CI back to upstream and get the latest updates.

JonathanHenson:
I think this is a great change, but we'd need to apply it evenly across all the repos (there's 8 more), so it will take time. It's also a lot of work when we get the cmake stuff wrong to get them upgraded. We're evaluating this and we'll apply it next time we touch our cmake. 

As for the linked issue, it looks like your code is out of date. The actual bug was fixed a while ago. 

Are you having problems with the latest commit?

Lectem:
> As for the linked issue, it looks like your code is out of date. The actual bug was fixed a while ago.
> 
> Are you having problems with the latest commit?

Sorry I missed the question last time.
Yes I was having issues when building on amazon linux 2, using the latest commits at the time. That would be 60d5f8de6d4e237c7e7ac2014266d4948116f9e3.

Should I try with latest master ?

PR #591, opened 2020-02-06
Bigint multiply

bretambrose:
Adds aws_bigint_multiply()


include/aws/common/bigint.h
```diff
@@ -153,6 +153,13 @@ int aws_bigint_add(struct aws_bigint *output, struct aws_bigint *lhs, struct aws
 AWS_COMMON_API
 int aws_bigint_subtract(struct aws_bigint *output, struct aws_bigint *lhs, struct aws_bigint *rhs);
 
+/*
+ * Multiplies two big integers, placing the result in output.  Output must have been initialized first.  Output
+ * may alias to either operand.
+ */
+AWS_COMMON_API
+int aws_bigint_multiply(struct aws_bigint *output, struct aws_bigint *lhs, struct aws_bigint *rhs);
```
graebm: lhs and rhs should be const

source/bigint.c
```diff
@@ -559,3 +559,79 @@ int aws_bigint_subtract(struct aws_bigint *output, struct aws_bigint *lhs, struc
 
     return AWS_OP_SUCCESS;
 }
+
+int aws_bigint_multiply(struct aws_bigint *output, struct aws_bigint *lhs, struct aws_bigint *rhs) {
+    struct aws_bigint temp_output;
+    AWS_ZERO_STRUCT(temp_output);
```
graebm: optional optimization idea:
If `output` is NOT an alias of `lhs` or `rhs`, then you don't need to use `temp_output`

PR #597, opened 2020-02-20
Move to pragma once, clang modules support, turn inlining off by defa…

JonathanHenson:


danielsn:
Do we have profiling data that the `inline static` declarations significantly improve performance? Maybe we should rip off the bandaid and just use the normal C calling style.

PR #598, opened 2020-02-21
ARM fixes

justinboswell:

### DIFF
--- a/cmake/AwsFeatureTests.cmake
+++ b/cmake/AwsFeatureTests.cmake
@@ -45,7 +45,7 @@ int main() {
     __asm__ __volatile__(\"\":\"=r\"(foo):\"r\"(bar):\"memory\");
 }" AWS_HAVE_GCC_INLINE_ASM)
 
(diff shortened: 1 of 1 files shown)


cmake/AwsFeatureTests.cmake
```diff
@@ -45,7 +45,8 @@ int main() {
     __asm__ __volatile__(\"\":\"=r\"(foo):\"r\"(bar):\"memory\");
 }" AWS_HAVE_GCC_INLINE_ASM)
 
-if(NOT LEGACY_COMPILER_SUPPORT)
+string(REGEX MATCH "^(aarch64|arm)" ARM_CPU ${CMAKE_SYSTEM_PROCESSOR})
+if(NOT LEGACY_COMPILER_SUPPORT OR ARM_CPU)
```
graebm: I'm not super familiar with all our helper cmake files.
Is there another more appropriate place where we should declare that new ARM_CPU variable?

justinboswell: I don't know either, it doesn't seem to be checked anywhere else.

PR #600, opened 2020-02-27
Temporarily disable the __CPROVER_{r,w}_ok macros

karkhaz:
This is while we work to resolve an issue with CBMC, described at
https://github.com/diffblue/cbmc/issues/5194

### DIFF
--- a/.cbmc-batch/include/proof_helpers/aws_byte_cursor_read_common.h
+++ b/.cbmc-batch/include/proof_helpers/aws_byte_cursor_read_common.h
@@ -20,14 +20,14 @@
 void aws_byte_cursor_read_common_harness() {
     /* parameters */
     struct aws_byte_cursor cur;
-    size_t length;
-    DEST_TYPE *dest = can_fail_malloc(length);
+    DEST_TYPE *dest = can_fail_malloc(sizeof(*dest));
 
     /* assumptions */
     ensure_byte_cursor_has_allocated_buffer_member(&cur);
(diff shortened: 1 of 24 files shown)


.cbmc-batch/include/proof_helpers/aws_byte_cursor_read_common.h
```diff
@@ -20,14 +20,14 @@
...
-    DEST_TYPE *dest = can_fail_malloc(length);
+    DEST_TYPE *dest = can_fail_malloc(sizeof(*dest));
 
     /* assumptions */
     ensure_byte_cursor_has_allocated_buffer_member(&cur);
     __CPROVER_assume(aws_byte_cursor_is_valid(&cur));
+    __CPROVER_assume(cur.len >= BYTE_WIDTH);
     __CPROVER_assume(AWS_MEM_IS_READABLE(cur.ptr, BYTE_WIDTH));
-    __CPROVER_assume(AWS_OBJECT_PTR_IS_WRITABLE(dest));
+    __CPROVER_assume(dest != NULL);
```
feliperodri: Instead of just replacing it, should we add also a comment? So we can put it back once the issue is fixed on CBMC?

feliperodri: Actually, now with the updated `AWS_OBJECT_PTR_IS_WRITABLE`, do we need to remove it from here?

PR #602, opened 2020-02-27
Allow cross-compiled tests via ALLOW_CROSS_COMPILED_TESTS

justinboswell:


graebm:
Would it be less confusing to just compile tests by default when cross-compiling?

JonathanHenson:
> Would it be less confusing to just compile tests by default when cross-compiling?

you usually can't run the tests when you're cross compiling

PR #603, opened 2020-03-03
Clock vtable support

bretambrose:


bretambrose:
You're right, I should just mock both functions in the cached provider.

PR #604, opened 2020-03-03
Add support for building with MinGW

trilader:
*Issue #, if available:*

*Description of changes:*
This PR makes it possible to build aws-c-common with the MinGW toolchain on Windows.

The main changes are adding checks for the `__MINGW32__` define which is preset on both 32 and 64 bit MinGW installations. Additionally MinGW needs changes related to format strings such as defining `__USE_MINGW_ANSI_STDIO` and using the `gnu_printf` attribute.

Building with MSVC still works as before these changes. The test pass on both MinGW and MSVC.

### DIFF
--- a/include/aws/common/byte_order.inl
+++ b/include/aws/common/byte_order.inl
@@ -19,11 +19,11 @@
 #include <aws/common/byte_order.h>
 #include <aws/common/common.h>
 
-#ifdef _MSC_VER
+#ifdef _WIN32
 #    include <stdlib.h>
 #else
 #    include <netinet/in.h>
-#endif /* _MSC_VER */
+#endif /* _WIN32 */
 
 AWS_EXTERN_C_BEGIN
 
@@ -49,7 +49,7 @@ AWS_STATIC_IMPL uint64_t aws_hton64(uint64_t x) {
     uint64_t v;
     __asm__("bswap %q0" : "=r"(v) : "0"(x));
     return v;
-#elif defined(_MSC_VER)
+#elif defined(_WIN32)
     return _byteswap_uint64(x);
 #else
     uint32_t low = (uint32_t)x;
@@ -69,7 +69,7 @@ AWS_STATIC_IMPL uint64_t aws_ntoh64(uint64_t x) {
  * Convert 32 bit integer from host to network byte order.
  */
 AWS_STATIC_IMPL uint32_t aws_hton32(uint32_t x) {
-#ifdef _MSC_VER
+#ifdef _WIN32
     return aws_is_big_endian() ? x : _byteswap_ulong(x);
 #else
     return htonl(x);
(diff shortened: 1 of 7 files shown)


include/aws/common/byte_order.inl
```diff
@@ -19,7 +19,7 @@
 #include <aws/common/byte_order.h>
 #include <aws/common/common.h>
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) || defined(__MINGW32__)
```
graebm, line 22: All of these `#if defined(_MSC_VER) || defined(__MINGW32__)` should actually be `#if defined(_WIN32)`

It's easy to mix up compiler/platform when a compiler is platform-specific like msvc. But the mix up does become clear when you see something like this

source/windows/environment.c
```diff
@@ -23,12 +23,16 @@ int aws_get_environment_value(
     const struct aws_string *variable_name,
     struct aws_string **value_out) {
 
-#pragma warning(push)
-#pragma warning(disable : 4996)
+#ifndef __MINGW32__
+#   pragma warning(push)
+#   pragma warning(disable : 4996)
```
graebm, line 28: please comment what warning this disables (and say why if it's not immediately obvious why one would disable it)

For example, this is how we usually disable some MSVC-specific warnings
```
#if _MSC_VER
#    pragma warning(disable : 4221) /* aggregate initializer using local variable addresses */
#    pragma warning(disable : 4204) /* non-constant aggregate initializer */
#endif
```

trilader: This (and the same construct in `system_info.c`) only looks like it's a newly disabled warning because of indentation changes. It's actually just disabling the MSVC specific #pragma when using MinGW as GCC can't understand MSVC warning codes.

include/aws/common/byte_order.inl
trilader: Good point. I've addressed this in the new version of the commit.

trilader:
I've rebased to current master in oder to trigger the CI again but it looks like it's still not working correctly

ihnorton:
Bump. As I mentioned over in https://github.com/aws/aws-sdk-cpp/issues/1426, we are interested in support for the mingw-w64 target (as are others like @nealrichardson).

@graebm, I see you are now using github actions: would you accept a PR to add a mingw-w64 target to the build matrix?

jeroen:
FYI, we have adopted a modified version of this patch in msys2 [mingw-w64-aws-sdk-cpp](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-aws-sdk-cpp). To install binaries from msys2 use 


```sh
pacman -Syu
pacman -S mingw-w64-{i686,x86_64}-aws-sdk-cpp
```

PR #609, opened 2020-03-17
min/max/sub functions for math.h

graebm:
Some Trivial math functions I was sick of not having.

Should I bother with the min/max functions for esoteric types like float/double/signed? Currently math.h only has u32/u64/size_t stuff in it, but the min/max functions were easy to write, so I added everything reasonable I could think of.

I didn't do an assembly version of the sub functions (the add functions have this) because I wanted to invest 1hr into this, not 8. If you want to implement them, you get a free promotion to SDE2.

### DIFF
--- a/include/aws/common/math.h
+++ b/include/aws/common/math.h
@@ -95,6 +95,28 @@ AWS_COMMON_MATH_API uint32_t aws_add_u32_saturating(uint32_t a, uint32_t b);
 AWS_COMMON_MATH_API int aws_add_u32_checked(uint32_t a, uint32_t b, uint32_t *r);
 
 /**
+ * Subtracts a - b. If the result overflows returns 0.
+ */
+AWS_STATIC_IMPL uint64_t aws_sub_u64_saturating(uint64_t a, uint64_t b);
+
+/**
+ * If a - b overflows, returns AWS_OP_ERR; otherwise subtracts
+ * a - b, returns the result in *r, and returns AWS_OP_SUCCESS.
+ */
+AWS_STATIC_IMPL int aws_sub_u64_checked(uint64_t a, uint64_t b, uint64_t *r);
+
+/**
+ * Subtracts a - b. If the result overflows returns 0.
+ */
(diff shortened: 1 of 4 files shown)


include/aws/common/math.h
```diff
@@ -132,6 +165,31 @@ AWS_STATIC_IMPL bool aws_is_power_of_two(const size_t x);
...
+AWS_STATIC_IMPL int64_t aws_min_i64(int64_t a, int64_t b);
+AWS_STATIC_IMPL int64_t aws_max_i64(int64_t a, int64_t b);
+AWS_STATIC_IMPL size_t aws_min_size(size_t a, size_t b);
+AWS_STATIC_IMPL size_t aws_max_size(size_t a, size_t b);
+AWS_STATIC_IMPL int aws_min_int(int a, int b);
+AWS_STATIC_IMPL int aws_max_int(int a, int b);
+AWS_STATIC_IMPL float aws_min_float(float a, float b);
+AWS_STATIC_IMPL float aws_max_float(float a, float b);
+AWS_STATIC_IMPL double aws_min_double(double a, double b);
+AWS_STATIC_IMPL double aws_max_double(double a, double b);
```
justinboswell: Shouldn't these be named ```aws_int_max```/```aws_int_min``` etc, to match all the rest of our APIs?

graebm: I was matching these:
```
aws_mul_u64_saturating()
aws_mul_u32_saturating()
aws_mul_size_saturating()
...
```

include/aws/common/math.inl
```diff
@@ -150,6 +196,102 @@ AWS_STATIC_IMPL int aws_round_up_to_power_of_two(size_t n, size_t *result) {
 #    pragma warning(pop)
 #endif /* _MSC_VER */
 
+AWS_STATIC_IMPL uint8_t aws_min_u8(uint8_t a, uint8_t b) {
```
danielsn: These are all basically the same. Why not have a macro that generates them?

graebm: I could see both ways. Going to keep it this way for the sake of simplicity/readability.

PR #613, opened 2020-03-26
ARN parsing and serialization added

DavidOgunsAWS:
*Issue #, if available:*

*Description of changes:*
* One new structure 'aws_resource_name' representing an ARN defined here: https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html
* Methods to parse ARNs from strings and serialize to strings. aws_parse_resource_name / aws_resource_name_to_str 

### DIFF
--- a/include/aws/common/resource_name.h
+++ b/include/aws/common/resource_name.h
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2010-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+#pragma once
+
+#include <aws/common/byte_buf.h>
+#include <aws/common/common.h>
+
+struct aws_resource_name {
+    struct aws_byte_cursor partition;
+    struct aws_byte_cursor service;
+    struct aws_byte_cursor region;
+    struct aws_byte_cursor account_id;
+    struct aws_byte_cursor resource_id;
+};
+
+AWS_EXTERN_C_BEGIN
+
+/**
+    Given an ARN "Amazon Resource Name" represented as an in memory a
+    structure representing the parts
+*/
+AWS_COMMON_API
+int aws_resource_name_init_from_cur(struct aws_resource_name *arn, const struct aws_byte_cursor *string);
+
+/**
+    Calculates the space needed to write an ARN to a byte buf
+*/
+AWS_COMMON_API
+int aws_resource_name_length(const struct aws_resource_name *arn, size_t *size);
+
+/**
+    Serializes an ARN structure into the lexical string format
+*/
+AWS_COMMON_API
+int aws_byte_buf_append_resource_name(struct aws_byte_buf *buf, const struct aws_resource_name *arn);
+
+AWS_EXTERN_C_END
--- a/source/resource_name.c
+++ b/source/resource_name.c
@@ -21,7 +21,7 @@
 static const char ARN_DELIMETER[] = ":";
 static const char ARN_DELIMETER_CHAR = ':';
 
-static const size_t DELIMETER_LEN = strlen("arn:::::");
+static const size_t DELIMETER_LEN = 8; /* strlen("arn:::::") */
 
 AWS_COMMON_API
 int aws_resource_name_init_from_cur(struct aws_resource_name *arn, const struct aws_byte_cursor *input) {
--- a/source/resource_name.c
+++ b/source/resource_name.c
@@ -15,16 +15,18 @@
 
 #include <aws/common/resource_name.h>
 
+#define ARN_SPLIT_COUNT ((size_t)5)
+#define ARN_PARTS_COUNT ((size_t)6)
+
 static const char ARN_DELIMETER[] = ":";
 static const char ARN_DELIMETER_CHAR = ':';
 
 AWS_COMMON_API
 int aws_resource_name_init_from_cur(struct aws_resource_name *arn, const struct aws_byte_cursor *input) {
-    const size_t split_count = 5;
-    struct aws_byte_cursor arn_parts[split_count + 1];
+    struct aws_byte_cursor arn_parts[ARN_PARTS_COUNT];
     struct aws_array_list arn_part_list;
-    aws_array_list_init_static(&arn_part_list, arn_parts, split_count + 1, sizeof(struct aws_byte_cursor));
-    if (aws_byte_cursor_split_on_char_n(input, ARN_DELIMETER_CHAR, split_count, &arn_part_list)) {
+    aws_array_list_init_static(&arn_part_list, arn_parts, ARN_PARTS_COUNT, sizeof(struct aws_byte_cursor));
+    if (aws_byte_cursor_split_on_char_n(input, ARN_DELIMETER_CHAR, ARN_SPLIT_COUNT, &arn_part_list)) {
         return aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
     }
 
--- a/source/resource_name.c
+++ b/source/resource_name.c
@@ -21,6 +21,8 @@
 static const char ARN_DELIMETER[] = ":";
 static const char ARN_DELIMETER_CHAR = ':';
 
+static const size_t DELIMETER_LEN = strlen("arn:::::");
+
 AWS_COMMON_API
 int aws_resource_name_init_from_cur(struct aws_resource_name *arn, const struct aws_byte_cursor *input) {
     struct aws_byte_cursor arn_parts[ARN_PARTS_COUNT];
@@ -44,7 +46,7 @@ int aws_resource_name_init_from_cur(struct aws_resource_name *arn, const struct 
     if (aws_array_list_get_at(&arn_part_list, &arn->region, 3)) {
         return aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
     }
-    if (aws_array_list_get_at(&arn_part_list, &arn->account_id, 4)) {
+    if (aws_array_list_get_at(&arn_part_list, &arn->account_id, 4) || aws_byte_cursor_eq_c_str(&arn->account_id, "")) {
         return aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
     }
     if (aws_array_list_get_at(&arn_part_list, &arn->resource_id, 5)) {
@@ -61,7 +63,8 @@ int aws_resource_name_length(const struct aws_resource_name *arn, size_t *size) 
     AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->account_id));
     AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->resource_id));
 
-    *size = arn->partition.len + arn->region.len + arn->service.len + arn->account_id.len + arn->resource_id.len + 8;
+    *size = arn->partition.len + arn->region.len + arn->service.len + arn->account_id.len + arn->resource_id.len +
+            DELIMETER_LEN;
 
     return AWS_OP_SUCCESS;
 }
--- a/source/resource_name.c
+++ b/source/resource_name.c
@@ -27,6 +27,12 @@ int aws_resource_name_init_from_cur(struct aws_resource_name *arn, const struct 
     if (aws_byte_cursor_split_on_char_n(input, ARN_DELIMETER_CHAR, split_count, &arn_part_list)) {
         return aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
     }
+
+    struct aws_byte_cursor *arn_prefix;
+    if (aws_array_list_get_at_ptr(&arn_part_list, (void **)&arn_prefix, 0) ||
+        !aws_byte_cursor_eq_c_str(arn_prefix, "arn")) {
+        return aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
+    }
     if (aws_array_list_get_at(&arn_part_list, &arn->partition, 1)) {
         return aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
     }
--- a/source/resource_name.c
+++ b/source/resource_name.c
@@ -21,9 +21,9 @@ const char ARN_DELIMETER_CHAR = ':';
 AWS_COMMON_API
 int aws_resource_name_init_from_cur(struct aws_resource_name *arn, const struct aws_byte_cursor *input) {
     const size_t split_count = 5;
-    struct aws_byte_cursor arn_parts[split_count+1];
+    struct aws_byte_cursor arn_parts[split_count + 1];
     struct aws_array_list arn_part_list;
-    aws_array_list_init_static(&arn_part_list, arn_parts, split_count+1, sizeof(struct aws_byte_cursor));
+    aws_array_list_init_static(&arn_part_list, arn_parts, split_count + 1, sizeof(struct aws_byte_cursor));
     if (aws_byte_cursor_split_on_char_n(input, ARN_DELIMETER_CHAR, split_count, &arn_part_list)) {
         return aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
     }
--- a/source/resource_name.c
+++ b/source/resource_name.c
@@ -15,32 +15,18 @@
 
 #include <aws/common/resource_name.h>
 
-#define ARN_DELIMETER ":"
-
-static int s_parse_resource_name_part(
-    struct aws_byte_cursor *token_cursor,
-    const struct aws_byte_cursor *cursor_split_on,
-    struct aws_byte_cursor *prev_token_cursor) {
-    *prev_token_cursor = *token_cursor;
-    aws_byte_cursor_advance(prev_token_cursor, 1);
-
-    if (aws_byte_cursor_find_exact(prev_token_cursor, cursor_split_on, token_cursor)) {
-        return aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
-    }
-    prev_token_cursor->len = token_cursor->ptr - prev_token_cursor->ptr;
-    return AWS_OP_SUCCESS;
-}
+const char ARN_DELIMETER[] = ":";
+const char ARN_DELIMETER_CHAR = ':';
 
 AWS_COMMON_API
 int aws_resource_name_init_from_cur(struct aws_resource_name *arn, const struct aws_byte_cursor *input) {
-    const size_t part_count = 6;
-    struct aws_byte_cursor arn_parts[6];
+    const size_t split_count = 5;
+    struct aws_byte_cursor arn_parts[split_count+1];
     struct aws_array_list arn_part_list;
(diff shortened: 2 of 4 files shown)


source/resource_name.c
```diff
@@ -0,0 +1,122 @@
...
+    }
+    if ((errc = s_parse_resource_name_part(&cursor, &cursor_split_on, &out->region))) {
+        return aws_raise_error(errc);
+    }
+    if ((errc = s_parse_resource_name_part(&cursor, &cursor_split_on, &out->account_id))) {
+        return aws_raise_error(errc);
+    }
+
+    out->resource_id.ptr = cursor.ptr + 1;
+    out->resource_id.len = in->len - 8 -
```
bretambrose: 8 is a bit opaque, could comment "arn:" + 4 ":"s or whatever it is

source/resource_name.c
```diff
@@ -0,0 +1,122 @@
...
+        (out->partition.len + out->service.len + out->region.len + out->account_id.len);
```
bretambrose: Is an empty resource id valid?

source/resource_name.c
```diff
@@ -0,0 +1,122 @@
...
+#include <aws/common/resource_name.h>
+
+#include <aws/common/common.h>
+#include <aws/common/byte_buf.h>
+
+#define ARN_DELIMETER ":"
+
+static int s_parse_resource_name_part(struct aws_byte_cursor *token_cursor, const struct aws_byte_cursor *cursor_split_on, struct aws_byte_cursor *prev_token_cursor) {
+    prev_token_cursor->ptr = token_cursor->ptr + 1;
+    prev_token_cursor->len = token_cursor->len;
```
bretambrose: Shouldn't this be len - 1?  Alternatively you could assign it with token_cursor and advance 1

source/resource_name.c
```diff
@@ -0,0 +1,122 @@
...
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->service));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->region));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->account_id));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->resource_id));
+
+    const size_t arn_text_length = arn->partition.len +
+            arn->region.len +
+            arn->service.len +
+            arn->account_id.len +
+            arn->resource_id.len + 8;
```
bretambrose: Magic 8 again.

source/resource_name.c
```diff
@@ -0,0 +1,122 @@
...
+
+    if (buf->capacity < arn_text_length) {
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
```
bretambrose: Either add an expected size function or don't require the buffer to be long enough and instead use append_dynamic

source/resource_name.c
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    struct aws_byte_cursor prefix = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("arn:");
```
bretambrose: could put "arn:" into a static string literal at this point

source/resource_name.c
```diff
@@ -0,0 +1,122 @@
...
+    const struct aws_byte_buf colon = aws_byte_buf_from_c_str(ARN_DELIMETER);
+    struct aws_byte_cursor colon_cur;
+
+    aws_byte_cursor_read(&prefix, buf, 4);
```
bretambrose: replace all the reads with append_dynamic imo

source/resource_name.c
```diff
@@ -0,0 +1,122 @@
...
+
+    out->resource_id.ptr = cursor.ptr + 1;
+    out->resource_id.len = in->len - 8 -
+        (out->partition.len + out->service.len + out->region.len + out->account_id.len);
+
+    return AWS_OP_SUCCESS;
+}
+
+AWS_COMMON_API
+int aws_resource_name_to_str(struct aws_resource_name *arn, struct aws_byte_buf *buf) {
```
bretambrose: arn can be const

source/resource_name.c
```diff
@@ -0,0 +1,122 @@
...
+        return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
+    }
+
+    struct aws_byte_cursor prefix = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("arn:");
+    const struct aws_byte_buf colon = aws_byte_buf_from_c_str(ARN_DELIMETER);
+    struct aws_byte_cursor colon_cur;
+
+    aws_byte_cursor_read(&prefix, buf, 4);
+    aws_byte_cursor_read(&arn->partition, buf, arn->partition.len);
+    colon_cur = aws_byte_cursor_from_buf(&colon);
```
bretambrose: can just set this once if you append_dynamic

source/resource_name.c
```diff
@@ -0,0 +1,122 @@
...
+    aws_byte_cursor_read(&colon_cur, buf, colon.len);
+
+    aws_byte_cursor_read(&arn->account_id, buf, arn->account_id.len);
+    colon_cur = aws_byte_cursor_from_buf(&colon);
+    aws_byte_cursor_read(&colon_cur, buf, colon.len);
+
+    aws_byte_cursor_read(&arn->resource_id, buf, arn->resource_id.len);
+
+    return AWS_OP_SUCCESS;
+}
```
bretambrose: add a newline or some platforms will cry

tests/resource_name_test.c
```diff
@@ -0,0 +1,85 @@
...
+    ASSERT_TRUE(aws_byte_cursor_eq_c_str(&arn_02.service, "cloudformation"));
+    ASSERT_TRUE(aws_byte_cursor_eq_c_str(&arn_02.region, "us-east-1"));
+    ASSERT_TRUE(aws_byte_cursor_eq_c_str(&arn_02.account_id, "1234567890"));
+    ASSERT_TRUE(aws_byte_cursor_eq_c_str(&arn_02.resource_id, "stack/FooBar"));
+
+    return AWS_OP_SUCCESS;
+}
+
+AWS_TEST_CASE(parse_resource_name_failures_test, s_test_parse_resource_name_failures)
+static int s_test_parse_resource_name_failures(struct aws_allocator *allocator, void *ctx) {
```
bretambrose: I'd make sure you hit every single failure if-block possibility here

source/resource_name.c
DavidOgunsAWS: I happened to encounter the non _dynamic version first. Both make sense for what might be useful. Should there be two versions of aws_byte_buf_append_resource_name() ? One for each strategy?

bretambrose: I'd go with one or the other.  You could also precalculate exactly how much room you need and then do a one-time reserve before you start appending.

source/resource_name.c
```diff
@@ -0,0 +1,119 @@
...
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->account_id));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->resource_id));
+
+    const struct aws_byte_cursor prefix = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("arn:");
+    const struct aws_byte_cursor colon_cur = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL(ARN_DELIMETER);
+
+    const size_t arn_text_length =
+        arn->partition.len + arn->region.len + arn->service.len + arn->account_id.len + arn->resource_id.len + 8;
+
+    if (buf->capacity < arn_text_length) {
```
bretambrose: This doesn't account for existing buffer contents.

source/resource_name.c
```diff
@@ -0,0 +1,119 @@
...
+        if (!aws_allocator_is_valid(buf->allocator)) {
+            return aws_raise_error(AWS_ERROR_DEST_COPY_TOO_SMALL);
+        }
+        if (aws_mem_realloc(buf->allocator, (void **)&buf->buffer, buf->capacity, arn_text_length)) {
```
bretambrose: Better to use aws_byte_buf_reserve_relative

tests/resource_name_test.c
```diff
@@ -0,0 +1,128 @@
...
+    AWS_ZERO_STRUCT(too_small_buffer);
+    ASSERT_SUCCESS(aws_byte_buf_init(&too_small_buffer, allocator, 16));
+    struct aws_resource_name arn_03 = {.partition = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("aws-cn"),
+                                       .service = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("dynamodb"),
+                                       .region = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("cn-northwest-1"),
+                                       .account_id = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("123456789"),
+                                       .resource_id = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("Table/Books")};
+    ASSERT_SUCCESS(aws_byte_buf_append_resource_name(&too_small_buffer, &arn_03));
+    ASSERT_TRUE(aws_byte_buf_eq_c_str(&too_small_buffer, "arn:aws-cn:dynamodb:cn-northwest-1:123456789:Table/Books"));
+    aws_byte_buf_clean_up(&too_small_buffer);
```
bretambrose: Since you've added support for static buffers, I'd do a failure check with a static buffer that's too short too.

tests/resource_name_test.c
```diff
@@ -0,0 +1,128 @@
...
+static int s_test_parse_resource_name(struct aws_allocator *allocator, void *ctx) {
+    (void)allocator;
+    (void)ctx;
+
+    struct aws_byte_cursor arn_string_01 =
+        AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("arn:aws-us-gov:iam::123456789012:user/*");
+    struct aws_resource_name arn_01;
+    AWS_ZERO_STRUCT(arn_01);
+    ASSERT_SUCCESS(aws_resource_name_init_from_cur(&arn_01, &arn_string_01));
+    ASSERT_TRUE(aws_byte_cursor_eq_c_str(&arn_01.partition, "aws-us-gov"));
```
justinboswell: We also have AWS_ASSERT_BIN_ARRAY_EQUALS, your call.

tests/resource_name_test.c
```diff
@@ -0,0 +1,128 @@
...
+
+    return AWS_OP_SUCCESS;
+}
+
+AWS_TEST_CASE(parse_resource_name_failures_test, s_test_parse_resource_name_failures)
+static int s_test_parse_resource_name_failures(struct aws_allocator *allocator, void *ctx) {
+    (void)allocator;
+    (void)ctx;
+
+    struct aws_byte_cursor arn_string_01 = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("arn:aws-us-gov:iam::123456789012");
```
justinboswell: Can you add a comment to each of these that explains why they should fail?

include/aws/common/resource_name.h
```diff
@@ -0,0 +1,43 @@
...
+    structure representing the parts
+*/
+AWS_COMMON_API
+int aws_resource_name_init_from_cur(struct aws_resource_name *arn, const struct aws_byte_cursor *string);
+
+/**
+    Serializes an ARN structure into the lexical string format
+*/
+AWS_COMMON_API
+int aws_byte_buf_append_resource_name(struct aws_byte_buf *buf, const struct aws_resource_name *arn);
```
justinboswell: Should we just have a resource_name -> buf function? Maybe in addition to this one? Especially for like debug printing these or interpolating them into JSON or something.

JonathanHenson: should we have a mode on this API that will give the user a length on how much memory to allocate so they don't have to use a dynamically resizing buffer? I don't like making allocation decisions for users we don't have to make.

source/resource_name.c
```diff
@@ -0,0 +1,119 @@
...
+    return AWS_OP_SUCCESS;
+}
+
+AWS_COMMON_API
+int aws_resource_name_init_from_cur(struct aws_resource_name *arn, const struct aws_byte_cursor *input) {
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(input));
+    AWS_PRECONDITION(input->ptr);
+    AWS_PRECONDITION(arn);
+
+    const struct aws_byte_cursor cursor_split_on = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL(ARN_DELIMETER);
```
JonathanHenson: This whole thing seems unnecessarily complicated. We have a split api. Why not?

1. split on the delimiter
2. Verify the split count == 5
3. Verify first split is "arn"
4. Assign each split segment to the appropriate resource name part
5. done

source/resource_name.c
JonathanHenson: buf->capacity - buf->len

source/resource_name.c
JonathanHenson: let's make up our mind. Either use append_dynamic, use the reserve api, or not do this at all and provide an api that will tell the user the string length before calling this api and then just hard fail if it isn't be enough.

tests/resource_name_test.c
```diff
@@ -0,0 +1,128 @@
...
+    struct aws_resource_name arn_01;
+    AWS_ZERO_STRUCT(arn_01);
+    ASSERT_FAILS(aws_resource_name_init_from_cur(&arn_01, &arn_string_01));
+
+    struct aws_byte_cursor arn_string_02 = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("arn:aws-us-gov:iam:");
+    struct aws_resource_name arn_02;
+    AWS_ZERO_STRUCT(arn_02);
+    ASSERT_FAILS(aws_resource_name_init_from_cur(&arn_02, &arn_string_02));
```
JonathanHenson: ASSERT_ERROR with a specific error code

tests/resource_name_test.c
```diff
@@ -0,0 +1,128 @@
...
+    struct aws_byte_buf too_small_buffer;
+    AWS_ZERO_STRUCT(too_small_buffer);
+    ASSERT_SUCCESS(aws_byte_buf_init(&too_small_buffer, allocator, 16));
+    struct aws_resource_name arn_03 = {.partition = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("aws-cn"),
+                                       .service = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("dynamodb"),
+                                       .region = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("cn-northwest-1"),
+                                       .account_id = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("123456789"),
+                                       .resource_id = AWS_BYTE_CUR_INIT_FROM_STRING_LITERAL("Table/Books")};
+    ASSERT_SUCCESS(aws_byte_buf_append_resource_name(&too_small_buffer, &arn_03));
+    ASSERT_TRUE(aws_byte_buf_eq_c_str(&too_small_buffer, "arn:aws-cn:dynamodb:cn-northwest-1:123456789:Table/Books"));
```
JonathanHenson: try to avoid ASSERT_TRUE as it makes test failure output less useful.
ASSERT_BIN_ARRAY_EQUALS and the like would be useful here.

tests/resource_name_test.c
JonathanHenson: no, ASSERT_TRUE shouldn't be used if it can be avoided. It makes test output less useful, because you can't see what actually happened.

tests/resource_name_test.c
```diff
@@ -0,0 +1,128 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/byte_buf.h>
+#include <aws/common/resource_name.h>
+#include <aws/testing/aws_test_harness.h>
+
```
JonathanHenson: how about a negative test case on nested ":"?

as in:

arn::bla:h::helad etc...

source/resource_name.c
DavidOgunsAWS: Yup, good catch! I'll also add a test for this

source/resource_name.c
DavidOgunsAWS: Went with aws_byte_buf_reserve_relative() and did this.

tests/resource_name_test.c
DavidOgunsAWS: Changed all of the c_str_equals to ASSERT_BIN_ARRAY_EQUALS.

source/resource_name.c
DavidOgunsAWS: I was avoiding the edge case around the there not always being 5 parts to an arn and my initial understanding of aws_array_list was that it would trigger unnecessary allocations. Once the resource ID part is found at the end, the rest of the characters are in it.

source/resource_name.c
DavidOgunsAWS: Going to go with letting users know the length up front and failing if there isn't enough space during writes.

source/resource_name.c
JonathanHenson: Array list has a static mode. That will not perform any allocations.

tests/resource_name_test.c
DavidOgunsAWS: Is this not a valid ARN where:

partition = ""
service = "bla"
region = "h"
account_id = ""
resource_id = "helad"

While this is quite likely forever not a real ARN, it doesn't seem like the CRT should have logic that understand valid values in each part of an ARN other than the prefix.

tests/resource_name_test.c
DavidOgunsAWS: Fixed

source/resource_name.c
DavidOgunsAWS: I went with not doing any allocations, in favor of the caller taking care of it however they see fit and failing if they have not provided enough space.

Read the impl o aws_byte_cursor_split_on_char_n() and it's doing what I need. Going to change it up one more time.

include/aws/common/resource_name.h
DavidOgunsAWS: > Should we just have a resource_name -> buf function? Maybe in addition to this one? Especially for like debug printing these or interpolating them into JSON or something.
I wouldn't be sure where to anchor a resource_name -> byte_buf function quite yet. Without being aware of the the use case details yet we can add this later.

> should we have a mode on this API that will give the user a length on how much memory to allocate so they don't have to use a dynamically resizing buffer? I don't like making allocation decisions for users we don't have to make.
Removed all resizing and allocations from functions.

tests/resource_name_test.c
DavidOgunsAWS: Done

source/resource_name.c
```diff
@@ -0,0 +1,118 @@
...
+
+AWS_COMMON_API
+int aws_resource_name_length(const struct aws_resource_name *arn, size_t *size) {
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->partition));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->service));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->region));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->account_id));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&arn->resource_id));
+
+    *size = arn->partition.len + arn->region.len + arn->service.len + arn->account_id.len + arn->resource_id.len + 8;
```
justinboswell: Is the ```+8``` for colons and arn:? If so, can you add a comment to that effect? Or store it in a ```const size_t delimeter_len = strlen("arn:::::");```

tests/resource_name_test.c
justinboswell: I believe it is invalid for account_id to be empty, right? Could we just validate emptiness?

PR #618, opened 2020-04-02
aws_hash_combine()

graebm:
We have hashes for single items.
This lets us combine them.

### DIFF
--- a/source/hash_table.c
+++ b/source/hash_table.c
@@ -992,12 +992,11 @@ uint64_t aws_hash_ptr(const void *item) {
 }
 
 uint64_t aws_hash_combine(uint64_t item1, uint64_t item2) {
-    /* Use high and low halves of item2 as input to hashlittle */
-    uint32_t *b = &((uint32_t *)&item2)[0];
-    uint32_t *c = &((uint32_t *)&item2)[1];
+    uint32_t b = (uint32_t)item2;
+    uint32_t c = (uint32_t)(item2 >> 32);
 
-    hashlittle2(&item1, sizeof(item1), c, b);
-    return item2;
+    hashlittle2(&item1, sizeof(item1), &c, &b);
+    return ((uint64_t)b << 32) | c;
 }
 
 bool aws_hash_callback_c_str_eq(const void *a, const void *b) {
--- a/source/hash_table.c
+++ b/source/hash_table.c
@@ -992,8 +992,8 @@ uint64_t aws_hash_ptr(const void *item) {
 }
 
 uint64_t aws_hash_combine(uint64_t item1, uint64_t item2) {
-    uint32_t b = (uint32_t)item2;
-    uint32_t c = (uint32_t)(item2 >> 32);
+    uint32_t b = item2 & 0xFFFFFFFF; /* LSB */
+    uint32_t c = item2 >> 32;        /* MSB */
 
     hashlittle2(&item1, sizeof(item1), &c, &b);
     return ((uint64_t)b << 32) | c;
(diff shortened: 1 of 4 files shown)


tests/hash_table_test.c
```diff
@@ -1258,3 +1258,39 @@ static int s_test_hash_table_byte_cursor_create_find_fn(struct aws_allocator *al
...
+        ASSERT_TRUE(c_bytes[i] != 0);
+    }
+
+    /* Check even low-entropy numbers result in something with high entropy */
+    c = aws_hash_combine(0, 1);
+    for (size_t i = 0; i < sizeof(c); ++i) {
+        ASSERT_TRUE(c_bytes[i] != 0);
+    }
+
+    /* Hash should NOT be cummutative */
```
danielsn, line 1292: commutative

tests/hash_table_test.c
```diff
@@ -1258,3 +1258,39 @@ static int s_test_hash_table_byte_cursor_create_find_fn(struct aws_allocator *al
...
+    /* Using all bytes, right? */
+    uint8_t *c_bytes = (uint8_t *)&c;
+    for (size_t i = 0; i < sizeof(c); ++i) {
+        ASSERT_TRUE(c_bytes[i] != 0);
+    }
+
+    /* Check even low-entropy numbers result in something with high entropy */
+    c = aws_hash_combine(0, 1);
+    for (size_t i = 0; i < sizeof(c); ++i) {
+        ASSERT_TRUE(c_bytes[i] != 0);
```
danielsn, line 1289: Can we do better than that? If we want entropy, we could actually measure entropy.

graebm: Yes we could do better, but this is not the test were we would do it.

We use `lookup3.inl`, an existing collection of functions for hashing. We've assumed so far that these do what they say. I just wanted to make sure I didn't do something dumb where the result never had its high or low bytes set.

tests/hash_table_test.c
graebm: derp. thanks

tests/hash_table_test.c
graebm: I just took this part of the test out. It was a naive test the quality of the underlying hash function, and this isn't the place for it

tests/hash_table_test.c
```diff
@@ -1258,3 +1258,39 @@ static int s_test_hash_table_byte_cursor_create_find_fn(struct aws_allocator *al
...
+    /* Same inputs gets same results, right? */
+    ASSERT_UINT_EQUALS(c, aws_hash_combine(a, b));
+
+    /* Using all bytes, right? */
+    uint8_t *c_bytes = (uint8_t *)&c;
+    for (size_t i = 0; i < sizeof(c); ++i) {
+        ASSERT_TRUE(c_bytes[i] != 0);
+    }
+
+    /* Check even low-entropy numbers result in something with high entropy */
```
justinboswell: If we wanted to test this, wouldn't we just fuzz it?

graebm: I just took this part of the test out. This is not the place to test the quality of the underlying hash library

source/hash_table.c
```diff
@@ -991,6 +991,14 @@ uint64_t aws_hash_ptr(const void *item) {
     return ((uint64_t)b << 32) | c;
 }
 
+uint64_t aws_hash_combine(uint64_t item1, uint64_t item2) {
+    uint32_t b = (uint32_t)item2;
```
feliperodri, line 995: This conversion could overflow, but it's intended, right? Also, would it make sense to replace it with the following line? 
```c
uint32_t b = item2 & 0xFFFFFFFF; /* LSB */
uint32_t c = item2 >> 32; /* MSB */
```

graebm: ok

PR #621, opened 2020-04-15
Android logcat logger

justinboswell:
Dependency of: https://github.com/awslabs/aws-crt-java/pull/182


source/android/logging.c
```diff
@@ -0,0 +1,187 @@
...
+     * Use this length for all but the last write, so we guarantee room for the newline even if we get truncated
+     */
+    size_t fake_total_length = formatting_data->total_length - 1;
+
+    if (current_index < fake_total_length) {
+        /*
+         * Add thread id and user content separator (" - ")
+         */
+        aws_thread_id_t current_thread_id = aws_thread_current_thread_id();
+        char thread_id[32];
```
JonathanHenson, line 60: Why not use the BUFSIZE macro here?

source/logging.c
```diff
@@ -144,6 +144,13 @@ int aws_logger_init_standard(
     struct aws_allocator *allocator,
     struct aws_logger_standard_options *options) {
 
+#ifdef ANDROID
+    (void)options;
+    extern int aws_logger_init_logcat(
```
JonathanHenson, line 149: Slap the API macro on here, since we deliberately hide symbols otherwise?

source/android/logging.c
justinboswell: Which macro? I can't find anything with BUFSIZE in it.

source/logging.c
justinboswell: We don't need the API macro, since this is inside the actual exported API function. Am I missing something?

source/android/logging.c
```diff
@@ -0,0 +1,187 @@
...
+
+    va_end(format_args);
+
+    if (result != AWS_OP_SUCCESS) {
+        return AWS_OP_ERR;
+    }
+
+    /* ANDROID_LOG_VERBOSE = 2, ANDROID_LOG_FATAL = 7 */
+    const int prio = 0x8 - log_level;
+    __android_log_write(prio, "AWSCRT", buffer);
```
graebm, line 155: Seems more useful to send the actual subject to logcat, instead of "AWSCRT" for everything.

I've always dreamed of filtering different topics at different levels and that's built into logcat. You prevent people from doing that by classifying everything as "AWSCRT"

source/android/logging.c
```diff
@@ -0,0 +1,187 @@
...
+    size_t next_index = current_index + amount;
+    if (next_index > maximum) {
+        next_index = maximum;
+    }
+
+    return next_index;
+}
+
+/* Override this for Android, as time and log level are taken care of by logcat */
+static int s_logcat_format(struct logcat_format_data *formatting_data, va_list args) {
```
bretambrose, line 43: I feel like there could be a little bit of refactoring shared functionality between this and the default via a private header or something.  Might be more apparent as we add other platform logger implementations.

source/android/logging.c
```diff
@@ -0,0 +1,187 @@
...
+            return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        }
+
+        current_index = s_advance_and_clamp_index(current_index, written_count, fake_total_length);
+    }
+
+    /*
+     * End with a newline.
+     */
+    int newline_written_count =
```
bretambrose, line 113: Do we ever log stack traces?  We may want to add a truncation indicator given the 1k limit.  Or maybe bump it up to 2-4k?

source/android/logging.c
justinboswell: I will fix this in a follow up

source/android/logging.c
justinboswell: For sure. We could probably just have a flagset or something per platform that indicates which parts of the format are required.

source/android/logging.c
justinboswell: good call. 4K it is. Will fix in follow-up.

justinboswell:
https://github.com/awslabs/aws-c-common/pull/623

PR #622, opened 2020-04-15
Add aws_run_command()

singku-china:
*Issue #, if available:*
For process creds provider, we need to capture external executable's output.

*Description of changes:*
Though the code for posix and windows are almost the same, except the difference between `_popen _pclose` and `popen pclose`. I didn't merge them together in case later we will need more functionalities.

### DIFF
--- a/include/aws/common/process.h
+++ b/include/aws/common/process.h
@@ -21,6 +21,47 @@ AWS_EXTERN_C_BEGIN
 
 AWS_COMMON_API int aws_get_pid(void);
 
+struct aws_run_command_result {
+    struct aws_allocator *allocator;
+
+    /* return code from running the command. */
+    int ret_code;
+
+    /**
+     * captured stdout message from running the command,
+     * caller is responsible for releasing the memory.
+     */
+    struct aws_string *std_out;
+
+    /**
+     * captured stderr message from running the command,
+     * caller is responsible for releasing the memory.
+     * It's currently not implemented and the value will be set to NULL.
+     */
+    struct aws_string *std_err;
+};
+
+struct aws_run_command_options {
+    struct aws_allocator *allocator;
+
+    /* command path and commandline options of running that command. */
+    const char *command;
+};
+
+AWS_COMMON_API struct aws_run_command_result *aws_run_command_result_new(struct aws_allocator *allocator);
+
+AWS_COMMON_API void aws_run_command_result_destroy(struct aws_run_command_result *result);
+
+/**
+ * Currently this API is implemented using popen on Posix system and
+ * _popen on Windows to capture output from running a command. Note
+ * that popen only captures stdout, and doesn't provide an option to
+ * capture stderr. We will add more options, such as acquire stderr
+ * in the future so probably will alter the underlying implementation
+ * as well.
+ */
+AWS_COMMON_API struct aws_run_command_result *aws_run_command(struct aws_run_command_options options);
+
 AWS_EXTERN_C_END
 
 #endif /* AWS_COMMON_PROCESS_H */
--- a/include/aws/common/process.h
+++ b/include/aws/common/process.h
@@ -22,8 +22,6 @@ AWS_EXTERN_C_BEGIN
 AWS_COMMON_API int aws_get_pid(void);
 
 struct aws_run_command_result {
-    struct aws_allocator *allocator;
-
     /* return code from running the command. */
     int ret_code;
 
@@ -42,15 +40,13 @@ struct aws_run_command_result {
 };
 
 struct aws_run_command_options {
-    struct aws_allocator *allocator;
-
     /* command path and commandline options of running that command. */
     const char *command;
 };
 
-AWS_COMMON_API struct aws_run_command_result *aws_run_command_result_new(struct aws_allocator *allocator);
+AWS_COMMON_API int aws_run_command_result_init(struct aws_allocator *allocator, struct aws_run_command_result *result);
 
-AWS_COMMON_API void aws_run_command_result_destroy(struct aws_run_command_result *result);
+AWS_COMMON_API void aws_run_command_result_cleanup(struct aws_run_command_result *result);
 
 /**
  * Currently this API is implemented using popen on Posix system and
@@ -60,7 +56,10 @@ AWS_COMMON_API void aws_run_command_result_destroy(struct aws_run_command_result
  * in the future so probably will alter the underlying implementation
  * as well.
  */
-AWS_COMMON_API struct aws_run_command_result *aws_run_command(struct aws_run_command_options options);
+AWS_COMMON_API int aws_run_command(
+    struct aws_allocator *allocator,
+    struct aws_run_command_options *options,
+    struct aws_run_command_result *result);
 
 AWS_EXTERN_C_END
 
--- a/include/aws/common/string.inl
+++ b/include/aws/common/string.inl
@@ -14,8 +14,8 @@
  * express or implied. See the License for the specific language governing
  * permissions and limitations under the License.
  */
-
 #include <aws/common/string.h>
+#include <ctype.h>
 
 AWS_EXTERN_C_BEGIN
 /**
@@ -57,5 +57,14 @@ bool aws_c_string_is_valid(const char *str) {
      */
     return str && AWS_MEM_IS_READABLE(str, 1);
 }
+
+/**
+ * Evaluates if a char is a white character.
+ */
+AWS_STATIC_IMPL
+bool aws_char_is_space(uint8_t c) {
+    return isspace((int)c) != 0;
+}
+
 AWS_EXTERN_C_END
 #endif /* AWS_COMMON_STRING_INL */
--- a/source/posix/process.c
+++ b/source/posix/process.c
@@ -14,8 +14,6 @@
  */
 
 #include <aws/common/process.h>
-
-#include <sys/types.h>
 #include <unistd.h>
 
 int aws_get_pid(void) {
--- a/source/process_common.c
+++ b/source/process_common.c
@@ -49,7 +49,7 @@ int aws_run_command(
     FILE *output_stream;
     char output_buffer[MAX_BUFFER_SIZE];
     struct aws_byte_buf result_buffer;
-    bool success = false;
+    int ret = AWS_OP_ERR;
     if (aws_byte_buf_init(&result_buffer, allocator, MAX_BUFFER_SIZE)) {
         goto on_finish;
     }
@@ -84,12 +84,9 @@ int aws_run_command(
             goto on_finish;
         }
     }
-    success = true;
+    ret = AWS_OP_SUCCESS;
 
 on_finish:
     aws_byte_buf_clean_up_secure(&result_buffer);
-    if (success) {
-        return AWS_OP_SUCCESS;
-    }
-    return AWS_OP_ERR;
+    return ret;
 }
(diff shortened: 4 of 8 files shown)


include/aws/common/process.h
```diff
@@ -21,6 +21,8 @@ AWS_EXTERN_C_BEGIN
 
 AWS_COMMON_API int aws_get_pid(void);
 
+AWS_COMMON_API struct aws_string *aws_get_command_output(struct aws_allocator *allocator, const char *command);
```
bretambrose, line 24: I think something like aws_run_command() or similar would be a better name.  Getting the output is reasonably expected; the real action is running something.  Also, I'd do the standard int return and the output as an output parameter

source/windows/process.c
```diff
@@ -14,8 +14,46 @@
...
+#include <ctype.h>
 #include <process.h>
+#include <stdio.h>
+#define MAX_BUFFER_SIZE (2048)
 
 int aws_get_pid(void) {
     return _getpid();
 }
+
+static bool s_isspace(uint8_t c) {
```
bretambrose, line 27: We have isspace(uint8_t) in like 5 places in our code base, we should just add it to common somewhere.

include/aws/common/process.h
graebm, line 24: wouldn't we also want the exit code?

source/posix/process.c
```diff
@@ -14,10 +14,48 @@
...
 
 int aws_get_pid(void) {
     return (int)getpid();
 }
+
+static bool s_isspace(uint8_t c) {
+    return isspace((int)c) != 0;
+}
+
+struct aws_string *aws_get_command_output(struct aws_allocator *allocator, const char *command) {
```
bretambrose, line 33: The only difference between the two implementations looks like popen vs. _popen and pclose vs. _pclose.  Seems refactorable such that about 98% of the code is shared, with a pair of function pointers being the only platform specific parts.

include/aws/common/process.h
graebm: How does this deal with stderr vs stdout? Do we get both? stdout only?

source/posix/process.c
singku-china: Thought about this, but if later we introduce more functionalities, and if those functions have bigger differences platform wide, we may have to put them in separate files without shared block. Then this function and those (future) functions are going to be here and there.

include/aws/common/process.h
graebm: Maybe we should just rip off python with all of our hearts: https://docs.python.org/3/library/subprocess.html#subprocess.run
and make something like:
```
struct aws_process_completed *aws_process_run(struct aws_allocator *allocator, const struct aws_process_run_options *options);
```

`options` could be something that gets more complex as time goes on and we have more diverse requirements
`aws_process_completed` could start with just having `return_code` and `stdout` in it, but we could add things like `stderr` in the future when we need it

singku-china: popen only gives you stdout , no stderr.
I'll take a look at the python definition, thanks.

include/aws/common/process.h
```diff
@@ -21,6 +21,54 @@ AWS_EXTERN_C_BEGIN
 
 AWS_COMMON_API int aws_get_pid(void);
 
+struct aws_run_command_result {
+    struct aws_allocator *allocator;
+
+    /* error code when calling aws_run_command() */
+    int error_code;
```
graebm, line 28: don't have the AWS_ERROR code be part of the result.
The error-reporting pattern for functions that return a pointer is:
if NULL is returned then check `aws_last_error()` to find out what went wrong.

that means you'd need to be sure `aws_raise_error(XXXX)` is called in your function before NULL is returned

include/aws/common/process.h
```diff
@@ -21,6 +21,54 @@ AWS_EXTERN_C_BEGIN
...
+
+    /* return code from running the command, currently not used. */
+    int ret_code;
```
graebm, line 31: `plcose()`/`_pclose()` returns the exit code of the process

source/windows/process.c
```diff
@@ -14,8 +14,48 @@
...
+
+struct aws_run_command_result *aws_run_command(struct aws_run_command_options options) {
+    FILE *output_stream;
+    char output_buffer[MAX_BUFFER_SIZE];
+    struct aws_run_command_result *result = aws_run_command_result_new(options.allocator);
+    if (!result) {
+        return NULL;
+    }
+
+    output_stream = _popen(aws_string_c_str(options.command), "r");
```
graebm, line 34: [debatable] The posix & windows implementations of `aws_run_command()` are non-trivial and implemented twice. The only difference is the "_" in front of popen and pclose

I know it's always a judgement call with multi-platform code, but a single implementation with an ifdef'ed popen call seems much more practical

```
#ifdef _WIN32
output_stream = _popen(options.command, "r");
#else
output_stream = popen(options.command, "r");
#endif
```

source/process_common.c
```diff
@@ -0,0 +1,60 @@
...
+    result->allocator = allocator;
+    return result;
+}
+
+void aws_run_command_result_destroy(struct aws_run_command_result *result) {
+    if (!result) {
+        return;
+    }
+    aws_string_destroy(result->std_out);
+    aws_string_destroy(result->std_err);
```
graebm, line 37: aws_string_destroy_secure()?
Not sure if this is actually sensitive data or not. I just noticed you were using aws_byte_buf_clean_up_secure() to clean up the intermediate buffer in `aws_run_command()`.

include/aws/common/string.inl
```diff
@@ -57,5 +57,14 @@ bool aws_c_string_is_valid(const char *str) {
      */
     return str && AWS_MEM_IS_READABLE(str, 1);
 }
+
+/**
+ * Evaluates if a char is a white character.
+ */
+AWS_STATIC_IMPL
+bool aws_char_is_space(uint8_t c) {
+    return isspace((int)c) != 0;
```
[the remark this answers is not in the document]
singku-china: This is more for the convenience of a bool return predict function used in trim. `isspace` accepts a int and returns int.

include/aws/common/process.h
```diff
@@ -21,6 +21,54 @@ AWS_EXTERN_C_BEGIN
...
+    struct aws_string *command;
+};
+
+AWS_COMMON_API struct aws_run_command_result *aws_run_command_result_new(struct aws_allocator *allocator);
+
+AWS_COMMON_API void aws_run_command_result_destroy(struct aws_run_command_result *result);
+
+AWS_COMMON_API struct aws_run_command_options *aws_run_command_options_new(struct aws_allocator *allocator);
+
+AWS_COMMON_API void aws_run_command_options_destroy(struct aws_run_command_options *options);
```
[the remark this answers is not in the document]
singku-china: Thought about these, but added these for the concern that in the future, the options may grow and become complicated, some of the option members might point to heap memory, though callers can allocate/deallocate them one by one, they might want a place to init/destroy them in a more manageable way.

include/aws/common/process.h
singku-china: Good idea.!

source/windows/process.c
singku-china: I think I can move them to process_common.c, which makes more sense.

source/process_common.c
singku-china: Yes, secure.

tests/process_test.c
```diff
@@ -28,3 +28,26 @@ static int s_get_pid_sanity_check_test_fn(struct aws_allocator *allocator, void
...
+    ASSERT_NOT_NULL(result);
+
+    ASSERT_BIN_ARRAYS_EQUALS(
+        result->std_out->bytes, result->std_out->len, s_expected_output->bytes, s_expected_output->len);
+
+    aws_run_command_result_destroy(result);
+    return AWS_OP_SUCCESS;
+}
+
+AWS_TEST_CASE(run_command_test, s_run_command_test_fn)
```
bretambrose, line 53: Can we add a case that runs a command that generates an error?

source/process_common.c
```diff
@@ -0,0 +1,87 @@
...
+        while (!feof(output_stream)) {
+            if (fgets(output_buffer, MAX_BUFFER_SIZE, output_stream) != NULL) {
+                struct aws_byte_cursor cursor = aws_byte_cursor_from_c_str(output_buffer);
+                if (aws_byte_buf_append_dynamic(&result_buffer, &cursor)) {
+                    goto on_finish;
+                }
+            }
+        }
+#ifdef _WIN32
+        result->ret_code = _pclose(output_stream);
```
bretambrose, line 74: Looking at docs it seems like this is only ever -1 or 0.  Is there any way to get the exit code of what was ran instead (or in addition)?

source/windows/process.c
```diff
@@ -14,6 +14,7 @@
  */
 #include <aws/common/process.h>
 
+#include <aws/common/string.h>
```
bretambrose, line 17: Is this needed?

source/process_common.c
```diff
@@ -0,0 +1,87 @@
...
+    if (aws_byte_buf_init(&result_buffer, result->allocator, MAX_BUFFER_SIZE)) {
+        goto on_finish;
+    }
+
+    if (output_stream) {
+        while (!feof(output_stream)) {
+            if (fgets(output_buffer, MAX_BUFFER_SIZE, output_stream) != NULL) {
+                struct aws_byte_cursor cursor = aws_byte_cursor_from_c_str(output_buffer);
+                if (aws_byte_buf_append_dynamic(&result_buffer, &cursor)) {
+                    goto on_finish;
```
bretambrose, line 69: This will change given a change in the API (result as an output parameter), but just to point out, an error here will result in a non-null result with no error code and no output strings, which isn't very reasonable from a caller perspective.

source/windows/process.c
singku-china: Nope.

tests/process_test.c
singku-china: Will do

source/process_common.c
singku-china: It says `The pclose() function waits for the associated process to terminate; it returns the exit status of the command, as returned by wait4(2).`

include/aws/common/process.h
```diff
@@ -21,6 +21,47 @@ AWS_EXTERN_C_BEGIN
...
+    /**
+     * captured stderr message from running the command,
+     * caller is responsible for releasing the memory.
+     * It's currently not implemented and the value will be set to NULL.
+     */
+    struct aws_string *std_err;
+};
+
+struct aws_run_command_options {
+    struct aws_allocator *allocator;
```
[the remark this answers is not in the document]
singku-china: Are you suggesting something like
```
int aws_run_command(struct aws_allocator *allocator, struct aws_run_command_options options, struct aws_run_command_result *result)
```
?

source/process_common.c
```diff
@@ -0,0 +1,95 @@
...
+    struct aws_run_command_result *result) {
+
+    AWS_FATAL_ASSERT(allocator);
+    AWS_FATAL_ASSERT(options);
+    AWS_FATAL_ASSERT(result);
+
+    FILE *output_stream;
+    char output_buffer[MAX_BUFFER_SIZE];
+    struct aws_byte_buf result_buffer;
+    bool success = false;
```
bretambrose, line 52: Minor nit but I always like to just have the return value be a variable that starts AWS_OP_ERR and changes to AWS_OP_SUCCESS, which turns the "if (success) return AWS_OP_SUCCESS; return AWS_OP_ERR" statements into just "return result;"

PR #627, opened 2020-04-27
remove unnecessary uses of __CPROVER_allocated_memory

danielsn:
*Issue #, if available:*

*Description of changes:*
`__CPOVER_allocated_memory` was used to allow CBMC proofs to check that memory was zeroed before being released.  However, this affects the entire proof, not just the check at the end, and could cause us to miss a use-after-free. Use `#pragma` in a surgical fashion instead.

Supercedes https://github.com/awslabs/aws-c-common/pull/626

### DIFF
--- a/.cbmc-batch/jobs/aws_hash_table_clean_up/Makefile
+++ b/.cbmc-batch/jobs/aws_hash_table_clean_up/Makefile
@@ -12,15 +12,14 @@
 # limitations under the License.
 
 ###########
-#4: 24s
-#8: 50s
+#4: 60s
+#8: 1m50s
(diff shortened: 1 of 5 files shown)


danielsn:
The missing checkmarks are a github rate limit, not reflecting failing proofs.  In particular, the two proofs affected by this PR both pass
https://d2lrx8u4fp9azc.cloudfront.net/aws_hash_table_clean_up-20200427-201141/out/html/index.html
https://d2lrx8u4fp9azc.cloudfront.net/aws_string_destroy_secure-20200427-200629/out/html/index.html

PR #634, opened 2020-05-08
Caches

TingDaoK:
*Issue #, if available:*

*Description of changes:*

- revet [remove linked_hash_table](https://github.com/awslabs/aws-c-common/pull/490)

- FIFO/LIFO/LRU caches based on aws_linked_hash_table

- APIs for LRU cache is changed here

### DIFF
--- a/include/aws/common/cache.h
+++ b/include/aws/common/cache.h
@@ -25,10 +25,6 @@ struct aws_cache_vtable {
     int (*remove)(struct aws_cache *cache, const void *key);
     void (*clear)(struct aws_cache *cache);
     size_t (*get_element_count)(const struct aws_cache *cache);
-
-    /* LRU-only functions */
-    void *(*use_lru_element)(struct aws_cache *cache);
-    void *(*get_mru_element)(const struct aws_cache *cache);
 };
 
 /**
@@ -39,6 +35,8 @@ struct aws_cache {
     const struct aws_cache_vtable *vtable;
     struct aws_linked_hash_table table;
     size_t max_items;
+
+    void *impl;
 };
 
 /* Default implementations */
@@ -66,7 +64,7 @@ int aws_cache_find(struct aws_cache *cache, const void *key, void **p_value);
 
 /**
  * Puts `p_value` at `key`. If an element is already stored at `key` it will be replaced. If the cache is already full,
- * item will be removed based on the implementation.
+ * an item will be removed based on the cache policy.
  */
 AWS_COMMON_API
 int aws_cache_put(struct aws_cache *cache, const void *key, void *p_value);
--- a/include/aws/common/cache.h
+++ b/include/aws/common/cache.h
@@ -1,0 +1,94 @@
+#ifndef AWS_COMMON_CACHE_H
+#define AWS_COMMON_CACHE_H
+/*
+ * Copyright 2010-2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/linked_hash_table.h>
+
+struct aws_cache;
+
+struct aws_cache_vtable {
+    void (*clean_up)(struct aws_cache *cache);
+    int (*find)(struct aws_cache *cache, const void *key, void **p_value);
+    int (*put)(struct aws_cache *cache, const void *key, void *p_value);
+    int (*remove)(struct aws_cache *cache, const void *key);
+    void (*clear)(struct aws_cache *cache);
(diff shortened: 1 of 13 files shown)


include/aws/common/cache.h
```diff
@@ -0,0 +1,94 @@
...
+
+struct aws_cache_vtable {
+    int (*find)(struct aws_cache *cache, const void *key, void **p_value);
+    int (*put)(struct aws_cache *cache, const void *key, void *p_value);
+    int (*remove)(struct aws_cache *cache, const void *key);
+    void (*clear)(struct aws_cache *cache);
+    size_t (*get_element_count)(const struct aws_cache *cache);
+
+    /* LRU-only functions */
+    void *(*use_lru_element)(struct aws_cache *cache);
```
justinboswell, line 30: I wonder about adding these here, what if we put them on an LRU user data that was part of aws_cache:
* Add void* impl to aws_cache
* Create LRU vtable, set it as the impl for LRU cache
* Dispatch from the LRU vtable in the LRU specific calls

include/aws/common/linked_hash_table.h
```diff
@@ -0,0 +1,132 @@
...
+ * NULL.
+ *
+ * If any errors occur AWS_OP_ERR will be returned.
+ */
+AWS_COMMON_API
+int aws_linked_hash_table_find(struct aws_linked_hash_table *table, const void *key, void **p_value);
+
+/**
+ * Finds element in the table by key. If found, AWS_OP_SUCCESS will be
+ * returned. And move it to the back of the list.
```
justinboswell, line 79: will be returned and the item will be moved to the back of the list

include/aws/common/linked_hash_table.h
```diff
@@ -0,0 +1,132 @@
...
+ * If not found, AWS_OP_SUCCESS will be returned and *p_value will be
+ * NULL.
+ *
+ * Note: it will break the order!!
```
justinboswell, line 83: Note: this will change the order of elements

include/aws/common/linked_hash_table.h
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * returns number of elements in the table.
+ */
+AWS_COMMON_API
+size_t aws_linked_hash_table_get_element_count(const struct aws_linked_hash_table *table);
+
+/**
+ * Move the aws_linked_hash_table_node to the end of the list.
+ *
+ * Note: it will break the order!!
```
justinboswell, line 115: Same as above

include/aws/common/cache.h
```diff
@@ -0,0 +1,94 @@
...
+ * returned. If not found, AWS_OP_SUCCESS will be returned and *p_value will be NULL.
+ *
+ * If any errors occur AWS_OP_ERR will be returned.
+ */
+AWS_COMMON_API
+int aws_cache_find(struct aws_cache *cache, const void *key, void **p_value);
+
+/**
+ * Puts `p_value` at `key`. If an element is already stored at `key` it will be replaced. If the cache is already full,
+ * item will be removed based on the implementation.
```
justinboswell, line 69: an item will be removed based on the cache policy

include/aws/common/cache.h
TingDaoK: fixed

source/lru_cache.c
```diff
@@ -100,16 +107,18 @@ static void *s_lru_cache_get_mru_element(const struct aws_cache *cache) {
 
 void *aws_lru_cache_use_lru_element(struct aws_cache *cache) {
     AWS_ASSERT(cache);
```
justinboswell, line 109: Probably want to:
```c
AWS_PRECONDITION(cache);
AWS_PRECONDITION(cache->impl);
```
For both of these functions

TingDaoK: Got it. Should the APIs for aws_cache also do the similar thing?
`int aws_cache_find(struct aws_cache *cache, const void *key, void **p_value) {
    AWS_ASSERT(cache);
    return cache->vtable->find(cache, key, p_value);
}`
to
`int aws_cache_find(struct aws_cache *cache, const void *key, void **p_value) {
    AWS_PRECONDITION(cache);
    return cache->vtable->find(cache, key, p_value);
}`

To be honest, I don't know what's the difference between AWS_PRECONDITION and AWS_ASSERT....

include/aws/common/cache.h
```diff
@@ -0,0 +1,92 @@
...
+void aws_cache_base_default_clear(struct aws_cache *cache);
+size_t aws_cache_base_default_get_element_count(const struct aws_cache *cache);
+
+AWS_EXTERN_C_BEGIN
+/**
+ * Cleans up the cache. Elements in the cache will be evicted and cleanup
+ * callbacks will be invoked.
+ */
+AWS_COMMON_API
+void aws_cache_clean_up(struct aws_cache *cache);
```
graebm, line 54: Should be `aws_cache_destroy()`

include/aws/common/cache.h
```diff
@@ -0,0 +1,92 @@
...
+
+struct aws_cache;
+
+struct aws_cache_vtable {
+    int (*find)(struct aws_cache *cache, const void *key, void **p_value);
+    int (*put)(struct aws_cache *cache, const void *key, void *p_value);
+    int (*remove)(struct aws_cache *cache, const void *key);
+    void (*clear)(struct aws_cache *cache);
+    size_t (*get_element_count)(const struct aws_cache *cache);
+};
```
graebm, line 28: Also need a `void (*destroy)(struct aws_cache *cache)` entry.
Otherwise, there's no way for a concrete type to free anything in `impl`.

PR #635, opened 2020-05-09
Added aws_message_bus 

justinboswell:
Introduces aws_message_bus, a mechanism for generic async broadcast/delivery of messages.

The design goal here is to make sending as cheap as possible, and delivery happen in a background thread by default. This is for collecting stats, dumping HARs, and other async tooling, not for decision making or APIs (at least not at present). I have intentionally not replicated the problem areas of EBuses (return values, sync by default, held lock across dispatch, etc). In this model, listeners are responsible for their own thread safety. We do not want to replicate the "well, just do it quickly because we're holding this hot mutex"-by-convention model.

There is a synchronous mode because in-thread that could be a speed up. That said, another approach could be to have a home thread, and not do locking within that thread, like Qt does. I think that might increase the complexity uncomfortably though.

The addressing scheme is designed to be simple, requiring minimal parameterization of the bus itself. If we want to do hierarchical addresses, we should chain buses to each other rather than have complicated addressing. 

Example addressing/send use-case: https://github.com/awslabs/aws-c-common/pull/635#issuecomment-627500531


justinboswell:
> I like the direction this is going. I think there's some room to do some brainstorming on the "address" notion wrt our use cases and possibly adding more structure to it (opaque structure with variants, ref counting are a couple of ideas that I'm currently keen on).

My current plan is to make it always ```uint64_t``` and hash WHATEVER to make the address, for instance:
```c
/* used for connection manager, slots, handlers, etc */
struct resource_id_addr {
    enum aws_resource_type type;
    void *address;
};

/* This could absolutely be macro'ed */
const struct resource_id_addr addr = {
    .type = AWS_RESOURCE_CONNECTION_MANAGER,
    .address = connection_manager,
};
const uint64_t address = aws_hash_blob(&addr, sizeof(addr));
struct aws_stat_record record = {
    .type = AWS_CONNECTION_COUNT,
    .value = connection_manager->connection_count,
};
aws_message_bus_send_copy(g_stat_bus, address, &record, sizeof(record));
```

include/aws/common/message_bus.h
```diff
@@ -0,0 +1,86 @@
...
+int aws_message_bus_unsubscribe(struct aws_message_bus *bus, uintptr_t address, aws_message_bus_listener_fn *listener, void *user_data);
+
+/*
+ * Allocates a new message to be sent. The message bus owns this memory, and is responsible
+ * for freeing it when message delivery is complete. If an error occurs before the message
+ * is sent, it may be freed with aws_message_bus_destroy_message()
+ *
+ * Use this if you need to do more complicated serialization into the buffer than `memcpy()`
+ */
+int aws_message_bus_new_message(struct aws_message_bus *bus, size_t size, struct aws_byte_buf *msg);
```
bretambrose, line 66: I like the idea that the payload is opaque bytes but I still feel like the metadata (which is likely immutable and for my use cases, ref counted) doesn't have a good story.  I would hate to try and parse out a pointer to metadata out of a raw byte buf and I would not want to have to copy that data repeatedly (and being forced to do so would limit the kinds of things we could tag).  I wonder if metadata could be a separate optional component in addition to the payload.

include/aws/common/message_bus.h
```diff
@@ -0,0 +1,78 @@
...
+ * when it is time to clean any state up.
+ * Listeners are owned by the subscriber, and are no longer referenced by the bus
+ * once unsubscribed.
+ * Message delivery happens in a separate thread from sending, so listeners are
+ * responsible for their own thread safety
+ */
+struct aws_message_bus;
+
+/* Signature for listener callbacks */
+typedef void(aws_message_bus_listener_fn)(uintptr_t msg_type, void *msg_data, void *user_data);
```
graebm, line 37: 1) Let's settle on a term: `msg_type` vs `address`. Both terms are being used and it's confusing.

2) Probably worth defining `#define AWS_BUS_ADDRESS_ALL 0`?

include/aws/common/message_bus.h
```diff
@@ -0,0 +1,78 @@
...
+enum aws_message_bus_policy {
+    /* Message delivery is queued */
+    AWS_BUS_ASYNC,
+    /* Message delivery is immediate */
+    AWS_BUS_SYNC,
+};
+
+/* Creates and returns a new message bus */
+struct aws_message_bus *aws_message_bus_new(
+    struct aws_allocator *allocator, size_t buffer_size, enum aws_message_bus_policy policy);
```
graebm, line 48: I'd suggest _new() taking an _options struct. Can totally see more args being added in the future

include/aws/common/message_bus.h
```diff
@@ -0,0 +1,78 @@
...
+ * responsible for their own thread safety
+ */
+struct aws_message_bus;
+
+/* Signature for listener callbacks */
+typedef void(aws_message_bus_listener_fn)(uintptr_t msg_type, void *msg_data, void *user_data);
+
+enum aws_message_bus_policy {
+    /* Message delivery is queued */
+    AWS_BUS_ASYNC,
```
graebm, line 41: super debatable, just throwing the idea out there: Should we just call **everything** `aws_bus_xyz` instead of `aws_message_bus_xyz`?

include/aws/common/message_bus.h
```diff
@@ -0,0 +1,78 @@
...
+ * Frees a message buffer allocated by aws_message_bus_new_message. Only necessary if the
+ * message is never sent
+ */
+void aws_message_bus_destroy_message(struct aws_message_bus *bus, struct aws_byte_buf *msg);
+
+/*
+ * Sends a message to any listeners. msg_data should have been allocated by aws_message_bus_new_message
+ * and will be cleaned up after the message is delivered
+ */
+int aws_message_bus_send(struct aws_message_bus *bus, uintptr_t address, struct aws_byte_buf *msg);
```
graebm, line 84: We're forcing messages to be linearly laid out data. That's going to put restrictions on the user. Like, if you need to send a struct that contains a a string and an unrelated int, you either store the string in a fixed array, or you invent your own serialization format where you prefix the string with length.

Should we do like hashtable and let users set a destructor? So that any arbitrary struct/class/object/string can be passed?

I see how the setup you have here *is* ideal for sending *just* a string, it allows each message to have its own length, and stores them all in linear memory.

include/aws/common/message_bus.h
```diff
@@ -0,0 +1,78 @@
...
+    /* Message delivery is queued */
+    AWS_BUS_ASYNC,
+    /* Message delivery is immediate */
+    AWS_BUS_SYNC,
+};
+
+/* Creates and returns a new message bus */
+struct aws_message_bus *aws_message_bus_new(
+    struct aws_allocator *allocator, size_t buffer_size, enum aws_message_bus_policy policy);
+
```
graebm, line 49: Shouldn't there be an on_message_received function that the bus-user can define?

Like, if this is an async bus, when a message is sent, the bus-user probably wants to schedule a task on their event-loop, so they can drain the messages from the desired thread. Or if they're a worker thread like the logger, they'd wake up the thread.

include/aws/common/message_bus.h
justinboswell: Yeah, this is just something I need to fix. I need to do a serious normalization pass still.

include/aws/common/message_bus.h
justinboswell: Yeah, it's also going to be ```init```. There's no reason this has to be heap allocated.

include/aws/common/message_bus.h
justinboswell: We could do it that way. The main reason I didn't is I'm concerned about the cost of allocating those messages at high frequency. On the other hand, I suppose that's something we can optimize on a case by case basis?

include/aws/common/message_bus.h
justinboswell: Totally on board with that. Also, do we even like the name bus?

In previous iterations in my career this has been called:
* event_stream (bad idea in this case, overloads that name in our domain)
* message_stream (still not great, because it's not really a serial stream)
* message_queue (correct in the async case, not correct in the sync case)
* event_bus/EBus

DISCUSS!

include/aws/common/message_bus.h
justinboswell: there's a ```condition_variable``` notify on the async impl. No one is ever responsible for draining the queue in the 2 impls I have here. We could make a manual drain impl, for sure, and implement async in terms of that. Is there value there?

include/aws/common/message_bus.h
justinboswell: So are you saying you'd like to be able to pass an additional void* that doesn't get copied, and thus is guaranteed to live effectively forever, or until the message is delivered? We can totally do that, I'm just trying to make sure I understand.

source/message_bus.c
```diff
@@ -0,0 +1,311 @@
...
+
+        struct bus_listener *listener = AWS_CONTAINER_OF(node, struct bus_listener, list_node);
+        if (listener->deliver == callback && listener->user_data == user_data) {
+            aws_linked_list_remove(node);
+            break;
+        }
+    }
+    /* END CRITICAL SECTION */
+    aws_mutex_lock(&bus->slots.mutex);
+    return AWS_OP_SUCCESS;
```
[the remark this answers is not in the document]
justinboswell: I'm not sure we do, actually. That would sort of violate the pattern we use of "it is always safe to clean up, no matter what".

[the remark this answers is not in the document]
justinboswell: Per phone discussion: Make this function void, assert if hash table find fails

include/aws/common/message_bus.h
justinboswell: Actually, wait for the next revision, I might have fixed this for you after talking to Mike.

bretambrose: New api looks good.  Makes sense that things like tagging and filtering and stuff can be stats-specific and go in there rather than here.

include/aws/common/bus.h
```diff
@@ -0,0 +1,83 @@
...
+typedef void(aws_bus_listener_fn)(uint64_t address, const void *payload, void *user_data);
+
+/* Creates and returns a new message bus */
+int aws_bus_init(struct aws_bus *bus, struct aws_bus_options *options);
+
+/* Cleans up a message bus by notifying all listeners and releasing the bus memory */
+void aws_bus_clean_up(struct aws_bus *bus);
+
+/* Subscribes a listener to a message type */
+int aws_bus_subscribe(struct aws_bus *bus, uint64_t address, aws_bus_listener_fn *listener, void *user_data);
```
bretambrose, line 66: I think we also need a wild card subscription option.

justinboswell: We have it, subscribe to address 0, or AWS_BUS_ANY

PR #639, opened 2020-05-26
Moved use of GCC builtin intrinsics to their own math inline file

justinboswell:
Fixes downstream C++ compile failures on ARM with GCC


include/aws/common/math.inl
```diff
@@ -52,6 +52,10 @@ AWS_EXTERN_C_BEGIN
 #    endif /*  AWS_HAVE_GCC_OVERFLOW_MATH_EXTENSIONS */
 #endif     /*  defined(AWS_HAVE_GCC_OVERFLOW_MATH_EXTENSIONS) && (defined(__clang__) || !defined(__cplusplus)) */
 
+#if defined(__clang__) || defined(__GNUC__)
```
rccarper, line 55: Probably just my lack of familiarity with the structure here--but I was wondering how the paths that include math.gcc_overflow.inl and math.gcc_x64_asm.inl get the clz/ctz functionality since it looks like it's only in math.gcc_builtin.inl now?  (Or I guess if they still need it?)

PR #642, opened 2020-06-04
aws versions of isspace() isalpha() etc

graebm:
These functions are affected by the user's "locale".
Ex: German letter ß in ISO-8859-1 fits in 1 byte, and would pass `isalpha()`

We need safe versions of these functions that ignore the locale.

Downstream PRs:
https://github.com/awslabs/aws-c-io/pull/278
https://github.com/awslabs/aws-c-auth/pull/84

### DIFF
--- a/source/byte_buf.c
+++ b/source/byte_buf.c
@@ -1444,3 +1444,38 @@ static struct aws_byte_cursor s_null_terminator_cursor = AWS_BYTE_CUR_INIT_FROM_
 int aws_byte_buf_append_null_terminator(struct aws_byte_buf *buf) {
     return aws_byte_buf_append_dynamic(buf, &s_null_terminator_cursor);
 }
+
+bool aws_isalnum(uint8_t ch) {
+    return (ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z') || (ch >= '0' && ch <= '9');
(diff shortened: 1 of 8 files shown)


source/byte_buf.c
```diff
@@ -1444,3 +1444,38 @@ static struct aws_byte_cursor s_null_terminator_cursor = AWS_BYTE_CUR_INIT_FROM_
...
+        case 0x0A: /* '\n' - line feed */
+            return true;
+        case 0x0B: /* '\v' - vertical tab */
+            return true;
+        case 0x0C: /* '\f' - form feed */
+            return true;
+        case 0x0D: /* '\r' - carriage return */
+            return true;
+        default:
+            return false;
```
marcomagdy, line 1479: Curious why you used the hex value instead of the character right away? just like the previous functions (isalpha, etc.).
as in:
```C
case '\t':
case '\f':
//...
```

graebm: I wanted to show both representations, hex and character.
So one representation was going to end up in the comment, and the other representation was going to end up in the switch statement.

graebm: I didn't realize escaped versions of the more esoteric ones event existed until after I'd written the switch statement with the hex values

PR #646, opened 2020-06-11
Bigint updates - constant time and api changes

bretambrose:
* Updates Bigint interface to be more consistent
* Changes some Bigint constructors to constant-time
* Changes Bigint comparison to constant-time


source/bigint.c
```diff
@@ -19,13 +19,16 @@
 #define BITS_PER_BYTE 8
```
justinboswell, line 19: Shouldn't these lines just be flipped, and ```BYTES_PER_BASE_DIGIT = (BASE_BITS / BITS_PER_BYTE)```?

source/bigint.c
```diff
@@ -244,82 +303,69 @@ struct aws_bigint *aws_bigint_new_from_hex(struct aws_allocator *allocator, stru
...
-    AWS_PRECONDITION(allocator);
-
-    struct aws_bigint *bigint = aws_mem_calloc(allocator, 1, sizeof(struct aws_bigint));
+    struct aws_bigint *bigint = aws_bigint_new_zero(allocator, desired_bit_width);
     if (bigint == NULL) {
         return NULL;
     }
 
-    bigint->allocator = allocator;
+    for (size_t i = 0; i < source.len; ++i) {
```
justinboswell, line 334: Why not just loop from source.len - 1 to 0?

source/bigint.c
```diff
@@ -244,82 +303,69 @@ struct aws_bigint *aws_bigint_new_from_hex(struct aws_allocator *allocator, stru
...
-    size_t digit_count = upper_digit > 0 ? 2 : 1;
-    if (aws_array_list_init_dynamic(&bigint->digits, allocator, digit_count, sizeof(uint32_t))) {
-        goto on_error;
-    }
+        uint32_t current_digit = 0;
+        size_t digit_index = i / BYTES_PER_BASE_DIGIT;
+        aws_array_list_get_at(&bigint->digits, &current_digit, digit_index);
 
-    aws_array_list_push_back(&bigint->digits, &lower_digit);
+        current_digit |= ((uint32_t)octet << BITS_PER_BYTE * (i % BYTES_PER_BASE_DIGIT));
```
justinboswell, line 345: /* shift the octet into place in the current 32-bit integer */

source/bigint.c
```diff
@@ -448,254 +456,287 @@ int aws_bigint_bytebuf_debug_output(const struct aws_bigint *bigint, struct aws_
...
-            current_digit <<= BITS_PER_BYTE;
+    size_t bytes_to_write = byte_length;
+    if (bytes_to_write > digit_count * BYTES_PER_BASE_DIGIT) {
+        bytes_to_write = digit_count * BYTES_PER_BASE_DIGIT;
+    }
 
-            if (!should_write) {
-                should_write = digit_byte > 0 || j + 1 == BYTES_PER_BASE_DIGIT;
-            }
+    /* dumb/wasteful but simple */
```
justinboswell, line 496: I bet this actually optimizes really well

justinboswell: Actually, come to think of it, if this vectorizes, it's going to not be constant time with the C impl. Does that matter? Do we need to disable or force vectorization here?

source/bigint.c
```diff
@@ -448,254 +456,287 @@ int aws_bigint_bytebuf_debug_output(const struct aws_bigint *bigint, struct aws_
...
+    }
+
+    return is_zero;
+}
+
 bool aws_bigint_is_positive(const struct aws_bigint *bigint) {
-    AWS_PRECONDITION(aws_bigint_is_valid(bigint));
+    AWS_FATAL_PRECONDITION(s_aws_bigint_is_valid(bigint));
+
+    volatile uint8_t sign_one = 1 - ((bigint->sign >> 1) & 0x01);
```
justinboswell, line 543: why is this volatile?

source/bigint.c
```diff
@@ -448,254 +456,287 @@ int aws_bigint_bytebuf_debug_output(const struct aws_bigint *bigint, struct aws_
...
-    }
+void aws_bigint_negate(struct aws_bigint *bigint) {
+    AWS_FATAL_PRECONDITION(s_aws_bigint_is_valid(bigint));
+
+    volatile int8_t is_zero = s_aws_bigint_is_zero(bigint);
 
-    uint32_t digit = 0;
-    aws_array_list_get_at(&bigint->digits, &digit, 0);
+    /* 1 if the the number is 0; -1 if the number is non-zero */
+    volatile int8_t negation_factor = is_zero + is_zero - 1;
```
justinboswell, line 560: Why is this volatile?

source/bigint.c
```diff
@@ -448,254 +456,287 @@ int aws_bigint_bytebuf_debug_output(const struct aws_bigint *bigint, struct aws_
...
+    bigint->sign *= negation_factor; /* is multiplication kosher? */
```
justinboswell, line 561: Why wouldn't it be kosher? It's constant time too for [-1|1]

source/bigint.c
```diff
@@ -448,254 +456,287 @@ int aws_bigint_bytebuf_debug_output(const struct aws_bigint *bigint, struct aws_
...
+    volatile uint8_t *gt_flag,
+    volatile uint8_t *eq_flag) {
+    volatile uint8_t gt = *gt_flag;
+    volatile uint8_t eq = *eq_flag;
 
-bool aws_bigint_not_equals(const struct aws_bigint *lhs, const struct aws_bigint *rhs) {
-    AWS_PRECONDITION(aws_bigint_is_valid(lhs));
-    AWS_PRECONDITION(aws_bigint_is_valid(rhs));
+    /* most significant digit queries */
+    volatile int64_t lhs_msd = (int64_t)s_get_most_significant_digit_index(lhs);
```
justinboswell, line 645: Isn't this const? Why would these vars need to be volatile?

source/bigint.c
```diff
@@ -448,254 +456,287 @@ int aws_bigint_bytebuf_debug_output(const struct aws_bigint *bigint, struct aws_
...
-    }
-}
+    /* Reduces to lhs_msd if min_mask is all 0s, rhs_msd if min_mask is all 1s */
+    volatile int64_t min_msd = lhs_msd ^ ((lhs_msd ^ rhs_msd) & min_mask);
 
-bool aws_bigint_greater_than(const struct aws_bigint *lhs, const struct aws_bigint *rhs) {
-    AWS_PRECONDITION(aws_bigint_is_valid(lhs));
-    AWS_PRECONDITION(aws_bigint_is_valid(rhs));
+    /*
+     * Magnitude check loop
```
justinboswell, line 665: /* for each digit, aggregate gt and eq */ or something

source/bigint.c
bretambrose: Probably fine, found some stuff hinting that even 32, 32 -> 64 was not constant time on some archivectures (early arm).

source/bigint.c
bretambrose: Natural mistrust of iterating downwards with an unsigned counter + would have to invert the i references anyways.

source/bigint.c
bretambrose: This one needs to be constant-time.  Maybe throwing volatile at it will help.

source/bigint.c
bretambrose: wanted to prevent an early out that would skip the function call.

bretambrose: reworked slightly to evaluate both conditions into volatile and then & the two at the end.

source/bigint.c
bretambrose: this one may have been overzealous

source/bigint.c
bretambrose: Dropped, if constant-time multiplication ever becomes an issue will think about then I guess.

source/bigint.c
bretambrose: standard attempt at preventing optmization, new explanatory comment with #define

PR #649, opened 2020-06-26
Add aws_byte_buf_append_dynamic_secure

bretambrose:
Adds aws_byte_buf_append_dynamic_secure() which is a dynamic append that also zeros any released memory before the free operation.


include/aws/common/byte_buf.h
```diff
@@ -368,6 +368,19 @@ int aws_byte_buf_append_with_lookup(
 AWS_COMMON_API
 int aws_byte_buf_append_dynamic(struct aws_byte_buf *to, const struct aws_byte_cursor *from);
 
+/**
+ * Copies from to to. If to is too small, the buffer will be grown appropriately and
```
singku-china, line 372: I suggest we rename all these args to `src` and `dest`

bretambrose: It's worth considering, but for now I'm just going to put quote ticks around the names for clarity.

PR #654, opened 2020-07-01
Trace event

ramosth-aws:
*Issue #, if available:*

Fixes for external dependencies and making the code towards enterprise. Still need to fix and issue that I made with all the error checking but I should be sending more PR's for check ups.
 
https://github.com/awslabs/aws-c-common/pull/635#issue-415607259


include/aws/common/cJSON.h
```diff
@@ -0,0 +1,293 @@
+/*
```
justinboswell, line 1: This should not be in the public header folder, it should be private.

include/aws/common/trace_event.h
```diff
@@ -0,0 +1,108 @@
...
+#include <aws/common/process.h>
+#include <aws/common/thread.h>
+AWS_EXTERN_C_BEGIN
+
+/* Must use INIT before calling any other macros*/
+#define AWS_TRACE_EVENT_INIT(address) aws_trace_event_init(address, aws_default_allocator())
+#define AWS_TRACE_EVENT_INIT_(address, allocator) aws_trace_event_init(address, allocator)
+
+/* For use in cpp programs */
+#define AWS_TRACE_EVENT_BEGIN_CPP(category, name) aws_trace_event_new(strdup(category), strdup(name), EVENT_PHASE_BEGIN)
```
justinboswell, line 20: Instead of duping this, create one on the stack, even if it's static, and pass it in. Use the do { ... } while(0) style macro to execute a block.

include/aws/common/trace_event.h
```diff
@@ -0,0 +1,108 @@
...
+/* For use in c programs */
+#define AWS_TRACE_EVENT_BEGIN(category, name) aws_trace_event_new(category, name, EVENT_PHASE_BEGIN)
+#define AWS_TRACE_EVENT_END(category, name) aws_trace_event_new(category, name, EVENT_PHASE_END)
+
+/* To clean up and to write trace event metadata to JSON file */
+#define AWS_TRACE_EVENT_CLEAN_UP_OUT(output_file) aws_trace_event_clean_up(1, output_file)
+#define AWS_TRACE_EVENT_CLEAN_UP_OUT_CPP(output_file) aws_trace_event_clean_up(1, strdup(output_file))
+
+/* To clean up without writing to JSON file */
+#define AWS_TRACE_EVENT_CLEAN_UP() aws_trace_event_clean_up(0, NULL)
```
justinboswell, line 32: Duplicate

include/aws/common/trace_event.h
```diff
@@ -0,0 +1,108 @@
...
+    char *category;
+    /* timestamp in milliseconds */
+    unsigned long long timestamp;
+    unsigned long int thread_id;
+    int process_id;
+    /* args for more metadata to be added later */
+};
+
+struct aws_trace_event {
+    struct cJSON *root, *event_array;
```
justinboswell, line 55: Hide this structure in the C file, it shouldn't be necessary to know its structure outside the library. Also, we do not want cJSON types to leak out

include/aws/common/trace_event.h
```diff
@@ -0,0 +1,108 @@
...
+};
+
+//! this function can go solely in the .c file
+/*
+ * Subscirbes to the aws_message_bus and writes trace event data
+ * to a JSON object when it is thread safe to do so.
+ */
+// AWS_COMMON_API
+// void aws_trace_event_listener(uint64_t address, const void *msg, void *user_data);
+
```
justinboswell, line 67: Should this just be removed?

include/aws/common/trace_event.h
```diff
@@ -0,0 +1,108 @@
...
+ */
+AWS_COMMON_API
+int aws_trace_event_new(char *category, char *name, char phase);
+
+/*
+ * Used for debugging
+ *
+ * AWS_COMMON_API
+ */
+struct cJSON *aws_trace_event_get_root(void);
```
justinboswell, line 105: Don't let this type leak out through this header. Maybe create a node type to wrap it.

justinboswell (changes requested):

include/aws/common/trace_event.h
```diff
@@ -0,0 +1,108 @@
...
+// void aws_trace_event_listener(uint64_t address, const void *msg, void *user_data);
+
+/*
+ * Starts the aws_message_bus in a background thread and subscribes the listener to it.
+ * Initializes a JSON object to store trace event data
+ * Must be called before using event_trace_add or close_event_trace
+ * Listener_id is the assigned id integer of the listener function
+ */
+AWS_COMMON_API
+int aws_trace_event_init(uint64_t address, struct aws_allocator *allocator);
```
graebm, line 75: naming suggestion:
aws_trace_init() / aws_trace_system_init() ?

the word "event" makes me think we're talking about one of the events.

possibly also: trace.h

also: don't need `address`

source/trace_event.c
```diff
@@ -0,0 +1,129 @@
...
+    cJSON_AddNumberToObject(event, "tid", trace_event_data->thread_id);
+    cJSON_AddNumberToObject(event, "ts", trace_event_data->timestamp);
+}
+
+/* Destructor of trace event metadata sent through the bus */
+void aws_trace_event_destroy(void *payload) {
+    struct aws_trace_event_metadata *trace_event_data = (struct aws_trace_event_metadata *)payload;
+    aws_mem_release(trace_event->allocator, trace_event_data->name);
+    aws_mem_release(trace_event->allocator, trace_event_data->category);
+}
```
graebm, line 44: We don't need to allocate anything hooray!

source/trace_event.c
```diff
@@ -0,0 +1,129 @@
...
+/*
+ * Private API
+ */
+
+/* Add trace event meta data to JSON object when the message bus allows it */
+void aws_trace_event_listener(uint64_t address, const void *msg, void *user_data) {
+    struct aws_trace_event_metadata *trace_event_data = (struct aws_trace_event_metadata *)msg;
+    cJSON *event = cJSON_CreateObject();
+
+    cJSON_AddItemToArray(trace_event->event_array, event);
```
graebm, line 23: need to check every cJSON_XYZ() call
but it's C and error-checking makes your code way harder to read & write
maybe just a macro that crashes the world if the allocation fails?

`#define CHECK(x) do { if (!x) AWS_FATAL_ASSERT(0 && "trace failure"); } while (0)`

then you can do:
`CHECK(cJSON_AddItemToArray(trace_event->event_array, event));`

source/trace_event.c
```diff
@@ -0,0 +1,129 @@
...
+
+    trace_event->root = cJSON_CreateObject();
+    trace_event->event_array = cJSON_CreateArray();
+    cJSON_AddItemToObject(trace_event->root, "traceEvents", trace_event->event_array);
+
+    aws_high_res_clock_get_ticks(&start_time);
+    listener_id = address;
+
+    if (allocator == NULL || trace_event->root == NULL || trace_event->event_array == NULL) {
+        return AWS_OP_ERR;
```
graebm, line 63: it's really hard to do great error handling. If you do 3 things, but the first 2 worked, you need to clean up those first 2 before returning error.

maybe just explode if something goes wrong here with the the CHECK() macro

source/trace_event.c
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /* start the message bus */
+    /* !!! POSSIBLY ADD OPTION TO CHOOSE SYNC VS ASYNC MAYBE? !!! */
+    struct aws_bus_options options = {
+        .allocator = allocator,
+        .policy = AWS_BUS_SYNC,
+    };
+    aws_bus_init(&(trace_event->bus), &options);
+    aws_bus_subscribe(&(trace_event->bus), listener_id, aws_trace_event_listener, NULL);
```
graebm, line 73: check these 2 functions for success/failure

source/trace_event.c
```diff
@@ -0,0 +1,129 @@
...
+
+    return AWS_OP_SUCCESS;
+}
+
+int aws_trace_event_clean_up(int code, char *filename) {
```
graebm, line 78: clean up functions are not allowed to fail

source/trace_event.c
```diff
@@ -0,0 +1,129 @@
...
+        .thread_id = thread_id,
+        .process_id = process_id,
+    };
+    trace_event_data.name = aws_mem_acquire(trace_event->allocator, ((sizeof(char) * strlen(name)) + 1));
+    trace_event_data.category = aws_mem_acquire(trace_event->allocator, ((sizeof(char) * strlen(category)) + 1));
+    strcpy(trace_event_data.name, name);
+    strcpy(trace_event_data.category, category);
+
+    /* send to the bus */
+    aws_bus_send(&(trace_event->bus), listener_id, &trace_event_data, aws_trace_event_destroy);
```
graebm, line 123: can this fail?

include/aws/common/trace_event.h
```diff
@@ -0,0 +1,84 @@
...
+ * MUST CALL TO AVOID MEMORY LEAKS
+ * Cleans up the aws_message_bus and unsubscribes the listener.
+ * Prints the JSON object that stores trace event data to the given filename
+ * as a .JSON file.
+ * Closes the JSON object and frees all allocated memory.
+ * Code 0 to not write out file and Code 1 to write out to file
+ * time_unit how seconds are displayed on chrome://tracing
+ */
+AWS_COMMON_API
+void aws_trace_system_clean_up(int code, int time_unit, const char *filename);
```
graebm, line 66: tldr; `aws_trace_system_init(struct aws_allocator *allocator, const char *filename)`

These arguments should go into the init() function, not the clean_up(). Currently, you're writing everything out during clean_up(), but eventually we want to be writing to the file incrementally as events roll in (so we don't have an infinitely growing JSON object in memory) so we'll need to know the filename up front.

also, maybe let's get rid of the `code` argument. "int code" isn't very descriptive compared to something like`bool write_to_file`. But also, we could just say that passing NULL to `filename` is how the user says they don't want to a file, so we use one argument that does what it says instead of 2 different arguments that influence the same thing.

also, let's drop the `time_unit` option for now, just pick one. I imagine there will be LOTS of configuration options for how we want this thing to run when all is said and done. It's a pain to add arguments to a function over time, it breaks the API and you need to track down everywhere that function is called and update how it's called...
Let's just keep the init() call simple for now, and hard-code misc options like time-unit. Later, when we're more certain what to expose, we can introduce some alternate init() function that takes a million arguments, or takes a struct where the million arguments are members of the struct.

include/aws/common/trace_event.h
```diff
@@ -0,0 +1,84 @@
...
+#define AWS_COMMON_TRACE_EVENT_H
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+/* Disable warnings with windows build */
+#ifdef _MSC_VER
+#    pragma warning(disable : C4996)
+#endif
```
graebm, line 11: Don't disable warnings from header files.
That causes the warnings to be disabled in ANY code that includes this header, including our customer's code.
Our customer might not want such a warning disabled.

move this to the .c file

A big difference between student/hobby projects and giant corporate codebases is:
In small code bases you put stuff in headers to simplify writing your .c files.
In big corporate codebases you only put what's absolutely necessary in your headers. You don't want to add anything to your user's .c files that they might not want (extra includes they don't necessarily need that increase compile time, disabled warnings they might not want disabled, stuff like that)

source/trace_event.c
```diff
@@ -0,0 +1,191 @@
...
+void aws_trace_system_write(int time_unit, const char *filename) {
+    if (time_unit) {
+        if (cJSON_AddStringToObject(trace_event->root, "displayTimeUnit", "ns") == NULL) {
+            aws_raise_error(AWS_ERROR_OOM);
+        }
+    }
+    char *out = cJSON_Print((trace_event->root));
+    if (out == NULL) {
+        aws_raise_error(AWS_ERROR_OOM);
+    }
```
graebm, line 86: calling `aws_raise_error()` doesn't really do anything. It sets a thread-local variable, but it doesn't automatically `return` or anything, this function is going to keep running. If this is in error, you'll want to either FATAL_ASSERT, or `return` or `goto` somewhere.

Seems like most of the error-handling in this file needs to be fixed up accordingly

source/trace_event.c
```diff
@@ -0,0 +1,191 @@
...
+    FILE *fp;
+    char fn[strlen(filename) + 6];
+
+    strcpy(fn, filename);
+    strncat(fn, ".json", 6);
+
+    fp = fopen(fn, "w");
+    if (fp == NULL) {
+        aws_raise_error(AWS_ERROR_OOM);
```
graebm, line 95: Not every problem is due to OOM. There are lots of reasons `fopen` could fail. If it fails it sets thread-local `errno` [docs](https://en.cppreference.com/w/c/io/fopen)

We don't have a 1:1 mapping of every possible `errno` code to aws-error codes, but we do have a function that makes an attempt. You could use it here like:
`aws_translate_and_raise_io_error(errno)`

source/trace_event.c
```diff
@@ -0,0 +1,191 @@
...
+    if (allocator == NULL) {
+        return AWS_OP_ERR;
+    }
+    trace_event = aws_mem_acquire(allocator, sizeof(struct aws_trace_event));
+    trace_event->allocator = allocator;
+
+    trace_event->root = cJSON_CreateObject();
+    if (trace_event->root == NULL) {
+        /* Error code may be changed on review */
+        aws_raise_error(AWS_ERROR_OOM);
```
graebm, line 117: The pain of C is that there are no destructors, nothing cleans up automatically.
If a function does 3 things: A, B, C
and it fails doing C
then you need to clean up B, A before returning

I could rant on the different styles of error handling, but this is what I suggest:

Use aws_mem_calloc() when creating structs so that everything starts zeroed out.
That way, you can tell later whether or not something needs cleanup based on whether it's zeroed out.
```
trace_event = aws_mem_calloc(allocator, 1, sizeof(struct aws_trace_event);
```

If anything goes wrong, `goto` a common block of code that handles all cleanup, regardless of what went wrong (also, note that you DON'T need to raise OOM here, because aws_mem_xzy() would have raised it already before returning NULL):
```
if (trace_event->root == NULL) {
    goto error;
}
```

The error block should blast everything. Since this code is likely to look a LOT like the code in your `clean_up()` function, you might as well just call your clean_up() function
```
error:
    aws_trace_system_clean_up();
    return AWS_OP_ERR;
}
```

and write your clean_up() function to handle a half-initialized struct:
```
void aws_trace_system_clean_up() {
    /* check whether system was already cleaned up (or failed creation) */
    if (!trace_system) {
        return;
    }

    /* Only write file if it was successfully opened */
    if (trace_system->file) {
        write_file()
        fclose(...)
    }
    ...
```

source/trace_event.c
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/*
+ * Public API
+ */
+
+/* starts the message bus and initializes the JSON root, and the event array for storing metadata */
+int aws_trace_system_init(struct aws_allocator *allocator) {
+    if (allocator == NULL) {
+        return AWS_OP_ERR;
```
graebm, line 109: Need to be sure the reason for the failure is set before -1 is returned
`return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);`

karkhaz:
Hi @ramosth-aws, please rebase this PR onto master to ensure that the tests pass. Cheers!

source/trace_event.c
```diff
@@ -0,0 +1,353 @@
...
+    char *out = cJSON_Print((trace_event->root));
+    if (out == NULL) {
+        aws_raise_error(AWS_ERROR_OOM);
+        AWS_FATAL_ASSERT(0 && "trace failure");
+    }
+    FILE *fp;
+    char fn[strlen(filename_out) + 6];
+
+    strcpy(fn, filename_out);
+    strncat(fn, ".json", 6);
```
rccarper, line 107: If you break out ".json" into a constant, you can grab the length from that and not have to hardcode the 6.

source/trace_event.c
```diff
@@ -0,0 +1,353 @@
...
+
+    struct aws_trace_event_metadata trace_event_data = {
+        .phase = phase,
+        .timestamp = timestamp,
+        .thread_id = thread_id,
+        .process_id = process_id,
+    };
+
+    strncpy(trace_event_data.name, name, 14);
+    strncpy(trace_event_data.category, category, 14);
```
rccarper, line 313: I think this will get cleaned up with some of the previous feedback, but try to avoid hardcoded constants for buffer sizes like this.  If someone changes the length of trace_event_data.category to something smaller, the 14 here becomes invalid, and it's not obvious to change.

tests/bus_test.c
```diff
@@ -0,0 +1,91 @@
...
+
+static struct {
+    int count;
+    bool payload_deleted;
+} s_sync_test;
+
+static const char s_test_payload[] = "TEST ME SENPAI";
+
+static void s_bus_sync_test_recv(uint64_t address, const void *msg, void *user_data) {
+    AWS_ASSERT(42 == address);
```
rccarper, line 28: 42 could be global constant.  It'll make it easier to change and arguably makes the mistake of using the wrong value less likely, since a typo will give you a compile error. :)

tests/bus_test.c
```diff
@@ -0,0 +1,91 @@
...
+    };
+
+    struct aws_bus bus;
+    ASSERT_SUCCESS(aws_bus_init(&bus, &options));
+
+    AWS_ZERO_STRUCT(s_sync_test);
+
+    ASSERT_SUCCESS(aws_bus_subscribe(&bus, 42, s_bus_sync_test_recv, &s_sync_test));
+
+    ASSERT_SUCCESS(aws_bus_send(&bus, 42, (void *)&s_test_payload[0], s_test_payload_dtor));
```
rccarper, line 54: Super trivial thing: &s_test_payload[0] == s_test_payload

graebm (approved):
Looks good!

danielsn:
Can you try rebasing against master and see if that fixes the CBMC proofs?
Also, if you wouldn't mind changing this to draft, that would prevent it spamming our tickets (we get a ticket for every proof failure on a non-draft PR)

PR #657, opened 2020-07-02
aws_byte_cursor_read_hex_u8() and other string helpers

graebm:
**new**  `aws_byte_cursor_read_hex_u8()`: Reads 2 hex characters from ASCII/UTF-8 text to produce an 8-bit number.

**new** `aws_lookup_table_hex_to_num_get()`:  Returns lookup table to go from ASCII/UTF-8 hex character to a number (0-15).

**new** `aws_string_new_from_cursor()` and `aws_string_new_from_buf()`: obviously

**fixed** `aws_byte_cursor_next_split()`: fixed to handle source with 0 len and NULL ptr


include/aws/common/byte_buf.h
```diff
@@ -651,7 +667,18 @@ AWS_COMMON_API bool aws_byte_cursor_read_float_be64(struct aws_byte_cursor *cur,
  * If there is insufficient space in the cursor, returns false, leaving the
  * cursor unchanged.
  */
-AWS_COMMON_API bool aws_byte_cursor_read_be64(struct aws_byte_cursor *cur, uint64_t *var);
+AWS_COMMON_API bool aws_byte_cursor_read_float_be64(struct aws_byte_cursor *cur, double *var);
```
singku, line 670: Maybe it's better to rename it as `read_double_be64`

source/byte_buf.c
```diff
@@ -162,48 +163,44 @@ bool aws_byte_cursor_next_split(
...
+
+    /* If substr is zeroed-out, then this is the first run. */
+    const bool first_run = substr->ptr == NULL;
+    if (first_run) {
+        if (input_str->ptr) {
+            *substr = *input_str;
+        } else {
+            /* NOTE: It's also legal for input_str to be zeroed out.
+             * In that case, set substr->ptr to something else so the next split() call doesn't look like first run. */
+            substr->ptr = (void *)"";
```
singku, line 176: Extreme case: what if the input_str also points to `(void*)""`

source/byte_buf.c
```diff
@@ -162,48 +163,44 @@ bool aws_byte_cursor_next_split(
...
 
-        if (substr->len == 0) {
-            /* If split character was last in the string, return empty substr. */
-            return true;
+    } else {
+        /* This is not the first run.
+         * Advance substr past the previous split. */
+        const uint8_t *input_end = input_str->ptr + input_str->len;
+        substr->ptr += substr->len + 1;
+        if (substr->ptr > input_end || substr->ptr < input_str->ptr) { /* 2nd check is overflow check */
```
singku, line 185: `substr->ptr >= input_end` ? Note that input_end already points to 1 after the last char of input_str

include/aws/common/byte_buf.h
graebm: I didn't rename this. I just moved it so all the `read_float` functions were together

source/byte_buf.c
graebm: That would be fine. In that case we'd take the `if (input_str->ptr)` branch above.
we just need `substr->ptr` to point to something non-null

source/byte_buf.c
graebm: Yeah, this is on purpose. If the `split_on` char is the last char in the string, then we want the last split to be "", where the ptr is actually pointing at `input_end`.

For example, given "12&34&", split on '&', we should get the following 3 splits: "12", "34", ""
That final "" is pointing at `input_end` (but with len 0 so the user never actually accesses the memory)

graebm: Here is my attempt to explain, in carefully aligned ascii art:
```
address:  1234567
input:   "AB&CD&"  ptr:1 len:6 input_end:7
split0:  "AB"      ptr:1 len:2
split1:     "CD"   ptr:4 len:2
split2:        ""  ptr:7 len:0
```

include/aws/common/byte_buf.h
```diff
@@ -624,24 +640,24 @@ AWS_COMMON_API bool aws_byte_cursor_read_be24(struct aws_byte_cursor *cur, uint3
...
- * Reads a 64-bit value in network byte order from cur, and places it in host
+ * Reads a 32-bit value in network byte order from cur, and places it in host
  * byte order into var.
  *
  * On success, returns true and updates the cursor pointer/length accordingly.
  * If there is insufficient space in the cursor, returns false, leaving the
  * cursor unchanged.
  */
-AWS_COMMON_API bool aws_byte_cursor_read_float_be64(struct aws_byte_cursor *cur, double *var);
+AWS_COMMON_API bool aws_byte_cursor_read_float_be32(struct aws_byte_cursor *cur, float *var);
```
graebm, line 660: These didn't really change names. I just moved them so all the whole-number `read` functions were together and all the `read_float` functions were together

include/aws/common/byte_buf.h
singku: I understand, what I meant was "It might be better to just rename it." because the argument itself has double type.

PR #664, opened 2020-07-07
aws_linked_list_move_all_back()

graebm:
O(1) helper functions to move all nodes from one list to another.

This is something we do pretty commonly while holding a lock. It's much less code using these helper functions.


include/aws/common/linked_list.inl
```diff
@@ -343,6 +343,7 @@ AWS_STATIC_IMPL struct aws_linked_list_node *aws_linked_list_pop_front(struct aw
 AWS_STATIC_IMPL void aws_linked_list_swap_contents(struct aws_linked_list *a, struct aws_linked_list *b) {
     AWS_PRECONDITION(aws_linked_list_is_valid(a));
     AWS_PRECONDITION(aws_linked_list_is_valid(b));
+    AWS_PRECONDITION(a != b);
```
JonathanHenson, line 346: put AWS_RESTRICT on them and this is guaranteed right?

graebm: I'll throw AWS_RESTRICT on these functions, but I'm keeping the asserts

After adding restrict, I modified a test to swap the same object.
clang 9 didn't complain :|
but that would be really nice if it did

PR #670, opened 2020-07-10
Add arm64 inline assembly for saturating ops

AGSaidi:
*Issue #, if available:*

*Description of changes:*


AGSaidi:
tested by changing `math.inl` to always use the assembling and running saturating tests.

include/aws/common/math.gcc_arm64_asm.inl
```diff
@@ -0,0 +1,184 @@
...
+ * If a * b overflows, returns AWS_OP_ERR; otherwise multiplies
+ * a * b, returns the result in *r, and returns AWS_OP_SUCCESS.
+ */
+AWS_STATIC_IMPL int aws_mul_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
+    /* We can use inline assembly to do this efficiently on arm64 by doing
+       a high-mul and checking  the upper 64 bits of a 64x64->128b multiply
+       are zero */
+
+    uint64_t tmp, res;
+    __asm__("umulh %x[hmul], %x[arga], %x[argb]\n"
```
JonathanHenson, line 49: do these __asm__ blocks need to be volatile?

AGSaidi: i don't see why they aren't modifying any state beyond what is marked as an output and if the compiler determines that the value isn't consumed it can be elided

PR #674, opened 2020-07-29
Moving XML parser from aws-c-auth into aws-c-common

rccarper:
*Description of changes:*
Adding XML parser from aws-c-auth.  Made minor changes for naming, also made a couple of types opaque and added accessors.

Justification for leaving aws_xml_attribute visible: it seems reasonable that we could expose an attribute as a name/value pair, and that this would be less awkward than needing a get-attribute, get-name, get-value each time when needing to use a single attribute.  (But I'm open to making this opaque too!)


graebm:
I know it's a lot more work, but should these go into a new repo, instead of aws-c-common?

PR #689, opened 2020-09-04
Updates post-conditions in aws_mul_size_checked proof

feliperodri:
Signed-off-by: Felipe R. Monteiro <felisous@amazon.com>

*Issue #, if available:*

N/A.

*Description of changes:*

Updates a post-condition in `aws_mul_size_checked` proof harness.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #693, opened 2020-09-14
Updates the postconditions in aws_mul_size_checked proof

feliperodri:

*Issue #, if available:*

N/A.

*Description of changes:*

Reverts https://github.com/awslabs/aws-c-common/pull/689. We need actually map the same check used in https://github.com/awslabs/aws-c-common/blob/master/include/aws/common/math.fallback.inl.

### DIFF
--- a/verification/cbmc/proofs/aws_mul_size_checked/aws_mul_size_checked_harness.c
+++ b/verification/cbmc/proofs/aws_mul_size_checked/aws_mul_size_checked_harness.c
@@ -24,7 +24,7 @@ void aws_mul_size_checked_harness() {
         if (!aws_mul_u64_checked(a, b, &r)) {
             assert(r == a * b);
         } else {
-            assert(__CPROVER_overflow_mult(a, b));
+            assert(a > 0 && b > 0 && a > (UINT64_MAX / b));
         }
     } else {
         /*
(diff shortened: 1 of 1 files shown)


SaswatPadhi (approved):
I see why the check in the harness should be consistent with the one in `math.fallback.inl`, but why don't we use `__CPROVER_overflow_mult` at both places? I'm guessing because it doesn't make the integer width (64bit vs 32bit) explicit?

verification/cbmc/proofs/aws_mul_size_checked/aws_mul_size_checked_harness.c
```diff
@@ -24,7 +24,7 @@ void aws_mul_size_checked_harness() {
         if (!aws_mul_u64_checked(a, b, &r)) {
             assert(r == a * b);
         } else {
-            assert(__CPROVER_overflow_mult(a, b));
```
danielsn, line 27: I think this is backwards. You want to use this as the check, but you want to make sure that you're linking against the actual C check in the math.h

feliperodri: Yep, I can disable the CBMC flag in this proof.

PR #694, opened 2020-09-15
Improves predicates for aws_array_list proofs

feliperodri:
*Issue #, if available:*

N/A.

*Description of changes:*

 - Adds `AWS_IMPLIES` and `AWS_IFF` macros;
 - Improves the `aws_array_list` validity function;
 - Improves predicates for CBMC proofs;
 - Fixes `format-check.sh` replacing `.cbmc-batch` with new `verification` dir.

### DIFF
--- a/format-check.sh
+++ b/format-check.sh
@@ -10,7 +10,7 @@ if NOT type $CLANG_FORMAT 2> /dev/null ; then
 fi
 
 FAIL=0
-SOURCE_FILES=`find source include tests .cbmc-batch -type f \( -name '*.h' -o -name '*.c' -o -name '*.inl' \)`
+SOURCE_FILES=`find source include tests verification -type f \( -name '*.h' -o -name '*.c' -o -name '*.inl' \)`
 for i in $SOURCE_FILES
 do
     $CLANG_FORMAT -output-replacements-xml $i | grep -c "<replacement " > /dev/null
--- a/include/aws/common/assert.h
+++ b/include/aws/common/assert.h
@@ -107,6 +107,15 @@ AWS_EXTERN_C_END
 #define __CPROVER_r_ok(base, len) (AWS_MEM_IS_READABLE(base, len))
 #define __CPROVER_w_ok(base, len) (AWS_MEM_IS_WRITEABLE(base, len))
 
+/* Logical consequence. */
+#define AWS_IMPLIES(a, b) (!(a) || (b))
+
+/**
+ * If and only if (iff) is a biconditional logical connective between statements a and b.
+ * Equivalent to (AWS_IMPLIES(a, b) && AWS_IMPLIES(b, a)).
+ */
+#define AWS_IFF(a, b) (!!(a) == !!(b))
+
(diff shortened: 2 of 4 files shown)


include/aws/common/assert.h
```diff
@@ -107,6 +107,15 @@ AWS_EXTERN_C_END
...
 #define __CPROVER_w_ok(base, len) (AWS_MEM_IS_WRITEABLE(base, len))
 
+/* Logical consequence. */
+#define AWS_IMPLIES(a, b) (!(a) || (b))
+
+/**
+ * If and only if (iff) is a biconditional logical connective between statements a and b.
+ * Equivalent to (AWS_IMPLIES(a, b) && AWS_IMPLIES(b, a)).
+ */
+#define AWS_IFF(a, b) (!!(a) == !!(b))
```
SaswatPadhi, line 117: At first glance, it's not clear why we need double negations `!!` here since they _should_ simply cancel each other out (in logic). But you need it here for `AWS_IFF` to work correctly for non-Boolean `a` and `b` values. I think this should be documented, so someone doesn't simply optimizie it to `a == b` in future.

Also, we don't seem to be using `AWS_IFF` within this PR.

SaswatPadhi (approved):
LGTM, except one minor comment. :+1:

format-check.sh
```diff
@@ -10,7 +10,7 @@ if NOT type $CLANG_FORMAT 2> /dev/null ; then
 fi
 
 FAIL=0
-SOURCE_FILES=`find source include tests .cbmc-batch -type f \( -name '*.h' -o -name '*.c' -o -name '*.inl' \)`
+SOURCE_FILES=`find source include tests verification -type f \( -name '*.h' -o -name '*.c' -o -name '*.inl' \)`
```
danielsn, line 13: good catch

include/aws/common/assert.h
danielsn: +1

verification/cbmc/sources/make_common_data_structures.c
```diff
@@ -91,13 +91,8 @@ bool aws_array_list_is_bounded(
...
 
 void ensure_array_list_has_allocated_data_member(struct aws_array_list *const list) {
-    if (list->current_size == 0 && list->length == 0) {
-        __CPROVER_assume(list->data == NULL);
-        list->alloc = can_fail_allocator();
-    } else {
-        list->data = bounded_malloc(list->current_size);
-        list->alloc = nondet_bool() ? NULL : can_fail_allocator();
-    }
+    list->data = can_fail_malloc(list->current_size);
```
danielsn, line 94: what happens if `list->current_size  == 0`

danielsn: Offline discussion: this is handled by the validity function.

danielsn (approved):
Approved, modulo adding the comment Saswat requested

include/aws/common/assert.h
feliperodri: I have implemented as suggested.

PR #697, opened 2020-09-17
Improved header checker

graebm:
- Checks against multiple CXX and C versions, not just C90 and C++11.
- Ability to just check against just CXX or just C (needed so we can use this with aws-crt-cpp)

Downside with this appraoch is that it makes 7 extra projects to build, instead of just 1. But it's something you need to opt into, so that seems fine to me.


graebm:
nevermind. aws-crt-cpp should have its own implementation of this.
aws-c-*** are meant to work with EVERY imaginable version of C and C++.
aws-crt-cpp is c++11+ only

PR #700, opened 2020-09-23
Don't hardcode the CoreFoundation path

### LINKED ISSUE #510: [OSX] Hardcoded CoreFoundation path in aws-c-common-targets.cmake
I don't have Xcode installed locally, only command line tools, so the linker cannot locate CoreFoundation:

```cmake
set_target_properties(AWS::aws-c-common PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "AWS_COMMON_USE_IMPORT_EXPORT"
  INTERFACE_COMPILE_OPTIONS "-pthread"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "-pthread;pthread;/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/System/Library/Frameworks/CoreFoundation.framework"
)
```

A temporary solution is [overriding it](https://github.com/apache/arrow/pull/5423/files#diff-2420b0c5b6bdad921f1d538f92d64b59R2546).

### What platform/OS are you using?

OSX Mojave

### Which version of the SDK?

Installed by home-brew:

```
aws-sdk-cpp: stable 1.7.180 (bottled), HEAD
AWS SDK for C++
https://github.com/aws/aws-sdk-cpp
/usr/local/Cellar/aws-sdk-cpp/1.7.180 (22,152 files, 426.2MB)
  Poured from bottle on 2019-09-18 at 15:36:12
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/aws-sdk-cpp.rb
```

Original issue: https://github.com/aws/aws-sdk-cpp/issues/1245

xhochy:
This would otherwise hardcode the Xcode version and we would have to recompile after each Xcode update.

*Issue #, if available:*

Fixes #510

*Description of changes:*

Instead of using the explicit path of the CoreFoundation library, we can instruct the linker to use it. I kept the `find_library` check in place as this gives a much nicer error to the user if not all parts of the Xcode tools are installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,7 +90,9 @@ else ()
             message(FATAL_ERROR "Core Foundation not found")
         endif ()
 
-        list(APPEND PLATFORM_LIBS Threads::Threads ${CORE_FOUNDATION_LIB})
+        list(APPEND PLATFORM_LIBS Threads::Threads)
+        # Don't add the exact path to CoreFoundation as this would hardcode the SDK version
+        list(APPEND PLATFORM_LINK_FLAGS -framework CoreFoundation)
     elseif (${CMAKE_SYSTEM_NAME} STREQUAL "Linux") # Android does not link to libpthread nor librt, so this is fine
         list(APPEND PLATFORM_LIBS m Threads::Threads rt)
     elseif(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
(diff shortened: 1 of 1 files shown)


xhochy:
@justinboswell Why was this closed without a comment? I'm happy to include fixes but this PR would be really helpful for people packaging this on OSX:

PR #705, opened 2020-10-03
shorter submodule paths

graebm:
Shorten actual path, as well as "logical path" in .gitmodules file. This is to aid with pain on windows when paths exceed 260 characters. We do a lot of recursive submodules, and these things add up.


feliperodri (approved):
LGTM.

PR #714, opened 2020-10-14
Updates aws_hash_table stubs for CBMC proofs

tegansb:
*Description of changes:*
Updates the hash_iterator_string_string_generator and hash_find_string_string_generator functions with non-NULL assumptions. 

### DIFF
--- a/verification/cbmc/stubs/hash_table_generators.c
+++ b/verification/cbmc/stubs/hash_table_generators.c
@@ -20,7 +20,9 @@ void hash_iterator_string_string_generator(struct aws_hash_iter *new_iter, const
     (void)old_iter;
     if (new_iter->status == AWS_HASH_ITER_STATUS_READY_FOR_USE) {
         new_iter->element.key = ensure_string_is_allocated_nondet_length();
+        __CPROVER_assume(new_iter->element.key != NULL);
         new_iter->element.value = ensure_string_is_allocated_nondet_length();
+        __CPROVER_assume(new_iter->element.value != NULL);
     }
 }
 
@@ -32,8 +34,12 @@ void hash_find_string_string_generator(
     const struct aws_hash_table *map,
     const void *key,
     struct aws_hash_element *p_elem) {
-    p_elem->key = ensure_string_is_allocated_nondet_length();
-    p_elem->value = ensure_string_is_allocated_nondet_length();
+    if (p_elem) {
+        p_elem->key = ensure_string_is_allocated_nondet_length();
+        __CPROVER_assume(p_elem->key != NULL);
+        p_elem->value = ensure_string_is_allocated_nondet_length();
+        __CPROVER_assume(p_elem->value != NULL);
+    }
 }
 
 /**


verification/cbmc/stubs/hash_table_generators.c
```diff
@@ -20,7 +20,9 @@ void hash_iterator_string_string_generator(struct aws_hash_iter *new_iter, const
     (void)old_iter;
     if (new_iter->status == AWS_HASH_ITER_STATUS_READY_FOR_USE) {
         new_iter->element.key = ensure_string_is_allocated_nondet_length();
+        __CPROVER_assume(new_iter->element.key != NULL);
         new_iter->element.value = ensure_string_is_allocated_nondet_length();
+        __CPROVER_assume(new_iter->element.value != NULL);
```
feliperodri, line 25: Do we guarantee these assumptions in this function as well? What happens with the actual function?

verification/cbmc/stubs/hash_table_generators.c
```diff
@@ -32,8 +34,12 @@ void hash_find_string_string_generator(
...
     const void *key,
     struct aws_hash_element *p_elem) {
-    p_elem->key = ensure_string_is_allocated_nondet_length();
-    p_elem->value = ensure_string_is_allocated_nondet_length();
+    if (p_elem) {
+        p_elem->key = ensure_string_is_allocated_nondet_length();
+        __CPROVER_assume(p_elem->key != NULL);
+        p_elem->value = ensure_string_is_allocated_nondet_length();
+        __CPROVER_assume(p_elem->value != NULL);
+    }
```
feliperodri, line 42: Same question here.

verification/cbmc/stubs/hash_table_generators.c
tegansb: So there is no matching function corresponding to this stub. It's a generator to generate items in a hash table when it's iterated over (it's the HASH_ITER_ELEMENT_GENERATOR from the aws_hash_iter_override stubs). So whether it's guaranteed is I guess an assumption over the hash_table we're iterating over. But in general, some functions might require hash tables with this guarantee and some might not, so maybe we could have two functions? I'll also think of a way to add this as a precondition to the function itself.

verification/cbmc/stubs/hash_table_generators.c
tegansb: Same as above (except for find element instead of iterate).

verification/cbmc/stubs/hash_table_generators.c
feliperodri: Could you create a GitHub issue to track whether we should map these as a precondition? @tegansb

feliperodri (approved):
LGTM.

verification/cbmc/stubs/hash_table_generators.c
tegansb: Made here https://github.com/awslabs/aws-c-common/issues/715. Could you add the cbmc label?

feliperodri:
This changes has been merged in PR https://github.com/awslabs/aws-c-common/pull/716.

PR #739, opened 2020-12-03
Do not use addr2line by default, backtrace_symbols will do the job

justinboswell:


tests/memtrace_test.c
```diff
@@ -132,11 +142,20 @@ static int s_test_memtrace_stacks(struct aws_allocator *allocator, void *ctx) {
     /* if this is not a debug build, there may not be symbols, so the test cannot
      * verify if a best effort was made */
 #if defined(DEBUG_BUILD)
```
rccarper, line 144: Do we still need the define? (It looks like the new approach is meant to handle symbols for a debug build and having no symbols for a non-debug build?)

rccarper (approved):

PR #742, opened 2020-12-16
Use aligned allocs for > PAGE_SIZE allocations, avoid unaligned memcpy/memset on large buffers

justinboswell:

### DIFF
--- a/source/allocator.c
+++ b/source/allocator.c
@@ -65,7 +65,7 @@ static void *s_default_realloc(struct aws_allocator *allocator, void *ptr, size_
 #if !defined(_WIN32)
     return realloc(ptr, newsize);
 #else
-    return _aligned_realloc(ptr, newsize, sizeof(void *) * (size > PAGE_SIZE ? 8 : 1));
+    return _aligned_realloc(ptr, newsize, sizeof(void *) * (newsize > PAGE_SIZE ? 8 : 1));
 #endif
 }
 
--- a/source/allocator.c
+++ b/source/allocator.c
@@ -36,15 +36,12 @@ bool aws_allocator_is_valid(const struct aws_allocator *alloc) {
 
 static void *s_default_malloc(struct aws_allocator *allocator, size_t size) {
     (void)allocator;
+    /* larger allocations should be aligned so that AVX and friends can run
+     * the aligned versions of memcpy and memset and friends on big buffers */
+    size_t alignment = sizeof(void*) * (size > PAGE_SIZE ? 8 : 1)
 #if !defined(_WIN32)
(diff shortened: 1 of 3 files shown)


source/allocator.c
```diff
@@ -25,30 +25,65 @@
 #    pragma warning(disable : 4100)
 #endif
 
+#ifndef PAGE_SIZE
+#    define PAGE_SIZE (4 * 1024)
```
bretambrose, line 29: Is there any way we can genuinely determine this per-platform as it's probably a kernel setting?

source/allocator.c
```diff
@@ -25,30 +25,65 @@
...
     /* An allocator must define mem_acquire and mem_release.  All other fields are optional */
     return alloc && AWS_OBJECT_PTR_IS_READABLE(alloc) && alloc->mem_acquire && alloc->mem_release;
 }
 
 static void *s_default_malloc(struct aws_allocator *allocator, size_t size) {
     (void)allocator;
-    return malloc(size);
+    /* larger allocations should be aligned so that AVX and friends can run
+     * the aligned versions of memcpy and memset and friends on big buffers */
+    const size_t alignment = sizeof(void *) * (size > PAGE_SIZE ? 8 : 1);
```
bretambrose, line 41: Can we add a comment for us mere mortals explicitly giving out the contract:

64 byte alignment for > page allocations on 64 bit systems
32 byte alignment for > page allocations on 32 bit systems
8 byte alignment for <= page allocations on 64 bit systems
4 byte alignment for <= page allocations on 32 bit systems

At least that's that the code does.  I'd feel more confident in the correctness if the comment echoed that contract.

source/allocator.c
justinboswell: On POSIX, it's defined as PAGE_SIZE. On windows, it's 4K unless you twiddle it yourself.

source/allocator.c
justinboswell: Good call.

source/allocator.c
bretambrose: Ok, that's helpful

justinboswell: Additionally, you could use madvise() on POSIX to switch to HUGEPAGE, which makes 2MB pages. This is usually best left to the system to worry about. We could also encode this as a separate value that we default to PAGE_SIZE, but can be overridden. Thoughts?

bretambrose (approved):

source/allocator.c
bretambrose: Brain buffer overrun to be honest.

PR #744, opened 2020-12-17
Task scheduler with a background thread.

JonathanHenson:

### DIFF
--- a/include/aws/common/task_scheduler.h
+++ b/include/aws/common/task_scheduler.h
@@ -33,7 +33,12 @@ struct aws_task {
     struct aws_linked_list_node node;
     struct aws_priority_queue_node priority_queue_node;
     const char *type_tag;
-    size_t reserved;
+
+    /* honor the ABI compat */
+    union {
+        bool scheduled;
+        size_t reserved;
+    } abi_extension;
 };
 
 struct aws_task_scheduler {
--- a/include/aws/common/priority_queue.h
+++ b/include/aws/common/priority_queue.h
@@ -41,7 +41,7 @@ struct aws_priority_queue {
 };
 
 struct aws_priority_queue_node {
-    /** The current index of the node in queuesion, or SIZE_MAX if the node has been removed. */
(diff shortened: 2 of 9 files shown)


include/aws/common/task_scheduler.h
```diff
@@ -33,7 +33,12 @@ struct aws_task {
     struct aws_linked_list_node node;
     struct aws_priority_queue_node priority_queue_node;
     const char *type_tag;
-    size_t reserved;
+
+    /* honor the ABI compat */
+    union {
+        bool scheduled;
```
justinboswell, line 39: Do you want to bitfield this?

include/aws/common/task_scheduler.h
```diff
@@ -33,7 +33,12 @@ struct aws_task {
...
+        size_t reserved;
+    } abi_extension;
```
justinboswell, line 41: This union could be anonymous and still achieve ABI compat.

justinboswell (commented):
So this will give us 1 background thread, which is useful, but I suspect in the use case we're looking at it's going to back up. Are we going to end up with a thread pool at some point, holding a collection of these?

justinboswell:
Also, I am totally going to replace aws_message_bus's async guts with this.

include/aws/common/task_scheduler.h
JonathanHenson: We can, but we don’t have to til we start using the other 3 bytes.

include/aws/common/task_scheduler.h
JonathanHenson: Yeah but it won’t compile because that’s a c12 feature apparently.

JonathanHenson:
> So this will give us 1 background thread, which is useful, but I suspect in the use case we're looking at it's going to back up. Are we going to end up with a thread pool at some point, holding a collection of these?

That’s the idea.... but that part isn’t hard, and I’m assuming you’re already doing that. Regardless it’s a separate implementation from this.

justinboswell (approved):

PR #749, opened 2020-12-28
Fix compiler errors with MinGW compilers

Biswa96:
* Description of changes:

  - Windows: Check _MSC_VER for MSVC specific pragma warnings directives. `#pragma warning` is Microsoft Visual Compiler specific and creates warning with GCC as those are unknown.
  - include/common: Check _WIN32 to prevent htonl and ntohl functions. This replaces checking `_MSC_VER`  as we need to check if the platform is Microsoft Windows or not. So, `_WIN32` is proper preprocessor to check it.
  - tests: Check if alloca is already defined. GCC has `alloca` already defined to `__builtin_alloca`.
 

Biswa96:
Also if I enable `BUILD_SHARED_LIBS` the following error is shown at CMake configure time:

```cmake
CMake Error: install(EXPORT "aws-c-common-targets" ...) includes target "aws-c-common" more than once in the export set.
```

Biswa96:
Ping! Any comment on this pr?

Biswa96:
Ping.

bretambrose:
Merged in https://github.com/awslabs/aws-c-common/pull/772

Issue #754, opened 2021-01-24
musl builds are broken 


aduskett:
source/posix/thread.c line 175 uses pthread_attr_setaffinity_np (introduced in commit 94efd6e062d2af5d9002674b60e1cbc9def8a581)  which MUSL does not have.

When compiling against a toolchain that uses musl, the following error occurs:
` error: implicit declaration of function ‘pthread_attr_setaffinity_np’; did you mean ‘pthread_setaffinity_np’?`

It should be a fairly straight-forward fix. I thought you guys would like to know!

r-burns:
Can confirm, it's easy to reproduce this in nixpkgs:
`nix-build '<nixpkgs>' -A pkgsStatic.aws-c-common`

A trivial stub definition could allow this to compile, as done here: 
https://github.com/rexray/rexray/issues/719#issuecomment-279506056
Of course, it wouldn't actually set the affinity, so performance would be impacted vs. glibc.

r-burns:
Fyi, this is resolved as of https://github.com/awslabs/aws-c-common/pull/766

aduskett:
> Fyi, this is resolved as of #766

Great! Thank you so much!

PR #755, opened 2021-01-26
Add pthread_setaffinity_np method to support musl builds

r-burns:
Some platforms, such as musl, do not have support for
the nonportable pthread_attr_setaffinity_np function.
Some of those without may be able to fall back to
pthread_setaffinity_np.

We can directly check for support using CMake's CheckSymbolExists,
and enable/disable this functionality accordingly.

Fixes https://github.com/awslabs/aws-c-common/issues/754

(By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.)


source/posix/thread.c
```diff
@@ -160,7 +160,7 @@ int aws_thread_launch(
  * NUMA or not is setup in interleave mode.
  * Thread afinity is also not supported on Android systems, and honestly, if you're running android on a NUMA
  * configuration, you've got bigger problems. */
-#if !defined(__MACH__) && !defined(__ANDROID__)
+#if !defined(__MACH__) && !defined(__ANDROID__) && defined(HAVE_PTHREAD_ATTR_SETAFFINITY_NP)
```
JonathanHenson, line 163: This code is actually vital if you want to hit full NIC entitlement on numa architectures. Are there any ways we can check for the alternative variants and use those where we can? I suspect a lot of Rust and C++ users link against musl and wouldn't like to see a massive performance reduction because of it.

r-burns: Do you mean like falling back to `pthread_setaffinity_np`? I think that extension is supported on musl.

JonathanHenson: Yes, that would work.

JonathanHenson: In fact, that might also work for BSD variants.

r-burns: Okay, the patch is a bit more complicated with this fallback mechanism, but it's definitely valuable to add support for musl and maybe even BSD. I don't have a BSD setup to test on but maybe I can try it out in a VM later this week. Would appreciate if you could take a look at the new revision I pushed when you have the time.

JonathanHenson (approved):

source/posix/thread.c
JonathanHenson: this looks a good deal better. Thanks! now we're going to have to work around github actions and move this to a branch so CI will run. I'll link it here.

cmake/AwsThreadAffinity.cmake
```diff
@@ -0,0 +1,40 @@
...
+
+    set(headers "pthread.h")
+    # BSDs put nonportable pthread declarations in a separate header.
+    if(CMAKE_SYSTEM_NAME MATCHES BSD)
+        set(headers "${headers};pthread_np.h")
+    endif()
+
+    # Using pthread attrs is the preferred method, but is glibc-specific.
+    check_symbol_exists(pthread_attr_setaffinity_np "${headers}" USE_PTHREAD_ATTR_SETAFFINITY)
+    if (USE_PTHREAD_ATTR_SETAFFINITY)
```
JonathanHenson, line 25: can we make these checks mutually exclusive so we don't accidentally use both functions if for some reason an OS distribution does something weird and wonky?

source/posix/thread.c
```diff
@@ -208,6 +210,23 @@ int aws_thread_launch(
         goto cleanup;
     }
 
+#if defined(USE_PTHREAD_SETAFFINITY)
```
JonathanHenson, line 213: can we make these mutually exclusive, just in case?

JonathanHenson: I would say the cmake check being hardened would be enough, but over time we're going to have a few builds that don't go through cmake and instead go through swift.package files and build.rs systems. Those will be emulating everything the cmake currently does and I'd like to prevent problems here if we get that part wrong.

r-burns: Thanks, I was wondering about this while adding these macros. I can see two possible approaches:

1. Add checks such as
#if (defined(X) && defined(Y)
#error
Probably simplest but would combinatorially increase if more mutually exclusive methods are added
2. Treat the method macros like an enum to be matched against possible values, e.g.
#if AWS_AFFINITY_METHOD == AWS_AFFINITY_METHOD_PTHREAD_ATTR

JonathanHenson: I _think_ N is very small here, so I'm not particularly worried about 1. 2. is fine too. I leave it up to you, whatever seems cleanest and easiest to grok for maintainers, and still prevents a misconfiguration step from borking the builds.

source/posix/thread.c
```diff
@@ -25,6 +25,22 @@
 typedef cpuset_t cpu_set_t;
 #endif
 
+#if !defined(AWS_AFFINITY_METHOD)
+#error "Must provide a method for setting thread affinity"
```
r-burns, line 29: This intentionally fails if the method is not specified. It could fall back to NONE, but I considered this more robust as non-cmake build systems can't accidentally build with degraded performance if they are not aware of this.

source/posix/thread.c
```diff
@@ -25,6 +25,22 @@
...
+// Possible methods for setting thread affinity
+#define AWS_AFFINITY_METHOD_NONE         0
+#define AWS_AFFINITY_METHOD_PTHREAD_ATTR 1
+#define AWS_AFFINITY_METHOD_PTHREAD      2
+
+// Ensure provided affinity method matches one of the supported values
+#if AWS_AFFINITY_METHOD != AWS_AFFINITY_METHOD_NONE \
+ && AWS_AFFINITY_METHOD != AWS_AFFINITY_METHOD_PTHREAD_ATTR \
+ && AWS_AFFINITY_METHOD != AWS_AFFINITY_METHOD_PTHREAD
+#error "Invalid thread affinity method"
```
r-burns, line 41: Possibly redundant, but extra checking. Can remove if deemed unnecessary.

source/posix/thread.c
r-burns: I ended up going with method 2, which is a bit more verbose but seems more self-documenting and maintainable.

source/posix/thread.c
ColdenCullen: It's not redundant, because in the above check you only check for existence, not value.

source/posix/thread.c
```diff
@@ -208,6 +224,33 @@ int aws_thread_launch(
...
+    if (options && options->cpu_id >= 0) {
+        AWS_LOGF_INFO(
+            AWS_LS_COMMON_THREAD,
+            "id=%p: cpu affinity of cpu_id %d was specified, attempting to honor the value.",
+            (void *)thread,
+            options->cpu_id);
+
+        cpu_set_t cpuset;
+        CPU_ZERO(&cpuset);
+        CPU_SET((uint32_t)options->cpu_id, &cpuset);
```
ColdenCullen, line 239: This is the same as above, you could move this out of the #if blocks.

r-burns: It needs to be guarded somehow though, I think because not all platforms have support for CPU_ZERO/SET. This was previously handled by the
```cpp
#if !defined(__MACH__) && !defined(__ANDROID__)
```

ColdenCullen: Ah good point.

r-burns:
Friendly ping :)

JonathanHenson (approved):

JonathanHenson:
closing this one so we can run CI and merge. The new PR is here:
https://github.com/awslabs/aws-c-common/pull/766

JonathanHenson:
See the PR runs on this. We missed apple OS family which doesn't have any mechanism to set thread affinity since it's always in interleave mode anyways.

r-burns:
Should I just push here so you can merge into the new PR branch?

JonathanHenson:
Yeah.... unfortunately I'll have to manually merge it due to github security rules and the way github actions works at the moment.

r-burns:
Oh right, the commits don't show up here now that the PR is closed. Anyway, I pushed the fixes to this branch.

JonathanHenson:
okay, your changes have been merged. Thanks!

r-burns:
Awesome, thanks! Will look forward to it in the next release :)

PR #756, opened 2021-01-27
Thread join

bretambrose:
* New managed thread system to support joining of threads that have a detach-like behavior (async cleanup, host resolver)


JonathanHenson:
what problem are you solving that can't be solved with:

```int aws_thread_current_at_exit(aws_thread_atexit_fn *callback, void *user_data)```?

**Comment by bretambrose** [comment]
> what problem are you solving that can't be solved with:
> 
> `int aws_thread_current_at_exit(aws_thread_atexit_fn *callback, void *user_data)`?

Knowing when it's safe to unload a dll that uses aws-c-* functionality.  We have an outstanding issue where the customer determined that simply completing the user-space thread function is insufficient:  (final two entries in) https://github.com/awslabs/aws-crt-cpp/issues/190

As far as I can tell, only completing a join gets you to a point where you know when the thread is truly gone.

File: source/thread_shared.c [hunk 1]
```diff
@@ -0,0 +1,107 @@
...
+}
+
+void aws_thread_decrement_unjoined_count(void) {
+    aws_mutex_lock(&s_managed_thread_lock);
+    --s_unjoined_thread_count;
+    aws_mutex_unlock(&s_managed_thread_lock);
+    aws_condition_variable_notify_one(&s_managed_thread_signal);
+}
+
+bool s_one_or_fewer_managed_threads_unjoined(void *context) {
```
sdavtaker, line 43: is there a typo in the name of this function?

bretambrose: it should be static

source/thread_shared.c
```diff
@@ -0,0 +1,107 @@
...
+
+void aws_thread_pending_join_add(struct aws_linked_list_node *node) {
+    struct aws_linked_list join_list;
+    aws_linked_list_init(&join_list);
+
+    aws_mutex_lock(&s_managed_thread_lock);
+    /*
+     * Swap out the pending join threads before adding this, otherwise we'd join against ourselves which won't work
+     */
+    aws_linked_list_swap_contents(&join_list, &s_pending_join_managed_threads);
```
justinboswell, line 92: So this implies that at most 1 joining thread could be left orphaned in this list until another thread joins or the shutdown is called, is that correct? If it's just 1, I don't care much. If it's more than that, then it would be unbounded and that's a problem. I can't see a way for that to happen, but I just want to confirm I'm reading this right.

Also there should probably be a comment to that effect.

tests/thread_test.c
```diff
@@ -85,3 +86,61 @@ static int s_test_thread_atexit(struct aws_allocator *allocator, void *ctx) {
...
+            aws_thread_launch(&threads[i], s_managed_thread_fn, (void *)&thread_data[i], &thread_options),
+            "thread creation failed");
+
+        aws_thread_clean_up(&threads[i]);
+
+        ASSERT_INT_EQUALS(
+            AWS_THREAD_MANAGED, aws_thread_get_detach_state(&threads[i]), "thread state should have returned JOINABLE");
+    }
+
+    aws_thread_join_all_managed();
```
justinboswell, line 128: This MUST be called from aws_common_library_clean_up() and it should be idempotent. If user code wants to call it, that's fine, but cleaning up the library should ENFORCE it. This test, for instance, should not have to call it directly.

source/thread_shared.c
bretambrose: Yes it's only ever one.

tests/thread_test.c
bretambrose: I'm a little uneasy making this mandatory within cleanup.  The join is unbounded and if the user makes a mistake with a thread it's an infinite wait.

kankri: I haven't reviewed the changes, but this comment just caught my eye. I don't like the idea of the SDK being unsafe by default and it seemed that on [Oct 9th](https://github.com/awslabs/aws-crt-cpp/issues/190#issuecomment-706233116) in  awslabs/aws-crt-cpp#190 you agreed:
>First possibility that jumps to mind is a possible flag in the API object that controls blocking or non-blocking shutdown with a safe (blocking) default.

If the flag is set to unblocking, it would be also nice to have a timeout in the `aws_thread_join_all_managed()` call so that the caller can wait for a bounded time and then bail out in application specific manner, if needed.

bretambrose: Over time, I've found myself going the other way.

An infinite wait is the same as a deadlock and represents a fundamental programming mistake somewhere, either ours or the users.

Waiting to a timeout is just as untrustworthy as not fixing the crash in that if you hit the condition you'll likely crash afterwards when you unload the dll, and if waiting for a timeout was preferable (or fixes the problem, after all we're talking about millisecond or lower level timing issues relative to the dll unload) that could have been done before this change.

We wait in aws_common_library_clean_up now and so if you absolutely cannot do the unbounded wait there'd need to be a new api to control that duration.  At this stage I'd prefer to leave this as is but if you guys feel strongly about the need I'll see if there's any objections to adding it.

justinboswell: This is the safe by default choice. We shut everything we own down before returning. There's no safe way to shutdown with threads still out there accessing shared state, which we are about to clean up synchronously. If there's a deadlock or fundamental programming error, then that is a bug, that's what we have the test suite for. We are going for reliable correctness here, not "how do we let the user do something unsafe when resources have not been cleaned up".

source/thread_shared.c
```diff
@@ -0,0 +1,163 @@
...
+        /*
+         * We lazily join old threads as newer ones finish their thread function.  This means that when called from
+         * the main thread, there will always be one last thread (whichever completion serialized last) that is our
+         * responsibility to join (as long as at least one managed thread was created).  So we wait for a count <= 1
+         * rather than what you'd normally expect (0).
+         *
+         * Absent a timeout, we only terminate if there are no threads left so it is possible to spin-wait a while
+         * if there is a single thread still running.
+         */
+        if (timeout_timestamp_ns > 0) {
```
justinboswell, line 94: Do we have a use case today that requires this? If so, why does it not shut down properly? I'm reticent to put a timeout in here, because that is a guarantee of an inconsistent and unrecoverable state for the library. Since we're doing this to facilitate shared library unloading/reloading, we MUST be in a consistent state when clean up completes.

tests/thread_test.c
bretambrose: Normally I would say wait and see, but given the difficulty/rarity of deploying to a device, any feature added after release may as well not exist.

justinboswell (approved):
Go ahead and unblock the customer, the fact that this is on a separate API is about the best we can do. It's out of the common path, so that'll work for me.

PR #778, opened 2021-03-19
Updates readable and writable macros

feliperodri:

*Issue #, if available:*
N/A.

*Description of changes:*
The current implementation of these macros (using `__CPROVER_*_ok`) should not be used in `*_is_valid` functions.
All validity functions are also used in assumptions for CBMC proofs, which should not contain `__CPROVER_*_ok` primitives. The use of these primitives in assumptions may lead to spurious results.
They can safely be used in assertions or if-statement conditions.


SaswatPadhi (approved):
LGTM

PR #789, opened 2021-04-15
Prevent log-subject ordering errors

graebm:
aws-c-common's log subjects were out of order, so memtracer logged as "xml-parser".

Fix this, and prevent order from ever getting screwed up anywhere ever again.

### DIFF
--- a/source/logging.c
+++ b/source/logging.c
@@ -381,6 +381,17 @@ void aws_register_log_subject_info_list(struct aws_log_subject_info_list *log_su
     const uint32_t min_range = log_subject_list->subject_list[0].subject_id;
     const uint32_t slot_index = min_range >> AWS_LOG_SUBJECT_STRIDE_BITS;
 
+#if DEBUG_BUILD
+    for (uint32_t i = 0; i < log_subject_list->count; ++i) {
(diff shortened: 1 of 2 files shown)


justinboswell (approved):

source/logging.c
```diff
@@ -381,6 +381,17 @@ void aws_register_log_subject_info_list(struct aws_log_subject_info_list *log_su
     const uint32_t min_range = log_subject_list->subject_list[0].subject_id;
     const uint32_t slot_index = min_range >> AWS_LOG_SUBJECT_STRIDE_BITS;
 
+#if DEBUG_BUILD
```
justinboswell, line 384: Actually, would this be better done in a test, so it will fail the build in CI?

graebm: doing it in the register() function means that it will catch the problem in any other repo's subjects, and not just aws-c-common's log subjects 

I'd be fine taking out the `if DEBUG` and doing this in release too

bretambrose (approved):

source/logging.c
graebm: keeping it as is
we chatted offline
this approach keeps a linear scan out of RELEASE, but catches the issue if anyone runs anything in DEBUG

Issue #790, opened 2021-04-16
Event loop stops working after 21 days


kankri:
We have several users having our Windows product using aws-crt-cpp stopped working after a while. I found out that the reason is incorrect handling of timestamp values which makes them invalid after ~21 days of uptime.

The problematic code is in [aws_high_res_clock_get_ticks()](https://github.com/awslabs/aws-c-common/blob/be3e4ff77f9f303262f43c186897967699e03a2f/source/windows/clock.c#L50):
```c
int aws_high_res_clock_get_ticks(uint64_t *timestamp) {
    LARGE_INTEGER ticks, frequency;
    /* QPC runs on sub-microsecond precision, convert to nanoseconds */
    if (QueryPerformanceFrequency(&frequency) && QueryPerformanceCounter(&ticks)) {
        *timestamp = aws_timestamp_convert((uint64_t)ticks.QuadPart, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_MICROS, NULL) /
                     (uint64_t)frequency.QuadPart;

        *timestamp = aws_timestamp_convert(*timestamp, AWS_TIMESTAMP_MICROS, AWS_TIMESTAMP_NANOS, NULL);
        return AWS_OP_SUCCESS;
    }

    return aws_raise_error(AWS_ERROR_CLOCK_FAILURE);
}
```
The problem is the line
```c
        *timestamp = aws_timestamp_convert((uint64_t)ticks.QuadPart, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_MICROS, NULL) /
                     (uint64_t)frequency.QuadPart;
```
which starts by multiplying `ticks` by 1,000,000. This will saturate uint64_t when `ticks` is > `UINT64_MAX` / 1000000 = 18446744073709. With a typical `frequency` of 10,000,000 this amounts to 21.3 days. The visible effect of this is that the time stops at 1844675407370000 ns and the event loop starts repeating the following message without ever reaching the task's timestamp:

    detected more scheduled tasks with the next occurring at 1844675407370000, using timeout of 1000. 

The problematic behavior stops only by rebooting the computer.

As an example of a correct way to handle the conversion, @EeroKurimo found out that QT5 function [ticksToNanoseconds()](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/kernel/qelapsedtimer_win.cpp#n68) (LGPL license) seems to handle the same issue in a way which doesn't have the risk of overflow.

[...]

PR #791, opened 2021-04-16
Avoid overflow on Windows high resolution timer queries 

bretambrose:
https://github.com/awslabs/aws-c-common/issues/790

### DIFF
--- a/source/windows/clock.c
+++ b/source/windows/clock.c
@@ -51,10 +51,15 @@ int aws_high_res_clock_get_ticks(uint64_t *timestamp) {
     LARGE_INTEGER ticks, frequency;
     /* QPC runs on sub-microsecond precision, convert to nanoseconds */
     if (QueryPerformanceFrequency(&frequency) && QueryPerformanceCounter(&ticks)) {
-        *timestamp = aws_timestamp_convert((uint64_t)ticks.QuadPart, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_MICROS, NULL) /
-                     (uint64_t)frequency.QuadPart;
+        uint64_t u64_ticks = ticks.QuadPart;
+        uint64_t u64_frequency = frequency.QuadPart;
 
-        *timestamp = aws_timestamp_convert(*timestamp, AWS_TIMESTAMP_MICROS, AWS_TIMESTAMP_NANOS, NULL);
+        uint64_t seconds = u64_ticks / u64_frequency;
+        uint64_t ticks_remainder = u64_ticks - seconds * u64_frequency;
+
+        *timestamp = aws_timestamp_convert(seconds, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
+        *timestamp += aws_timestamp_convert(ticks_remainder, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL) / u64_frequency;
+
         return AWS_OP_SUCCESS;
     }
 
--- a/source/windows/clock.c
+++ b/source/windows/clock.c
@@ -58,7 +58,8 @@ int aws_high_res_clock_get_ticks(uint64_t *timestamp) {
         uint64_t ticks_remainder = u64_ticks - seconds * u64_frequency;
 
         *timestamp = aws_timestamp_convert(seconds, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
-        *timestamp += aws_timestamp_convert(ticks_remainder, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL) / u64_frequency;
+        *timestamp +=
+            aws_timestamp_convert(ticks_remainder, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL) / u64_frequency;
 
         return AWS_OP_SUCCESS;
     }
--- a/source/windows/clock.c
+++ b/source/windows/clock.c
@@ -58,8 +58,7 @@ int aws_high_res_clock_get_ticks(uint64_t *timestamp) {
         uint64_t ticks_remainder = u64_ticks - seconds * u64_frequency;
 
         *timestamp = aws_timestamp_convert(seconds, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
-        *timestamp +=
-            aws_timestamp_convert(ticks_remainder, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL) / u64_frequency;
+        *timestamp += aws_timestamp_convert_u64(ticks_remainder, u64_frequency, AWS_TIMESTAMP_NANOS, NULL);
 
         return AWS_OP_SUCCESS;
     }
--- a/source/windows/clock.c
+++ b/source/windows/clock.c
@@ -51,14 +51,8 @@ int aws_high_res_clock_get_ticks(uint64_t *timestamp) {
     LARGE_INTEGER ticks, frequency;
     /* QPC runs on sub-microsecond precision, convert to nanoseconds */
     if (QueryPerformanceFrequency(&frequency) && QueryPerformanceCounter(&ticks)) {
-        uint64_t u64_ticks = ticks.QuadPart;
-        uint64_t u64_frequency = frequency.QuadPart;
-
-        uint64_t seconds = u64_ticks / u64_frequency;
-        uint64_t ticks_remainder = u64_ticks - seconds * u64_frequency;
-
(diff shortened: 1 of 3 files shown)


source/windows/clock.c
```diff
@@ -51,10 +51,16 @@ int aws_high_res_clock_get_ticks(uint64_t *timestamp) {
...
-                     (uint64_t)frequency.QuadPart;
+        uint64_t u64_ticks = ticks.QuadPart;
+        uint64_t u64_frequency = frequency.QuadPart;
+
+        uint64_t seconds = u64_ticks / u64_frequency;
+        uint64_t ticks_remainder = u64_ticks - seconds * u64_frequency;
+
+        *timestamp = aws_timestamp_convert(seconds, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
+        *timestamp +=
+            aws_timestamp_convert(ticks_remainder, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL) / u64_frequency;
```
graebm, line 62: aws_add_u64_saturating()

source/windows/clock.c
```diff
@@ -51,10 +51,16 @@ int aws_high_res_clock_get_ticks(uint64_t *timestamp) {
     LARGE_INTEGER ticks, frequency;
     /* QPC runs on sub-microsecond precision, convert to nanoseconds */
     if (QueryPerformanceFrequency(&frequency) && QueryPerformanceCounter(&ticks)) {
```
graebm, line 53: not necessary, but noticed in the [docs](https://docs.microsoft.com/en-us/windows/win32/api/profileapi/nf-profileapi-queryperformancefrequency) that we don't need to query frequency every time:
> The frequency of the performance counter is fixed at system boot and is consistent across all processors. Therefore, the frequency need only be queried upon application initialization, and the result can be cached.

source/windows/clock.c
```diff
@@ -51,10 +51,16 @@ int aws_high_res_clock_get_ticks(uint64_t *timestamp) {
...
-        *timestamp = aws_timestamp_convert((uint64_t)ticks.QuadPart, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_MICROS, NULL) /
-                     (uint64_t)frequency.QuadPart;
+        uint64_t u64_ticks = ticks.QuadPart;
+        uint64_t u64_frequency = frequency.QuadPart;
+
+        uint64_t seconds = u64_ticks / u64_frequency;
+        uint64_t ticks_remainder = u64_ticks - seconds * u64_frequency;
```
graebm, line 58: uint64_t ticks_remainder = u64_ticks % u64_frequency;

source/windows/clock.c
graebm, line 62: this isn't right. the tick_remainder is not in SECS, it's in u64_frequency

graebm: The correct math for converting between different units is pretty tough, we have it in aws_timestamp_convert(). Should we add an alternative aws_timetamp_convert() that takes arbitrary u64 as the `convert_to` and `convert_from` and we could just call it like:
```
uint64-t timestamp = aws_timestamp_convert_any(
    u64_ticks/*timestamp*,
    u64_frequency/*convert_from*/,
    AWS_TIMESTAMP_NANOS/*convert_to*/,
    NULL/*remainder*/);
```

graebm: ok. I see now how your math actually is correct. I didn't see that divide at the end. Although, it might not be correct if the frequency is greater than nanoseconds

still think we're better off adding a generic conversion function

kankri: It is calculating `ticks_remainder` * 1e9 / `u64_frequency` which looks good to me. Mathematically `ticks_remainder`/ `u64_frequency` is in seconds and that can be converted from SECS to NANOS by multplying by 1e9.

bretambrose: Oh hey, I could just do 

*timestamp += aws_timestamp_convert(ticks_remainder, u64_frequency, AWS_TIMESTAMP_NANOS, NULL);

source/windows/clock.c
bretambrose: po-tay-to, po-tah-to

source/windows/clock.c
```diff
@@ -51,10 +51,15 @@ int aws_high_res_clock_get_ticks(uint64_t *timestamp) {
...
+
+        *timestamp = aws_timestamp_convert(seconds, AWS_TIMESTAMP_SECS, AWS_TIMESTAMP_NANOS, NULL);
+        *timestamp += aws_timestamp_convert_u64(ticks_remainder, u64_frequency, AWS_TIMESTAMP_NANOS, NULL);
```
graebm, line 61: you don't need to do all this other math. you can just do:
`*timestamp = aws_timestamp_convert_u64(u64_tick, u64_frequency, AWS_TIMESTAMP_NANOS, NULL);`

bretambrose: Magic

graebm (approved):
☑️✔️

kankri:
The code as merged calculates the timestamp as:

    nanos = timestamp * (1000000000 / frequency)

Using the example frequency of 3MHz from [Acquiring high-resolution time stamps](https://docs.microsoft.com/en-us/windows/win32/sysinfo/acquiring-high-resolution-time-stamps#resolution-precision-accuracy-and-stability) and a timestamp after 30 days uptime this gives:

    nanos = (30 * 24 * 3600 * 3000000) * (1000000000 / 3000000) = 7776000000000 * (1000000000 / 3000000) = 7776000000000 * 333 = 2589408000000000

The correct value is 7776000000000 / 3000000 * 1000000000 = 2592000000000000. The result given by the code is off by 2592 seconds.

The code before the review calculated the result as:

    seconds = u64_ticks / u64_frequency
    ticks_remainder = u64_ticks - seconds * u64_frequency
    nanos = seconds * 1000000000 + ticks_remainder * 1000000000 / u64_frequency

This gives:

    seconds = 2592000
    ticks_remainder = 7776000000000 - 2592000 * 3000000 = 7776000000000 - 7776000000000 = 0
    nanos = 2592000 * 1000000000 + 0 * 1000000000 / 3000000 = 2592000000000000

This version of the code gave the correct answer.

With frequencies like 15MHz or 16MHz the current code would err 20,736s - 25,920s in the same time period while the pre-review code would give the correct results again.

Since `ticks_remainder` < `u64_frequency`, `ticks_remainder * 1000000000` would saturate if `u64_frequency >= UINT64_MAX / 1000000000` = ~18.4GHz which seems an unlikely big frequency, but the effects of saturation would still remain <1s in the pre-review code.

E.g. at 20GHz and 32*2^30 Hz the errors for 1 tick before 30 days would be 0.003 and 0.013.

However, for the current code the result for 20GHz is totally incorrect 2,810,252 ns compared to the correct result 2,592,000,000,000,000 ns and for the second frequency the code crashes with division by zero. This is because calling [`aws_timestamp_convert(ticks, frequency, AWS_TIMESTAMP_NANOS, NULL);`](https://github.com/awslabs/aws-c-common/blob/6d85cb93daa8b1dab363cda0738870cacf169b40/source/windows/clock.c#L55) truncates the `enum aws_timestamp_unit convert_from` parameter to 32 bits.

Calling `aws_timestamp_convert_u64()` directly fixes that problem and makes the result for 20GHz correct but the other is still 798s off.

I prefer the pre-review version. Maybe there should be some tests for these conversions.

bretambrose:
Yeah, should've thought closer about this.  Test-wise we'd  need to add some functions for mocking the Query* functions which shouldn't be difficult although platform-specific mocks are ugly.  I'll rework it this weekend and have it ready by Monday.

kankri:
I believe the the first attempt was good enough: previously the frequency has been roughly 1/1000th of the CPU clock speed as indicated in the document I referred to earlier.  The same document mentions:

> Cases might exist where QueryPerformanceFrequency doesn't return the actual frequency of the hardware tick generator. For example, in many cases, QueryPerformanceFrequency returns the TSC frequency divided by 1024; and on Hyper-V, the performance counter frequency is always 10 MHz when the guest virtual machine runs under a hypervisor that implements the hypervisor version 1.0 interface. As a result, don't assume that QueryPerformanceFrequency will return the precise TSC frequency.

However, it seems that recent Windows 10 versions return 10 MHz also when not running inside Hyper-V. So it seems we aren't getting past 10 MHz anytime soon so the code supporting frequencies up to 18 GHz seems to have big enough headroom.

An alternative would be to calculate `ticks * 1000000000 / freq` using [_umul128](https://docs.microsoft.com/en-us/cpp/intrinsics/umul128?view=msvc-160) and [_udiv128](https://docs.microsoft.com/en-us/cpp/intrinsics/udiv128?view=msvc-160), however the latter is only available since VS2019 so it should be implemented by hand to support earlier compilers.

As for testing, if `aws_high_res_clock_get_ticks()` just calls `QueryPerformanceFrequency()` and `QueryPerformanceCounter()` and passes their values to a function which does all the calculations, it should be enough to test the latter function. Maybe that is easier.

PR #792, opened 2021-04-19
Fix cmake script to respect the CMAKE_INSTALL_LIBDIR / CMAKE_INSTALL_BINDIR settings for all platforms

mbalfour-amzn:
*Description of changes:*
The build file was setting LIBRARY_DIRECTORY and RUNTIME_DIRECTORY to defaults, and then overwrote the defaults with the values of CMAKE_INSTALL_LIBDIR / CMAKE_INSTALL_BINDIR only for UNIX non-Apple builds, and regardless of whether or not those values were set.

This change sets the LIBRARY_DIRECTORY / RUNTIME_DIRECTORY values to use CMAKE_INSTALL_LIBDIR / CMAKE_INSTALL_BINDIR for *all* platforms, unless the value doesn't exist, at which point it uses the default.

NOTE:  I will also submit an identical change to aws-checksums, which has the same error.

### DIFF
--- a/cmake/AwsSharedLibSetup.cmake
+++ b/cmake/AwsSharedLibSetup.cmake
@@ -1,13 +1,21 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0.
 
-set(LIBRARY_DIRECTORY lib)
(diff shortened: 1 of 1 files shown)


graebm:
Fixing in a new PR...

PR #794, opened 2021-04-21
Aws fopen

TingDaoK:
*Issue #, if available:*

*Description of changes:*


tests/file_open_test.c
```diff
@@ -0,0 +1,58 @@
...
+    size_t read_len = fread(read_result, sizeof(char), strlen(content), readfile);
+    ASSERT_UINT_EQUALS(strlen(content), read_len);
+    fclose(readfile);
+    ASSERT_SUCCESS(strcmp(content, read_result));
+
+#ifdef _WIN32
+    wchar_t w_file_path[1000];
+    /* plus one for the EOS */
+    size_t file_path_len = strlen(file_path) + 1;
+    MultiByteToWideChar(CP_UTF8, 0, file_path, -1, w_file_path, (int)file_path_len);
```
TingDaoK, line 28: I don't like this... It's still use the same function we want to check it's correct or not... But, I don't know what else can we do to check the result... An alternative I can think of is to use `[WideCharToMultiByte](WideCharToMultiByte )`, feels like it's still basically the same...

source/file_open.c
```diff
@@ -0,0 +1,30 @@
...
+    wchar_t w_file_path[1000];
+
+    /* the default encoding is utf-8 or ascii */
+    MultiByteToWideChar(CP_UTF8, 0, file_path, -1, w_file_path, (int)(strlen(file_path) + 1));
+    wchar_t w_mode[10];
+    MultiByteToWideChar(CP_UTF8, 0, mode, -1, w_mode, (int)(strlen(mode) + 1));
+    FILE *file;
+    if (_wfopen_s(&file, w_file_path, w_mode)) {
+        /* errno will be set */
+        return NULL;
```
graebm, line 24: Is all of this really necessary? seems like [fopen_s() supports utf-8](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-s-wfopen-s?view=msvc-160#unicode-support):
> fopen_s supports Unicode file streams. To open a new or existing Unicode file, pass a ccs flag that specifies the desired encoding to fopen_s

tests/file_open_test.c
graebm: you could just commit a file with a utf-8 name to our repo and have the test open THAT

same way we have files committed to `aws-c-io/tests/resources/*`and the aws-c-io tests use those

graebm: also, you don't need to worry about cleaning up the test files.
we have tons of tests that create local domain sockets and just leave the files on disk, in the test bin directory

graebm: on second thought... do clean up the files so running the results of one test run don't interfere the NEXT time you run the tests. Those domain sockets are OK because the OS makes sure they're uniquely named.

or randomize the filename to prevent collisions, use [tmpname()](https://en.cppreference.com/w/c/io/tmpnam) or something

graebm:
As a followup you'll need to find/replace ALL uses of `fopen()` in our codebase.

You can start here in aws-c-common though. There are several uses of `fopen()` in aws-c-common

source/file_open.c
TingDaoK: I think it's saying the content of the file instead of file name? And I just tested it with [fopen_s](https://github.com/awslabs/aws-c-common/pull/794/commits/dfec173a5990016107f87e3d15435be5c4020484), it failed on windows

tests/file_open_test.c
TingDaoK: sounds good, I'll add another test to read an existing file, and keep those tests for testing writing

source/file_open.c
```diff
@@ -0,0 +1,30 @@
...
+
+#include <aws/common/file_open.h>
+#include <stdio.h>
+#ifdef _WIN32
+#    include <windows.h>
+#endif
+
+FILE *aws_fopen(const char *file_path, const char *mode) {
+
+#ifdef _WIN32
```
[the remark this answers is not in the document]
TingDaoK: 😂but in this case, for the posix, it will be just call `fopen()`. Seems a little weird.

source/windows/file.c
```diff
@@ -0,0 +1,24 @@
...
+#include <aws/common/file.h>
+#include <stdio.h>
+#include <windows.h>
+
+FILE *aws_fopen(const char *file_path, const char *mode) {
+
+    wchar_t w_file_path[1000];
+
+    /* the default encoding is utf-8 or ascii */
+    MultiByteToWideChar(CP_UTF8, 0, file_path, -1, w_file_path, (int)(strlen(file_path) + 1));
```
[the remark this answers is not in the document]
graebm: also
!!! 💣 BUFFER OVERFLOW 💣 !!!
last argument is
> cchWideChar: Size, in characters, of the buffer indicated by lpWideCharStr

you should be passing 1000 (or `AWS_ARRAY_SIZE(w_file_path)`). The way you have it now MultiByteToWideChar() could write beyond the end of w_file_path and corrupt the stack. This is a very dangerous [vulnerability](http://kylehalladay.com/blog/2019/12/04/Recreating-A-Dirty-Gamedev-Hack.html)!!!

source/windows/file.c
```diff
@@ -0,0 +1,24 @@
...
+    wchar_t w_mode[10];
+    MultiByteToWideChar(CP_UTF8, 0, mode, -1, w_mode, (int)(strlen(mode) + 1));
```
graebm, line 17: same feedback as the other call

DavidOgunsAWS (commented):
A reasonable addition might be a Windows overload for aws_fopen that directly has wchar_t parameters so the conversion doesn't have to happen if the caller already has wchar_t strings around.

`FILE *aws_fopen(const wchar_t *file_path, const w_chart *mode)` for Window-only so `aws_fopen(L"file.in", L"r")` can be written in MSVC land without conversion coming into the picture.

source/windows/file.c
```diff
@@ -0,0 +1,32 @@
...
+
+    /* the default encoding is utf-8 or ascii */
+    if (!MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, file_path, -1, w_file_path, AWS_ARRAY_SIZE(w_file_path))) {
+        /* When error happens, we need to set errno to invalid argument, since the function will set the Windows
+         * specific error that we don't handle */
+        errno = EINVAL;
+        return NULL;
+    }
+    wchar_t w_mode[10];
+    if (!MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, mode, -1, w_mode, AWS_ARRAY_SIZE(w_file_path))) {
```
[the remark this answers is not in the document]
TingDaoK: good catch... My bad😵

graebm (approved):

PR #796, opened 2021-04-30
Diagnostics fixes/improvements

justinboswell:
* aws_backtrace_log now takes a log level instead of forcing TRACE
* aws_backtrace_print no longer forces the use of addr2line or atos which produces crap output most of the time
* test_stack_trace_decoding is now more permissive for linkers that do weird things to symbols/debug info

### DIFF
--- a/tests/system_info_tests.c
+++ b/tests/system_info_tests.c
@@ -37,8 +37,8 @@ static int s_test_stack_trace_decoding(struct aws_allocator *allocator, void *ct
     test_logger_init(&test_log, allocator, AWS_LL_TRACE, 0);
     aws_logger_set(&test_log);
 
-    int line = 0;                           /* captured on line of aws_backtrace_log call to match call site */
-    (void)line;                             /* may not be used if debug info is unavailable */
+    int line = 0; /* captured on line of aws_backtrace_log call to match call site */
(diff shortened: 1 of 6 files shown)


tests/system_info_tests.c
```diff
@@ -59,30 +59,36 @@ static int s_test_stack_trace_decoding(struct aws_allocator *allocator, void *ct
...
+        char fileline[4096];
+        uint32_t found_file_line = 0;
+        for (int lineno = line - 1; lineno <= line + 1; ++lineno) {
+            snprintf(fileline, sizeof(fileline), "%s:%d", file, lineno);
+            found_file_line |= strstr((const char *)buffer->buffer, fileline) != NULL;
+            if (found_file_line) {
+                break;
+            }
+        }
+        if (!found_file_line) {
```
DavidOgunsAWS, line 85: Would this be more straight forward to read using `bool found_file_line` with a normal assignment operator if the line is found? I don't see a purpose for `|=` on line 80, and all reads of this variable seem to be treating it like a boolean rather than using the address returned by strstr()

DavidOgunsAWS (approved):

tests/system_info_tests.c
justinboswell: Good point.

PR #798, opened 2021-05-07
Adding CMake option to disable warnings as errors

DavidOgunsAWS:
*Description of changes:*
Developers using newer compilers than CI seeing existing code generate warnings and turn them into errors shouldn't block a customer. This makes it easier for a developer to build anyways without editing the CMake file directly to achieve the effect, but defaults to treating all warnings as errors for general code hygiene where we are using specific compilers in CI or locally.

### DIFF
--- a/cmake/AwsCFlags.cmake
+++ b/cmake/AwsCFlags.cmake
@@ -8,6 +8,7 @@ include(CMakeParseArguments) # needed for CMake v3.4 and lower
 option(AWS_ENABLE_LTO "Enables LTO on libraries. Ensure this is set on all consumed targets, or linking will fail" OFF)
 option(LEGACY_COMPILER_SUPPORT "This enables builds with compiler versions such as gcc 4.1.2. This is not a 'supported' feature; it's just a best effort." OFF)
 option(AWS_SUPPORT_WIN7 "Restricts WINAPI calls to Win7 and older (This will have implications in downstream libraries that use TLS especially)" OFF)
+option(WARNINGS_ARE_ERRORS "Compiler warning is treated as an error. Try turning this off when observing errors on a new or unpopular compiler" ON)
 
 # This function will set all common flags on a target
 # Options:
@@ -53,7 +54,11 @@ function(aws_set_common_properties target)
         list(APPEND AWS_C_FLAGS "${_FLAGS}")
 
     else()
-        list(APPEND AWS_C_FLAGS -Wall -Werror -Wstrict-prototypes -fno-omit-frame-pointer)
+        list(APPEND AWS_C_FLAGS -Wall -Wstrict-prototypes -fno-omit-frame-pointer)
+
+        if(WARNINGS_ARE_ERRORS)
+	    list(APPEND AWS_C_FLAGS -Werror)
+        endif()
 
         if(NOT SET_PROPERTIES_NO_WEXTRA)
             list(APPEND AWS_C_FLAGS -Wextra)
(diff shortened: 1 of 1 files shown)


cmake/AwsCFlags.cmake
```diff
@@ -25,7 +26,12 @@ function(aws_set_common_properties target)
             set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}" PARENT_SCOPE)
         endif()
 
-        list(APPEND AWS_C_FLAGS /W4 /WX /MP)
+        list(APPEND AWS_C_FLAGS /W4 /MP)
+
+        if(WARNINGS_ARE_ERRORS)
+	    list(APPEND AWS_C_FLAGS /WX)
```
justinboswell, line 32: White space looks mixed here

cmake/AwsCFlags.cmake
```diff
@@ -53,7 +59,11 @@ function(aws_set_common_properties target)
         list(APPEND AWS_C_FLAGS "${_FLAGS}")
 
     else()
-        list(APPEND AWS_C_FLAGS -Wall -Werror -Wstrict-prototypes -fno-omit-frame-pointer)
+        list(APPEND AWS_C_FLAGS -Wall -Wstrict-prototypes -fno-omit-frame-pointer)
+
+        if(WARNINGS_ARE_ERRORS)
+	    list(APPEND AWS_C_FLAGS -Werror)
```
justinboswell, line 65: Same here

justinboswell (approved):
Fix-n-ship

cmake/AwsCFlags.cmake
```diff
@@ -8,6 +8,7 @@ include(CMakeParseArguments) # needed for CMake v3.4 and lower
 option(AWS_ENABLE_LTO "Enables LTO on libraries. Ensure this is set on all consumed targets, or linking will fail" OFF)
 option(LEGACY_COMPILER_SUPPORT "This enables builds with compiler versions such as gcc 4.1.2. This is not a 'supported' feature; it's just a best effort." OFF)
 option(AWS_SUPPORT_WIN7 "Restricts WINAPI calls to Win7 and older (This will have implications in downstream libraries that use TLS especially)" OFF)
+option(WARNINGS_ARE_ERRORS "Compiler warning is treated as an error. Try turning this off when observing errors on a new or unpopular compiler" ON)
```
graebm, line 11: trivial: should we namespace this option with "AWS_"?
AWS_WARNINGS_ARE_ERRORS?

cmake/AwsCFlags.cmake
```diff
@@ -8,6 +8,7 @@ include(CMakeParseArguments) # needed for CMake v3.4 and lower
 option(AWS_ENABLE_LTO "Enables LTO on libraries. Ensure this is set on all consumed targets, or linking will fail" OFF)
 option(LEGACY_COMPILER_SUPPORT "This enables builds with compiler versions such as gcc 4.1.2. This is not a 'supported' feature; it's just a best effort." OFF)
 option(AWS_SUPPORT_WIN7 "Restricts WINAPI calls to Win7 and older (This will have implications in downstream libraries that use TLS especially)" OFF)
```
graebm, line 10: I vote we turn this OFF by default, and fast-follow this PR with changes to our CI so it sets the flag.


I just think philosophically, a default value should be aimed at an external user. If an option is there to help developers of THIS library, then it should be opt in.

cmake/AwsCFlags.cmake
DavidOgunsAWS: Makes sense. It is a very general of a name that could clash with another project's naming

cmake/AwsCFlags.cmake
DavidOgunsAWS: I want to avoid the window of disabling warnings as errors for downstream consumers, and not have to preemptively add turning it on in those consumers.

graebm: We can tackle this in the future (I just want that future now 😢)

PR #806, opened 2021-05-21
Let's get started moving to Rust!

JonathanHenson:
aws-c-common now builds as a standalone crate along with a builder that will let other libraries link their c-bits together across crate boundaries.

*Issue #, if available:*

*Description of changes:*


aws_crt_common_ws/aws-crt-c-flags/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "aws-crt-c-flags"
+version = "0.1.0"
+authors = ["AWS Common Runtime Team (aws-sdk-common-runtime-team@amazon.com)"]
```
rcoh, line 4:
```suggestion
authors = ["AWS Common Runtime Team <aws-sdk-common-runtime-team@amazon.com>"]
```

.gitignore
```diff
@@ -67,3 +67,7 @@ cmake-build*
 # CBMC files
 *.goto
 *.log
+
+# Rust and Cargo
+*.lock
+aws_crt_common_ws/target
```
rcoh, line 73: probably want `target/` — if you build in a submodule you may get another target dir

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+        let crt_module_env_var_res = env::var("CRT_MODULE_".to_owned() + dependency + "BUILD_CFG");
+
+        if crt_module_env_var_res.is_ok() {
+            let parse_res: Result<CRTModuleBuildInfo> =
+                serde_json::from_str(crt_module_env_var_res.unwrap().as_str());
+
+            if parse_res.is_ok() {
+                self.crt_module_deps.push(parse_res.unwrap());
+                return self;
+            }
```
rcoh, line 85:
```suggestion
            if let Ok(parse_res) = parse_res {
                self.crt_module_deps.push(parse_res);
                return self;
            }
```

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+        }
+
+        panic!("Module: ".to_owned() + dependency + " does not appear to be a part of your dependency chain. Alternatively, the dependencies build script does not call CRTModuleBuildInfo::run_build() correctly");
```
rcoh, line 88: this will be invalid in Rust 2021
```suggestion
        panic!("Module: `{}` does not appear to be a part of your dependency chain. Alternatively, the dependencies build script does not call CRTModuleBuildInfo::run_build() correctly", dependency);
```

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+    ///
+    /// * `c_flag` - compiler flag to apply to the build. Example `-fPIC`
+    ///
+    /// # Examples
+    /// ```should_panic
+    /// use aws_crt_c_flags::{CRTModuleBuildInfo};
+    /// let mut build_info = CRTModuleBuildInfo::new("aws_crt_common_sys");
+    /// build_info.add_public_cflag("-fPIC");
+    /// ```
+    pub fn add_public_cflag(&mut self, c_flag: &str) -> &mut CRTModuleBuildInfo {
```
rcoh, line 104: nit: other Rust append builders frequently omit the add here (eg. `http::Request::builder().header("key", "value").build();`)

aws_crt_common_ws/aws-crt-common-sys/build.rs
```diff
@@ -0,0 +1,337 @@
...
+        .write_generated_file_to_output_path(config_file_str.to_string().borrow(), output_path);
+}
+fn main() {
+    let mut build_info = CRTModuleBuildInfo::new("aws-c-common");
+    load_compiler_configuration(build_info.borrow_mut());
+    let include_path = Path::new("../../include/aws");
+    build_info.add_include_dir_and_copy_to_build_tree(include_path);
+
+    build_info
+        .add_file_to_build(Path::new("../../source/allocator.c"))
```
rcoh, line 181: is there every file in `source`? Might want https://docs.rs/include_dir/0.6.0/include_dir/

rcoh (commented):
Skimmed and left a couple ideas. Might want to run `cargo clippy` it will probably have a lot of ideas.

.gitignore
```diff
@@ -67,3 +67,7 @@ cmake-build*
 # CBMC files
 *.goto
 *.log
+
+# Rust and Cargo
+*.lock
```
carllerche, line 72: There should only be a `Cargo.lock` at the root assuming you use workspaces (which it looks like you do).

.gitignore
justinboswell: +1

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+    ///                  if you're building aws-checksums, your sys crate would be aws_crt_checksums_sys, and you
+    ///                  would declare a dependency here on aws_crt_common_sys.
+    ///
+    /// # Examples
+    /// ```should_panic
+    /// use aws_crt_c_flags::{CRTModuleBuildInfo};
+    /// let mut build_info = CRTModuleBuildInfo::new("aws_crt_checksums_sys");
+    /// build_info.add_module_dependency("aws_crt_common_sys");
+    /// ```
+    pub fn add_module_dependency(&mut self, dependency: &str) -> &mut CRTModuleBuildInfo {
```
justinboswell, line 75: This looks like a builder, and @rcoh 's feedback seems to echo that, if you remove all the `add_`, this should be `with_dependency` or similar

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
carllerche: Also, you can use `&mut Self` as the return type. It is a little bit less typing, but it depends on your personal preferences.

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// # Examples
+    /// ```should_panic
+    /// use aws_crt_c_flags::{CRTModuleBuildInfo};
+    /// use std::path::Path;
+    /// let mut build_info = CRTModuleBuildInfo::new("aws_crt_common_sys");
+    /// build_info.add_third_party_include_dir(Path::new("/opt/where/i/installed/my/manually/built/libcrypto/include"))
+    ///     .set_linker_search_path(Path::new("/opt/where/i/installed/my/manually/built/libcrypto/lib"))
+    ///     .add_link_target("crypto");
+    /// ```
+    pub fn add_third_party_include_dir(&mut self, dir: &Path) -> &mut CRTModuleBuildInfo {
```
justinboswell, line 234: Another way to do this might be to have `add_include_dir` that takes an enum(str) of include directory types (`e.g. External` (third party), `Private`(only needed to build), `Public` (implies copy to output)), which would be more semantically clear

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+            env::var_os("OUT_DIR").unwrap().to_str().unwrap()
+        );
+        fs::write(Path::new(&target_location.as_str()), to_compile).expect("File write failed");
+        test_build.file(&target_location);
+        let res = test_build.try_compile("test");
+        fs::remove_dir_all(&output_location).expect("Cleanup of try compile step failed!");
+        res
+    }
+
+    fn load_to_build(&mut self) {
```
justinboswell, line 361: After reading it, isn't this just `configure_cc_toolchain` or similar?

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+    ///```should_panic
+    /// use aws_crt_c_flags::{CRTModuleBuildInfo};
+    /// use std::path::Path;
+    /// let mut build_info = CRTModuleBuildInfo::new("aws_crt_common_sys");
+    /// let content = "test content".to_string();
+    /// build_info.write_generated_file_to_output_path(&content, Path::new("include/aws/common/config.h"));
+    ///```
+    pub fn write_generated_file_to_output_path(
+        &mut self,
+        generated_content: &String,
```
carllerche, line 299: `&str`? I couldn't see a reason for `&String` here.

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+                .add_private_cflag("-Werror")
+                .add_private_cflag("-Wstrict-prototypes")
+                .add_private_cflag("-fno-omit-frame-pointer")
+                .add_private_cflag("-Wextra")
+                .add_private_cflag("-pedantic")
+                .add_private_cflag("-Wno-long-long")
+                .add_private_cflag("-fPIC");
+        }
+
+        if self.build_toolchain.is_flag_supported("-Wgnu").is_ok() {
```
justinboswell, line 384: Don't have to do this now, but I feel like we should be able to encapsulate these checks in their own individual flag structs that implement a `CFlag` (or whatever) trait that encapsulates the check and the resultant flag per platform, and slap the flags on the `CRTModuleBuildInfo`

justinboswell: Like as a consumer/user of this, I just want to opt-in to `WarningsAsErrors` or `PositionIndependentCode` or `NoGNUZeroVariadicMacroArguments`, etc.

```rust
let module = CRTModuleBuildInfo::new()
    .with_cflags(WarningsAsErrors)
    .with_cflags(PositionIndependentCode)
...
```

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+    ///
+    /// * `to_compile` - C code to attempt compilation of.
+    ///
+    /// # Examples
+    /// ```should_panic
+    /// use aws_crt_c_flags::{CRTModuleBuildInfo};
+    /// let mut build_info = CRTModuleBuildInfo::new("aws_crt_common_sys");
+    /// build_info.try_compile("int main() { return 0; }").expect("This should have compiled");
+    /// ```
+    pub fn try_compile(&self, to_compile: &str) -> core::result::Result<(), cc::Error> {
```
carllerche, line 339: I found the `core` usage here confusing at first. Usually `core` is referenced when `std` is not being used. `std::result::Result` is a re-export of `core::result::Result` so they are the same types.

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+            );
+        }
+
+        for link_flag in &self.link_targets {
+            println!("cargo:rustc-link-lib={}", link_flag)
+        }
+
+        let module_name_cpy = self.crt_module_name.clone();
+        env::set_var(
+            "CRT_MODULE_".to_owned() + module_name_cpy.as_str() + "BUILD_CFG",
```
justinboswell, line 465: This has the same problem as the string @rcoh commented on above, use `"{}"`

aws_crt_common_ws/aws-crt-common-sys/build.rs
```diff
@@ -0,0 +1,337 @@
...
+ * configure time to inline function implementations. The macros defined here
+ * should be considered to be an implementation detail, and can change at any
+ * time.
+ */\n"
+        .to_string();
+
+    build_config.add_public_define("AWS_NO_STATIC_IMPL", "1");
+    config_file_str = format!("{}#define AWS_NO_STATIC_IMPL 1\n", config_file_str);
+
+    #[target_feature(enable = "avx2")]
```
justinboswell, line 25: This doesn't work in a cross-compile situation. Do we need to worry about that?

aws_crt_common_ws/aws-crt-common-sys/build.rs
```diff
@@ -0,0 +1,337 @@
...
+    {
+        build_config.add_private_define("HAVE_AVX2_INTRINSICS", "1");
+        build_config.add_private_define("HAVE_MM256_EXTRACT_EPI64", "1");
+
+        if build_config.get_toolchain().get_compiler().is_like_msvc() {
```
justinboswell, line 30: This is something that should be encapsulated like the cflags stuff above, we're going to end up with nests of this all over otherwise.

aws_crt_common_ws/aws-crt-common-sys/build.rs
```diff
@@ -0,0 +1,337 @@
...
+            }
+        }
+
+        build_info
+            .add_link_target("Kernel32")
+            .add_link_target("BCrypt")
+            .add_link_target("Ws2_32");
+    }
+
+    #[cfg(not(Windows))]
```
carllerche, line 249: Is it not case-sensitive? I would not expect this to match on windows.

aws_crt_common_ws/aws-crt-common-sys/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+        allocator: *const CAllocator,
+        deprecated: *const CAllocator,
+        level: TraceLevel,
+        stack_frame_size: usize,
+    ) -> *mut CAllocator;
+    pub fn aws_mem_tracer_destroy(allocator: *const CAllocator) -> *mut CAllocator;
+    pub fn aws_mem_tracer_bytes(allocator: *const CAllocator) -> usize;
+    pub fn aws_default_allocator() -> *const CAllocator;
+
+    pub fn aws_string_new_from_array(
```
justinboswell, line 116: Do we want to separate these into separate files/modules within this one? e.g. list.rs, string.rs, and then `pub mod` them here? For maintainability, and for namespacing? Or do we just want to single-file yolo this? I'd love @rcoh or @carllerche's opinion on this for the long term

source/clock.c
```diff
@@ -0,0 +1,8 @@
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#if AWS_NO_STATIC_IMPL
+#    include <aws/common/clock.inl>
+#endif
```
JonathanHenson, line 8:
```suggestion
#endif

```

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+///
+#[derive(Clone)]
+pub struct CallbackFuture<T>
+where
+    T: Clone,
+{
+    data: Arc<Mutex<FutureSharedState<T>>>,
+}
+
+impl<T> FutureSharedState<T>
```
justinboswell, line 79: Isn't this basically just a promise? How does it resolve when an error occurs?

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
JonathanHenson: that was a clion insert thing. I'll move it over. good catch!

aws_crt_common_ws/aws-crt-common-sys/build.rs
JonathanHenson: good catch!

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+    const RUST_ERRORS: [CErrorInfo; 1] = [error_info!(INVALID_UTF8, "String was not valid UTF-8")];
+
+    pub const CRUST_ERRORS: CErrorInfoList = CErrorInfoList {
+        error_list: RUST_ERRORS.as_ptr(),
+        count: RUST_ERRORS.len() as u16,
+    };
+}
+
+#[cfg(test)]
+mod tests {
```
carllerche, line 482: I try to put as many tests as possible *outside* of `src`. If a test exclusively uses public APIs, it can live in a `tests` directory. The reason I like to do this is, when the test is in the `src` directory, changing a test requires recompiling the entire library. This probably does not matter for this lib right now as it is pretty small, but it is something to keep in mind.

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+    // it will point to static memory.
+    let error_str = unsafe { aws_error_debug_str(error_code) };
+    // convert the underlying c string memory into a rust representation of it.
+    // error_str cannot be null.
+    let error_c_str = unsafe { CStr::from_ptr(error_str) };
+    error_c_str.to_str()
+}
+
+/// Returns the error represented by error_code as a string.
+pub fn error_str(error_code: i32) -> Result<&'static str, Utf8Error> {
```
justinboswell, line 195: Please also include `error_name`

aws_crt_common_ws/aws-crt-common/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+authors = ["AWS Common Runtime Team (aws-sdk-common-runtime-team@amazon.com)"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[build-dependencies]
+
+[dependencies]
+aws-crt-common-sys = { path = "../aws-crt-common-sys" }
+futures = "0.3"
```
carllerche, line 13: It *looks* like you are only using `Waker`. `futures::task::Waker` is a re-export of `std::task::Waker`, so you should not need the `futures` crate here. The `futures` crate is pretty heavy. If you *do* need functionality only provided by `futures`, you probably will want to depend on the specific sub-crates (e.g. `futures-core`).

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+///
+/// Upon creation call, CallbackFuture::to_callback_ptr() to get the pointer to pass to C functions.
+/// When your callback completes, fill-in an instance of type T, and call FutureSharedState::resolve()
+/// with the user_data pointer from your c callback.
+///
+/// Assuming you do these things, you've just converted C-style async code into co-routine ready
+/// Rust futures.
+///
+#[derive(Clone)]
+pub struct CallbackFuture<T>
```
carllerche, line 72: This is pretty much Tokio's [`oneshot` channel](https://docs.rs/tokio/1.6.0/tokio/sync/oneshot/index.html). The only difference is you have a conversion to raw pointers. We could add to/from raw pointers to the Tokio channel as this is a pretty common pattern when bridging to C. It also is possible to only depend on Tokio's sync primitives w/o pulling in the rest if that is a path you want to explore.

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+    /// Returns the length of the string in bytes, not including any null terminators.
+    pub fn length(&self) -> usize {
+        // This is sound if you have an instance of this class.
+        unsafe {
+            return (*self.c_str).len;
+        }
+    }
+
+    /// Returns a pointer to an array of utf8 characters for the underlying String
+    pub fn bytes(&self) -> *const u8 {
```
carllerche, line 381: Idiomatically, this would be named `as_ptr()`

carllerche: e.g. https://doc.rust-lang.org/stable/std/string/struct.String.html#method.as_ptr

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+pub struct AwsString {
+    c_str: *const CString,
+    owned: bool,
+}
+
+impl AwsString {
+    /// Creates a new String for a Rust std::string::String instance.
+    ///
+    /// Returns a result containing the new String or an int-based error code on failure.
+    pub fn new_from_std_string(
```
carllerche, line 318: `fn from_*` would be the rust idiom e.g. `from_std_str`).

aws_crt_common_ws/aws-crt-common-sys/src/lib.rs
rcoh: for just the header style sys module I think this is fine. When we have bigger crates down the line, we might need to separate stuff out

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+    /// use aws_crt_common::AwsString;
+    /// let allocator = todo!();
+    /// let aws_str = AwsString::new_from_str(allocator, "Hello")
+    ///                       .expect("Creation of c string failed!");
+    /// let ptr = aws_str.c_str();
+    /// unsafe {
+    ///     *ptr;
+    /// }
+    /// ```
+    pub fn c_str(&self) -> *const c_char {
```
carllerche, line 418: `fn as_c_str()`

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+        // it's specified as the default_allocator, it doesn't need to be cleaned up.
+        unsafe {
+            let _ = aws_mem_tracer_destroy(self.c_allocator);
+        }
+    }
+}
+
+/// String type which wraps the internal aws_string structure. This type is immutable.
+#[non_exhaustive]
+pub struct AwsString {
```
carllerche, line 309: You will probably want to add `Deref` impls with `str` as the target. You will get a bunch of methods for "free". See: https://doc.rust-lang.org/stable/std/string/struct.String.html#deref-methods

You will get `as_ptr()` as well, which you implemented ass `bytes()`.

aws_crt_common_ws/aws-crt-common/src/lib.rs
rcoh: TIL, that's clever.

aws_crt_common_ws/aws-crt-common/src/lib.rs
carllerche: You won't actually need this fn if you implement `Deref<Target = str>`

carllerche: Actually, this type does not contain `UTF-8` data, so you will not want to deref to `str`, instead you can deref to `[u8]`.

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+    c_str: *const CString,
+    owned: bool,
+}
```
rcoh, line 312: at some point I had a PR that refactored this into an `OwnedAwsString` and a `BorrowedAwsString` — it simplified the code a bit, may want to do the same thing here.

Alternatively, could have this be an `enum` like `Cow`

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+        // This is sound if you have an instance of this class.
+        unsafe {
+            return aws_string_bytes(self.c_str) as *const c_char;
+        }
+    }
+}
+
+impl ToString for AwsString {
+    fn to_string(&self) -> String {
+        // this is sound if you have an instance of this class and the original AwsString is valid utf-8
```
rcoh, line 428: is there somewhere that we guarantee that the `AwsString` can only be created from valid UTF-8? Otherwise this is UB.

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
+/*
```
rcoh, line 1: probably makes sense to split this into 3 modules for errors, strings, and allocators

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+/// Trait that gives a c-style const pointer to an underlying resource.
+/// Ideally you use this when calling an ffi function that needs a pointer
+/// to your wrapped Rust object.
+pub trait NativeResource<T> {
+    fn get_native_resource(&self) -> *const T;
+}
+
+pub struct FutureSharedState<T>
+where
+    T: Clone,
```
rcoh, line 55: the `T: Clone` bound seems like it could be problematic

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+    type Output = T;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut option_val: MutexGuard<FutureSharedState<T>> = self.data.lock().unwrap();
+        if option_val.data.is_none() {
+            option_val.waker = Some(cx.waker().clone());
+            return Poll::Pending;
+        }
+
+        return Poll::Ready(option_val.data.as_ref().expect("We're in a ready state implying the future has completed but the underlying value hasn't been set!").clone());
```
rcoh, line 114: use `option_val.data.take()` and remove the `Clone` bound

rcoh: https://github.com/awslabs/aws-crt-rs/compare/russell/aws-string-safety?expand=1#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L101-L112

aws_crt_common_ws/aws-crt-common/src/lib.rs
rcoh: https://github.com/awslabs/aws-crt-rs/compare/russell/aws-string-safety

(remove non-owned variants of AwsString, they don't actually end up being used)

aws_crt_common_ws/aws-crt-common-sys/build.rs
JonathanHenson: it's not every file in source and the files are often platform specific, so in this case I'd prefer to keep it explicit.

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
rcoh: this is slightly different (although I'd use the format macro here personally). The problem I mentioned above is specific to `panic!` which will only accept static strings as the first argument in Rust 2021

aws_crt_common_ws/aws-crt-c-flags/src/lib.rs
JonathanHenson: i honestly don't think that's any more descriptive or clear.

aws_crt_common_ws/aws-crt-common-sys/build.rs
JonathanHenson: wait how? target_feature is derived from the target arch of the compile run.

aws_crt_common_ws/aws-crt-common-sys/src/lib.rs
JonathanHenson: later

aws_crt_common_ws/aws-crt-common/src/lib.rs
JonathanHenson: https://docs.rs/futures/0.3.15/futures/future/trait.Future.html

JonathanHenson: It resolves to whatever you want. in the case of something that can fail, it stores a Result<T, Error> object.

aws_crt_common_ws/aws-crt-common/src/lib.rs
JonathanHenson: huh?

aws_crt_common_ws/aws-crt-common/src/lib.rs
JonathanHenson: yeah, this con't compile, and the docs got nothing. I'll come back to it at a later date.

aws_crt_common_ws/aws-crt-common/src/lib.rs
JonathanHenson: I pulled this from head. I'll circle back when I'm trying to get the rust part in order.

aws_crt_common_ws/aws-crt-common/src/lib.rs
```diff
@@ -0,0 +1,480 @@
...
+    }
+}
+
+impl<T> CallbackFuture<T>
+where
+    T: Clone,
```
rcoh, line 119: you shouldn't need this clone bound anymore–you might need to impl `Clone` manually, the compiler can't tell that `T` is actually inside an `Arc` but you never actually clone `T`

praetp:
Any update on this ?

xomyakus:
:(

DmitriyMusatkin:
It will still happen, just closing this PR for now :)

PR #824, opened 2021-07-09
Update datetime to support millisecond precision

yourslab:
This adds a millisecond member to `aws_date_time` so that `aws_date_time_as_millis` and `aws_date_time_as_nanos` can return millisecond precision. The "init" functions are also updated appropriately to set the millisecond member.

### DIFF
--- a/source/date_time.c
+++ b/source/date_time.c
@@ -12,6 +12,7 @@
 #include <aws/common/time.h>
 
 #include <ctype.h>
+#include <math.h>
 
 static const char *RFC822_DATE_FORMAT_STR_MINUS_Z = "%a, %d %b %Y %H:%M:%S GMT";
 static const char *RFC822_DATE_FORMAT_STR_WITH_Z = "%a, %d %b %Y %H:%M:%S %Z";
@@ -187,8 +188,9 @@ void aws_date_time_init_epoch_millis(struct aws_date_time *dt, uint64_t ms_since
 }
 
 void aws_date_time_init_epoch_secs(struct aws_date_time *dt, double sec_ms) {
-    dt->timestamp = (time_t)(sec_ms * AWS_TIMESTAMP_MILLIS);
-    dt->milliseconds = 0U;
+    double integral = 0;
+    dt->milliseconds = (uint16_t)(round(modf(sec_ms, &integral) * AWS_TIMESTAMP_MILLIS));
+    dt->timestamp = (time_t)integral;
     dt->gmt_time = s_get_time_struct(dt, false);
     dt->local_time = s_get_time_struct(dt, true);
 }
--- a/source/date_time.c
+++ b/source/date_time.c
@@ -170,21 +170,25 @@ struct tm s_get_time_struct(struct aws_date_time *dt, bool local_time) {
 }
 
 void aws_date_time_init_now(struct aws_date_time *dt) {
-    uint64_t current_time = 0;
-    aws_sys_clock_get_ticks(&current_time);
-    dt->timestamp = (time_t)aws_timestamp_convert(current_time, AWS_TIMESTAMP_NANOS, AWS_TIMESTAMP_SECS, NULL);
+    uint64_t current_time_ns = 0, current_time_ms = 0;
+    aws_sys_clock_get_ticks(&current_time_ns);
+    current_time_ms = aws_timestamp_convert(current_time_ns, AWS_TIMESTAMP_NANOS, AWS_TIMESTAMP_MILLIS, NULL);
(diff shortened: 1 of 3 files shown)


DavidOgunsAWS:
Looks good. Add a basic unit test though that inits a time with millis, and verifies reading the precision back out. Possibly a second test to verify that if datetime is initialized with seconds, that the fractional parts are zeroed out (nanos, and millis).

source/date_time.c
```diff
@@ -170,21 +171,26 @@ struct tm s_get_time_struct(struct aws_date_time *dt, bool local_time) {
 }
 
 void aws_date_time_init_now(struct aws_date_time *dt) {
-    uint64_t current_time = 0;
-    aws_sys_clock_get_ticks(&current_time);
-    dt->timestamp = (time_t)aws_timestamp_convert(current_time, AWS_TIMESTAMP_NANOS, AWS_TIMESTAMP_SECS, NULL);
+    uint64_t current_time_ns = 0, current_time_ms = 0;
+    aws_sys_clock_get_ticks(&current_time_ns);
+    current_time_ms = aws_timestamp_convert(current_time_ns, AWS_TIMESTAMP_NANOS, AWS_TIMESTAMP_MILLIS, NULL);
```
JonathanHenson, line 176: You could just call init_epoch_millis() here right?

source/date_time.c
```diff
@@ -747,11 +754,12 @@ double aws_date_time_as_epoch_secs(const struct aws_date_time *dt) {
 }
 
 uint64_t aws_date_time_as_nanos(const struct aws_date_time *dt) {
-    return (uint64_t)dt->timestamp * AWS_TIMESTAMP_NANOS;
+    return ((uint64_t)dt->timestamp * AWS_TIMESTAMP_NANOS) +
```
JonathanHenson, line 757: Why not use the time conversion functions?

source/date_time.c
```diff
@@ -170,21 +171,26 @@ struct tm s_get_time_struct(struct aws_date_time *dt, bool local_time) {
...
+    current_time_ms = aws_timestamp_convert(current_time_ns, AWS_TIMESTAMP_NANOS, AWS_TIMESTAMP_MILLIS, NULL);
+    dt->timestamp = (time_t)(current_time_ms / AWS_TIMESTAMP_MILLIS);
+    dt->milliseconds = (uint16_t)(current_time_ms % AWS_TIMESTAMP_MILLIS);
     dt->gmt_time = s_get_time_struct(dt, false);
     dt->local_time = s_get_time_struct(dt, true);
 }
 
 void aws_date_time_init_epoch_millis(struct aws_date_time *dt, uint64_t ms_since_epoch) {
     dt->timestamp = (time_t)(ms_since_epoch / AWS_TIMESTAMP_MILLIS);
+    dt->milliseconds = (uint16_t)(ms_since_epoch % AWS_TIMESTAMP_MILLIS);
```
bretambrose, line 185: Same here, should be able to collapse to:

`dt->timestamp = aws_timestamp_convert(ms_since_epoch, AWS_TIMESTAMP_MILLIS, AWS_TIMESTAMP_SECS, &dt->milliseconds);`

source/date_time.c
yourslab: Yes, I believe so

source/date_time.c
yourslab: I just assumed there was a reason why the original code didn't use them everywhere either. Maybe for performance reasons?

source/date_time.c
```diff
@@ -170,21 +171,23 @@ struct tm s_get_time_struct(struct aws_date_time *dt, bool local_time) {
...
+    uint64_t current_time_ns = 0;
+    aws_sys_clock_get_ticks(&current_time_ns);
+    aws_date_time_init_epoch_millis(
+        dt, aws_timestamp_convert(current_time_ns, AWS_TIMESTAMP_NANOS, AWS_TIMESTAMP_MILLIS, NULL));
 }
 
 void aws_date_time_init_epoch_millis(struct aws_date_time *dt, uint64_t ms_since_epoch) {
-    dt->timestamp = (time_t)(ms_since_epoch / AWS_TIMESTAMP_MILLIS);
+    dt->timestamp = (time_t)aws_timestamp_convert(
+        ms_since_epoch, AWS_TIMESTAMP_MILLIS, AWS_TIMESTAMP_SECS, (uint64_t *)&dt->milliseconds);
```
bretambrose, line 182: Oh god, I take that back, that cast is totally broken.  You'll need a temp local to hold the millisecond remainder.

bretambrose (changes requested):

bretambrose (approved):

PR #827, opened 2021-07-18
Fixes specifications for aws_hash proofs

feliperodri:

*Issue #, if available:*
N/A.

*Description of changes:*
1. Use the correct invariant for `aws_string`;
2. Fixes a bug in `hash_iter_delete` stub.

### DIFF
--- a/source/hash_table.c
+++ b/source/hash_table.c
@@ -1000,7 +1000,7 @@ bool aws_hash_callback_string_eq(const void *a, const void *b) {
     AWS_PRECONDITION(aws_string_is_valid(a));
     AWS_PRECONDITION(aws_string_is_valid(b));
     bool rval = aws_string_eq(a, b);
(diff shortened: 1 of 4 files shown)


SaswatPadhi (commented):
Regarding the `aws_c_string_is_valid` -> `aws_string_is_valid` change, the `aws_c_string_is_valid` function seems to accept `char *` and `aws_string_is_valid` seems to accept `struct aws_string *`. So how did this error go unnoticed before?

May be this is another of those weird C things that I am unaware of, but can a `struct aws_string *` be passed as a `char *`?

How do we catch this automatically in the future?

feliperodri:
> May be this is another of those weird C things that I am unaware of, but can a struct aws_string * be passed as a char *?

You can convert a struct into an array of bytes.

SaswatPadhi (approved):
Thanks! C is scary ...

The PR LGTM 🙂

PR #829, opened 2021-08-03
callback error out aws_hash_table_foreach

TingDaoK:
*Issue #, if available:*

*Description of changes:*


source/hash_table.c
```diff
@@ -222,7 +222,7 @@ int aws_hash_table_init(
     AWS_PRECONDITION(alloc != NULL);
     AWS_PRECONDITION(hash_fn != NULL);
     AWS_PRECONDITION(equals_fn != NULL);
-
+    map->allocator = alloc;
```
justinboswell, line 225: You're storing the allocator here, but surely it needs to be used whenever the table is re-sized right?

justinboswell (changes requested):

source/hash_table.c
TingDaoK: Yeah, I was wanting to expose the allocator as what we did for other structures(like aws_byte_buffer). Now I see it's awkward to expose this as we have the underlying structure... Will remove.eiifccugt

include/aws/common/hash_table.h
```diff
@@ -321,6 +322,10 @@ int aws_hash_table_remove_element(struct aws_hash_table *map, struct aws_hash_el
  *     element (if not set, iteration stops)
  * # AWS_COMMON_HASH_TABLE_ITER_DELETE   - Deletes the current value and
  *     continues iteration.  destroy_fn will NOT be invoked.
+ * # AWS_COMMON_HASH_TABLE_ITER_ERROR   - Error out and stop iteration not
```
bretambrose, line 325: This is very hard to understand grammar wise.

TingDaoK: Sorry about it... Updated the comment to be more grammar-correct.

justinboswell (approved):

PR #830, opened 2021-08-03
Directory traversal support [Draft, do not merge]

JonathanHenson:
mem_acquire no longer allows returning NULL, initial prototype of file-system traversal with posix implementation.

### DIFF
--- a/include/aws/common/file.h
+++ b/include/aws/common/file.h
@@ -4,7 +4,58 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
  */
+#include <aws/common/byte_buf.h>
 #include <aws/common/common.h>
+#include <aws/common/linked_list.h>
+#include <aws/common/ref_count.h>
+
+#ifdef _WIN32
+static const char AWS_PATH_DELIM = '\\';
+#else
+static const char AWS_PATH_DELIM = '/';
+#endif
+
+enum aws_file_type {
+    AWS_FILE_TYPE_NONE,
+    AWS_FILE_TYPE_FILE,
+    AWS_FILE_TYPE_SYM_LINK,
+    AWS_FILE_TYPE_DIRECTORY,
+};
+
+struct aws_directory_entry {
+    struct aws_allocator *allocator;
+    struct aws_string *path;
+    struct aws_string *relative_path;
+    enum aws_file_type file_type;
+    int64_t file_size;
+    struct aws_linked_list children;
+    struct aws_linked_list_node node;
+    struct aws_directory_entry *parent;
+    struct aws_ref_count ref_count;
+    void *impl;
+};
+
+/**
+ * Platform implementation of opening a file and loading its immediate children (not recursive).
+ * return AWS_OP_SUCCESS if successful.
+ */
+int aws_directory_entry_open_internal(struct aws_directory_entry *entry);
+
+/**
+ * Destroy any resources allocated in aws_directory_entry_open_internal.
+ */
+void aws_directory_entry_destroy_internal(struct aws_directory_entry *entry);
+
+/**
+ * General function for allocating and setting up a directory entry. Use for allocating and setting up the
+ * platform agnostic portions of aws_directory_entry.
+ *
+ * returns an instance of aws_directory_entry on success and NULL on failure.
+ */
+struct aws_directory_entry *aws_directory_entry_open_base(
+    struct aws_allocator *allocator,
+    struct aws_byte_cursor path,
+    struct aws_byte_cursor relative_path);
 
 AWS_EXTERN_C_BEGIN
 
@@ -16,6 +67,54 @@ AWS_EXTERN_C_BEGIN
  */
 AWS_COMMON_API
 FILE *aws_fopen(const char *file_path, const char *mode);
+
+/**
+ * Opens a directory at path. It's ref-count will be initialized to 1 on successful return.
+ * You must call aws_directory_entry_release() when finished or the resource will leak.
+ *
+ * This function allocates resources for children in the directory, but only with a depth of one. To get to the
+ * next level, you call aws_directory_entry_descend() which will repeat this process.
+ */
+AWS_COMMON_API struct aws_directory_entry *aws_directory_entry_open(
+    struct aws_allocator *allocator,
+    struct aws_byte_cursor path,
+    struct aws_byte_cursor relative_path);
+
+/**
+ * Holds a reference to the entry. Call this before seating an instance of aws_directory_entry().
+ */
+AWS_COMMON_API void aws_directory_entry_acquire(struct aws_directory_entry *entry);
+
+/**
+ * Releases reference to entry. When the last reference is dropped, the resource will be freed.
+ */
+AWS_COMMON_API void aws_directory_entry_release(struct aws_directory_entry *entry);
+
+/**
+ * Get's the next directory entry at the current directory depth level as entry. Returns NULL when there are no more
+ * sibling entries.
+ */
+AWS_COMMON_API struct aws_directory_entry *aws_directory_entry_get_next_sibling(struct aws_directory_entry *entry);
+
+/**
+ * Get's the previous directory entry at the current directory depth level as entry. Returns NULL when there are no more
+ * sibling entries.
+ */
+AWS_COMMON_API struct aws_directory_entry *aws_directory_entry_get_prev_sibling(struct aws_directory_entry *entry);
+
+/**
+ * Descends into the next depth level for entry. Returns the first child in entry. Returns NULL if entry is not a
+ * directory or it's empty. The return value's reference count is owned by entry. You do not need to acquire unless
+ * you'll be seating the return value beyond the scope of the root entry.
+ */
+AWS_COMMON_API struct aws_directory_entry *aws_directory_entry_descend(struct aws_directory_entry *entry);
+
+/**
+ * Returns the parent for the current entry unless the entry is the initial root directory from
+ * aws_directory_entry_open(). In that case it returns NULL. Also note, you do not need to acquire or release a
+ * reference to the returned value unless you plan on seating it beyond the scope of the root entry.
+ */
+AWS_COMMON_API struct aws_directory_entry *aws_directory_entry_parent(struct aws_directory_entry *entry);
 
 AWS_EXTERN_C_END
 
--- a/source/allocator.c
+++ b/source/allocator.c
@@ -119,7 +119,9 @@ void *aws_mem_acquire(struct aws_allocator *allocator, size_t size) {
 
     void *mem = allocator->mem_acquire(allocator, size);
     if (!mem) {
-        aws_raise_error(AWS_ERROR_OOM);
+        AWS_FATAL_ASSERT(
+            mem && "memory allocation failed. If you don't want to exit when this happens, handle OOM in your "
+                   "allocator implementation");
     }
     return mem;
 }
@@ -135,14 +137,16 @@ void *aws_mem_calloc(struct aws_allocator *allocator, size_t num, size_t size) {
      */
     size_t required_bytes;
     if (aws_mul_size_checked(num, size, &required_bytes)) {
-        return NULL;
+        AWS_FATAL_ASSERT("illegal size for calloc()");
     }
 
     /* If there is a defined calloc, use it */
     if (allocator->mem_calloc) {
         void *mem = allocator->mem_calloc(allocator, num, size);
         if (!mem) {
-            aws_raise_error(AWS_ERROR_OOM);
+            AWS_FATAL_ASSERT(
+                mem && "memory allocation failed. If you don't want to exit when this happens, handle OOM in your "
+                       "allocator implementation");
         }
         return mem;
     }
@@ -150,8 +154,9 @@ void *aws_mem_calloc(struct aws_allocator *allocator, size_t num, size_t size) {
     /* Otherwise, emulate calloc */
     void *mem = allocator->mem_acquire(allocator, required_bytes);
     if (!mem) {
-        aws_raise_error(AWS_ERROR_OOM);
-        return NULL;
+        AWS_FATAL_ASSERT(
+            mem && "memory allocation failed. If you don't want to exit when this happens, handle OOM in your "
+                   "allocator implementation");
     }
     memset(mem, 0, required_bytes);
     AWS_POSTCONDITION(mem != NULL);
@@ -186,8 +191,9 @@ void *aws_mem_acquire_many(struct aws_allocator *allocator, size_t count, ...) {
 
         allocation = aws_mem_acquire(allocator, total_size);
         if (!allocation) {
-            aws_raise_error(AWS_ERROR_OOM);
-            goto cleanup;
+            AWS_FATAL_ASSERT(
+                allocation && "memory allocation failed. If you don't want to exit when this happens, handle OOM in "
+                              "your allocator implementation");
         }
 
         uint8_t *current_ptr = allocation;
@@ -204,7 +210,6 @@ void *aws_mem_acquire_many(struct aws_allocator *allocator, size_t count, ...) {
         }
     }
 
-cleanup:
     va_end(args_allocs);
     return allocation;
 }
@@ -235,7 +240,9 @@ int aws_mem_realloc(struct aws_allocator *allocator, void **ptr, size_t oldsize,
     if (allocator->mem_realloc) {
         void *newptr = allocator->mem_realloc(allocator, *ptr, oldsize, newsize);
         if (!newptr) {
-            return aws_raise_error(AWS_ERROR_OOM);
+            AWS_FATAL_ASSERT(
+                newptr && "memory allocation failed. If you don't want to exit when this happens, handle OOM in your "
+                          "allocator implementation");
         }
         *ptr = newptr;
         return AWS_OP_SUCCESS;
@@ -248,7 +255,9 @@ int aws_mem_realloc(struct aws_allocator *allocator, void **ptr, size_t oldsize,
 
     void *newptr = allocator->mem_acquire(allocator, newsize);
     if (!newptr) {
-        return aws_raise_error(AWS_ERROR_OOM);
+        AWS_FATAL_ASSERT(
+            newptr && "memory allocation failed. If you don't want to exit when this happens, handle OOM in your "
+                      "allocator implementation");
     }
 
     memcpy(newptr, *ptr, oldsize);
@@ -274,10 +283,6 @@ static void *s_cf_allocator_allocate(CFIndex alloc_size, CFOptionFlags hint, voi
 
     void *mem = aws_mem_acquire(allocator, (size_t)alloc_size + sizeof(size_t));
 
-    if (!mem) {
-        return NULL;
-    }
-
     size_t allocation_size = (size_t)alloc_size + sizeof(size_t);
     memcpy(mem, &allocation_size, sizeof(size_t));
     return (void *)((uint8_t *)mem + sizeof(size_t));
@@ -301,9 +306,7 @@ static void *s_cf_allocator_reallocate(void *ptr, CFIndex new_size, CFOptionFlag
     size_t original_size = 0;
     memcpy(&original_size, original_allocation, sizeof(size_t));
 
-    if (aws_mem_realloc(allocator, &original_allocation, original_size, (size_t)new_size)) {
-        return NULL;
-    }
+    aws_mem_realloc(allocator, &original_allocation, original_size, (size_t)new_size);
 
     size_t new_allocation_size = (size_t)new_size;
     memcpy(original_allocation, &new_allocation_size, sizeof(size_t));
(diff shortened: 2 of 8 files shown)


source/posix/file.c
```diff
@@ -4,8 +4,85 @@
...
+
+int aws_directory_entry_open_internal(struct aws_directory_entry *entry) {
+    if (entry->impl) {
+        return AWS_OP_SUCCESS;
+    }
+
+    entry->impl = opendir(aws_string_c_str(entry->path));
+
+    if (!entry->impl) {
+        return aws_raise_error(AWS_ERROR_FILE_INVALID_PATH);
```
justinboswell, line 26: AWS_LOGF_ERROR before returning this to say that the path was invalid

include/aws/common/file.h
```diff
@@ -4,7 +4,58 @@
...
+#include <aws/common/ref_count.h>
+
+#ifdef _WIN32
+static const char AWS_PATH_DELIM = '\\';
+#else
+static const char AWS_PATH_DELIM = '/';
+#endif
+
+enum aws_file_type {
+    AWS_FILE_TYPE_NONE,
```
justinboswell, line 19: The way this is used, it's more like UNKNOWN, isnt it?

JonathanHenson: Well, for one, the info needs to be stored on the object. lstat is expensive and I don't wanna call it on demand, but I could do a bit mask and add helper functions.

source/allocator.c
```diff
@@ -135,23 +137,26 @@ void *aws_mem_calloc(struct aws_allocator *allocator, size_t num, size_t size) {
...
     }
 
     /* Otherwise, emulate calloc */
     void *mem = allocator->mem_acquire(allocator, required_bytes);
     if (!mem) {
-        aws_raise_error(AWS_ERROR_OOM);
-        return NULL;
+        AWS_FATAL_ASSERT(
+            mem && "memory allocation failed. If you don't want to exit when this happens, handle OOM in your "
+                   "allocator implementation");
```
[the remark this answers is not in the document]
JonathanHenson: that's untrue. We just can't handle it at this level. But this is basically what GC heaps do all the time, also overcommit etc....

include/aws/common/file.h
```diff
@@ -17,6 +68,54 @@ AWS_EXTERN_C_BEGIN
...
+ * Opens a directory at path. It's ref-count will be initialized to 1 on successful return.
+ * You must call aws_directory_entry_release() when finished or the resource will leak.
+ *
+ * This function allocates resources for children in the directory, but only with a depth of one. To get to the
+ * next level, you call aws_directory_entry_descend() which will repeat this process.
+ */
+AWS_COMMON_API struct aws_directory_entry *aws_directory_entry_open(
+    struct aws_allocator *allocator,
+    struct aws_byte_cursor path,
+    struct aws_byte_cursor relative_path);
```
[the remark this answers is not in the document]
JonathanHenson: I can define it better, but trust me, you're gonna want it.

source/file.c
```diff
@@ -0,0 +1,117 @@
...
+    struct aws_allocator *allocator,
+    struct aws_byte_cursor path,
+    struct aws_byte_cursor relative_path) {
+    struct aws_directory_entry *entry = aws_directory_entry_open_base(allocator, path, relative_path);
+
+    if (aws_directory_entry_open_internal(entry)) {
+        aws_string_destroy(entry->path);
+        aws_string_destroy(entry->relative_path);
+        aws_mem_release(entry->allocator, entry);
+        return NULL;
```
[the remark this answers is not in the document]
JonathanHenson: this is simpler.

include/aws/common/file.h
```diff
@@ -4,7 +4,58 @@
...
+    AWS_FILE_TYPE_DIRECTORY,
+};
+
+struct aws_directory_entry {
+    struct aws_allocator *allocator;
+    struct aws_string *path;
+    struct aws_string *relative_path;
+    enum aws_file_type file_type;
+    int64_t file_size;
+    struct aws_linked_list children;
```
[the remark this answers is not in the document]
JonathanHenson: I"m unclear what problem this is solving?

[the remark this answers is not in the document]
JonathanHenson: the more I think on it, you're right. This object is basically an iterator already, but as we want to expand to actual file system operations: move/create/delete/link etc.... we wind up with a tangled mess.

include/aws/common/file.h
```diff
@@ -4,19 +4,108 @@
...
+/**
+ * Returns true if the directory currently exists. Otherwise, it returns false.
+ */
+AWS_COMMON_API bool aws_directory_exists(struct aws_string *dir_path);
+/**
+ * Deletes a directory. If the directory is not empty, this will fail unless the recursive parameter is set to true.
+ * If recursive is true then the entire directory and all of its contents will be deleted. If it is set to false,
+ * the directory will be deleted only if it is empty. Returns AWS_OP_SUCCESS if the operation was successful. Otherwise,
+ * aws_last_error() will contain the error that occurred.
+ */
```
graebm, line 76: Document whether it's SUCCESS or ERROR if the directory does not exist

include/aws/common/allocator.h
```diff
@@ -52,14 +52,20 @@ void aws_wrapped_cf_allocator_destroy(CFAllocatorRef allocator);
...
 
 /**
- * Returns at least `size` of memory ready for usage or returns NULL on failure.
+ * Returns at least `size` of memory ready for usage. In versions v0.6.8 and prior, this function was allowed to return
+ * NULL. In later versions, if allocator->mem_acquire() returns NULL, this function will assert and exit. To handle
+ * conditions where OOM is not a fatal error, allocator->mem_acquire() is responsible for finding/reclaiming/running a
+ * GC etc...before returning.
  */
 AWS_COMMON_API
 void *aws_mem_acquire(struct aws_allocator *allocator, size_t size);
```
graebm, line 61: Food for thought: Do we want to let users register a global callback that fires before we terminate?

In the [Herb Sutter paper that kicked off this whole discussion](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0709r4.pdf#page=38&zoom=auto,-317,734), it's mentioned as something they might do, in the theoretical world where C++ went this route:

> We can additionally allow the program to register a global function to allow customizing what happens when 
exiting each function body with an error, intended to be used only by the owner of the whole program. This is 
useful for several purposes: 
> • To integrate with an existing system’s error handling or logging policy.

graebm: Man, I love that paper:
> Microsoft Word. Word also had been intentionally written to handle allocation failure gracefully. Re-
cently (summer 2019), the Word team conducted similar allocation fault injection test using their own 
low-memory test environment. It discovered that the code they believed was allocation failure-resilient 
actually failed pervasively when allocation failure was actually encountered; for example, 98% of execu-
tions crashed if <5MB memory was available. Examination showed that the code was doing a consist-
ently good job of checking for failures, but was consistently unable to meaningfully handle them. As an 
experiment, the team built Word to unconditionally terminate on allocation failure, gradually rolled it 
out to larger and larger subsets of users, and then made it universal in for the production release; there 
was no statistically significant change in crash rates on Windows, Mac, iOS, or Android. The net result 
was that Word now has fewer untested (and security-exploitable) error code paths, clearer code, and 
improvements to execution performance and binary size because of elimination of “dead code” logging 
and recovery logic (without loss of functionality because that code was already unused and nonworking 
in practice).

include/aws/common/file.h
```diff
@@ -4,19 +4,108 @@
...
+
+struct aws_directory_entry {
+    /**
+     * Absolute path to the entry from the current process root.
+     */
+    struct aws_byte_cursor path;
+    /**
+     * Path to the entry relative to the current working directory.
+     */
+    struct aws_byte_cursor relative_path;
```
graebm, line 36: if we're going to require users to pass `aws_string*` or `const char *` to our file APIs, we may as well make `path` and `relative_path` either `aws_string *` or `const char *` too

include/aws/common/allocator.h
JonathanHenson: we already have that hook in the allocator implementation? want one? define that allocator and use it. done

include/aws/common/file.h
JonathanHenson: don't need to though. I can guarantee they're properly constructed. This does nothing except for increase allocaton overhead for no good reason.

JonathanHenson: the api and type system will already force users to pass an aws_string to turn around and call the other apis. I don't know, I see the argument on ergonomics, but I don't think the ergonomics make practical sense here unless we go rewrite the aws_string implementation.

include/aws/common/file.h
```diff
@@ -4,19 +4,108 @@
...
+ * Returns true if the directory currently exists. Otherwise, it returns false.
+ */
+AWS_COMMON_API bool aws_directory_exists(struct aws_string *dir_path);
+/**
+ * Deletes a directory. If the directory is not empty, this will fail unless the recursive parameter is set to true.
+ * If recursive is true then the entire directory and all of its contents will be deleted. If it is set to false,
+ * the directory will be deleted only if it is empty. Returns AWS_OP_SUCCESS if the operation was successful. Otherwise,
+ * aws_last_error() will contain the error that occurred.
+ */
+AWS_COMMON_API int aws_directory_delete(struct aws_string *dir_path, struct aws_allocator *allocator, bool recursive);
```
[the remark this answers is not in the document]
JonathanHenson: sure, I did it for delete, but not traverse. If someone doesnt like how we allocate, they can do it themselves.

include/aws/common/file.h
```diff
@@ -4,19 +4,108 @@
...
 AWS_COMMON_API
 FILE *aws_fopen(const char *file_path, const char *mode);
 
+/**
+ * Creates a directory if it doesn't currently exist. If the directory already exists, it's ignored and assumed
+ * successful.
+ *
+ * Returns AWS_OP_SUCCESS on success. Otherwise, check aws_last_error().
+ */
+AWS_COMMON_API int aws_directory_create(struct aws_string *dir_path);
```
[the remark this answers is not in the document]
JonathanHenson: talked offline, determined there's no good answer here, and going to break in the direction of doing as little as possible under the hood and taking the types we need to safely operate the API.

include/aws/common/file.h
```diff
@@ -4,7 +4,58 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
  */
+#include <aws/common/byte_buf.h>
 #include <aws/common/common.h>
+#include <aws/common/linked_list.h>
+#include <aws/common/ref_count.h>
```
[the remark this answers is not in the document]
JonathanHenson: no longer relevant.

source/allocator.c
```diff
@@ -51,9 +51,13 @@ static void *s_default_malloc(struct aws_allocator *allocator, size_t size) {
...
 #if !defined(_WIN32)
     void *result = NULL;
-    return (posix_memalign(&result, alignment, size)) ? NULL : result;
+    posix_memalign(&result, alignment, size);
+    AWS_PANIC_OOM(result, "posix_memalign failed to allocate memory");
+    return result;
 #else
-    return _aligned_malloc(size, alignment);
+    void *mem = _aligned_malloc(size, alignment);
+    AWS_FATAL_POSTCONDITION(mem && "_aligned_malloc failed to allocate memory");
```
feliperodri, line 59: The explanation should be after a comma.
```c
AWS_FATAL_POSTCONDITION(mem != NULL, "_aligned_malloc failed to allocate memory");
```

@JonathanHenson

JonathanHenson: we changed it to a different macro, see latest diff.

include/aws/common/assert.h
```diff
@@ -18,6 +18,24 @@ void aws_fatal_assert(const char *cond_str, const char *file, int line) AWS_ATTR
 
 AWS_EXTERN_C_END
 
+#if defined(CBMC)
+#    define AWS_PANIC_OOM(mem, msg)                                                                                    \
+        do {                                                                                                           \
+            if (!(mem)) {                                                                                              \
+                fprintf(stderr, "%s: %s, line %d", msg, __FILE__, __LINE__);                                           \
+                exit(-1);                                                                                              \
```
feliperodri, line 26: Why `exit(-1)` and not return `AWS_OP_FAILURE` or `NULL` (depending on the caller) and let the caller handle an appropriate response in case of failure?

source/allocator.c
feliperodri: Am I looking at the wrong diff? I see a new macro at `assert.h` called `AWS_PANIC_OOM`, but I don't see any updates to `AWS_FATAL_POSTCONDITION`. So if you we don't add a comma here, we will still treat the explanation as part of the condition.

include/aws/common/assert.h
JonathanHenson: not sure I understand. The entire point is to never let NULL escape an allocation call.

JonathanHenson: there is no appropriate response to an allocation failure. Doesn't exist

source/allocator.c
```diff
@@ -69,34 +74,31 @@ static void s_default_free(struct aws_allocator *allocator, void *ptr) {
...
         return ptr;
     }
 
     /* newsize is > oldsize, need more memory */
     void *new_mem = s_default_malloc(allocator, newsize);
-    if (new_mem) {
-        memcpy(new_mem, ptr, oldsize);
-        s_default_free(allocator, ptr);
-    }
+    AWS_PANIC_OOM(new_mem, "Unhandled OOM encountered in s_default_malloc");
```
feliperodri, line 85: At this point, both `new_mem` and `ptr` should be different than NULL.

feliperodri:
```suggestion
    AWS_PANIC_OOM(new_mem && ptr, "Unhandled OOM encountered in s_default_malloc");
```

source/allocator.c
```diff
@@ -133,26 +134,20 @@ void *aws_mem_calloc(struct aws_allocator *allocator, size_t num, size_t size) {
     /* Defensive check: never use calloc with size * num that would overflow
      * https://wiki.sei.cmu.edu/confluence/display/c/MEM07-C.+Ensure+that+the+arguments+to+calloc%28%29%2C+when+multiplied%2C+do+not+wrap
      */
-    size_t required_bytes;
-    if (aws_mul_size_checked(num, size, &required_bytes)) {
-        return NULL;
-    }
+    size_t required_bytes = 0;
+    AWS_FATAL_POSTCONDITION(!aws_mul_size_checked(num, size, &required_bytes) && "calloc computed size > SIZE_MAX");
```
feliperodri, line 138:
```suggestion
    AWS_FATAL_POSTCONDITION(!aws_mul_size_checked(num, size, &required_bytes), "calloc computed size > SIZE_MAX");
```

source/allocator.c
feliperodri:
```suggestion
    AWS_FATAL_POSTCONDITION(mem, "_aligned_malloc failed to allocate memory");
```

source/allocator.c
```diff
@@ -133,26 +134,20 @@ void *aws_mem_calloc(struct aws_allocator *allocator, size_t num, size_t size) {
...
     void *mem = allocator->mem_acquire(allocator, required_bytes);
-    if (!mem) {
-        aws_raise_error(AWS_ERROR_OOM);
-        return NULL;
-    }
+    AWS_PANIC_OOM(mem, "Unhandled OOM encountered in aws_mem_acquire with allocator");
+
     memset(mem, 0, required_bytes);
     AWS_POSTCONDITION(mem != NULL);
     return mem;
```
feliperodri, line 153: Shouldn't we remove line 152 `AWS_POSTCONDITION(mem != NULL);` ?

source/allocator.c
```diff
@@ -347,9 +329,7 @@ CFAllocatorRef aws_wrapped_cf_allocator_new(struct aws_allocator *allocator) {
 
     cf_allocator = CFAllocatorCreate(NULL, &context);
 
-    if (!cf_allocator) {
-        aws_raise_error(AWS_ERROR_OOM);
-    }
+    AWS_FATAL_ASSERT(cf_allocator && "creation of cf allocator failed!");
```
feliperodri, line 332:
```suggestion
    AWS_FATAL_ASSERT(cf_allocator, "creation of cf allocator failed!");
```

source/string.c
```diff
@@ -4,6 +4,123 @@
...
+    struct aws_byte_cursor convert_cur = aws_byte_cursor_from_string(to_convert);
+    return aws_string_convert_to_wchar_from_byte_cursor(allocator, &convert_cur);
+}
+
+struct aws_string *aws_string_convert_to_wchar_from_byte_cursor(
+    struct aws_allocator *allocator,
+    const struct aws_byte_cursor *to_convert) {
+    /* if a length is passed for the to_convert string, converted size does not include the null terminator,
+     * which is a good thing. */
+    int converted_size = MultiByteToWideChar(CP_UTF8, 0, (const char *)to_convert->ptr, (int)to_convert->len, NULL, 0);
```
feliperodri, line 22: Accessing `to_convert->` without checking for nullness could lead to null-pointer dereferences.

source/string.c
```diff
@@ -4,6 +4,123 @@
...
+
+    /* double the size because the return value above is # of characters, not bytes size. */
+    if (aws_mul_size_checked(sizeof(wchar_t), converted_size, &str_len_size)) {
+        return NULL;
+    }
+
+    /* UTF-16, the NULL terminator is two bytes. */
+    if (aws_add_size_checked(sizeof(struct aws_string) + 2, str_len_size, &malloc_size)) {
+        return NULL;
+    }
```
feliperodri, line 40: Shall we create macros to return NULL?

```c
#define AWS_RETURN_NULL_IF_ERROR(cond)                                  \
    do {                                                                \
        if (cond == AWS_OP_ERR) {                                       \
            return NULL;                                                \
        }                                                                                                      \
    } while (0)
```

```c
AWS_RETURN_NULL_IF_ERROR(aws_mul_size_checked(sizeof(wchar_t), converted_size, &str_len_size));
AWS_RETURN_NULL_IF_ERROR(aws_add_size_checked(sizeof(struct aws_string) + 2, str_len_size, &malloc_size));
```

source/string.c
```diff
@@ -4,6 +4,123 @@
...
+    const struct aws_string *to_convert) {
+    return s_convert_from_wchar(
+        allocator, aws_string_wchar_c_str(to_convert), (int)aws_string_wchar_num_chars(to_convert));
+}
+struct aws_string *aws_string_convert_from_wchar_c_str(struct aws_allocator *allocator, const wchar_t *to_convert) {
+    return s_convert_from_wchar(allocator, to_convert, -1);
+}
+
+const wchar_t *aws_string_wchar_c_str(const struct aws_string *str) {
+    return (wchar_t *)str->bytes;
```
feliperodri, line 110:
```suggestion
    AWS_PRECONDITION(str);
    return (wchar_t *)str->bytes;
```

source/allocator.c
JonathanHenson: we don't really care about what ptr is at this point though right?

source/allocator.c
JonathanHenson: the assert macro isn't done that way.

source/string.c
JonathanHenson: no

source/allocator.c
feliperodri: We only need to avoid calling `memcpy` with a NULL pointer. So your current solution also works.

include/aws/common/file.h
```diff
@@ -4,18 +4,113 @@
...
 AWS_EXTERN_C_BEGIN
 
 /**
- * To support non-ascii file path across platform.
- * For windows, _wfopen will be invoked under the hood. For other platforms, same as calling fopen
- * Functionality is the same as fopen.
- * On error, errno will be set, and NULL will be returned. Same as fopen.
+ * Don't use this. It never should have been added in the first place. It's now deprecated.
+ */
+AWS_COMMON_API FILE *aws_fopen(const char *file_path, const char *mode);
```
feliperodri, line 58: Why don't just completely remove it?

source/allocator.c
feliperodri: Oh, you're right! So the string here is useless, right?

justinboswell: it will be reported as part of the failed condition. This is how we've attached context to assert failures for 40 years :)

include/aws/common/file.h
justinboswell: That would be an API break. Instead, we need to deprecate it and remove it later.

include/aws/common/string.h
```diff
@@ -50,6 +52,74 @@ struct aws_string {
...
 
+#ifdef _WIN32
+/**
+ * For windows only. Converts `to_convert` to a windows whcar format (UTF-16) for use with windows OS interop.
+ *
+ * Note: `to_convert` is assumed to be UTF-8 or ASCII.
+ *
+ * returns NULL on failure.
+ */
+AWS_COMMON_API struct aws_string *aws_string_convert_to_wchar_str(
```
graebm, line 63: Seems like a bad idea to SOMETIMES put UTF-16 in an `aws_string`.

I'd prefer we add a new type like `aws_wstring` or `aws_string16` or `aws_utf16`. We wouldn't need to have the full suite of functions we have for vanilla `aws_string`, just add enough to get by

PR #842, opened 2021-09-29
USE_CPU_EXTENSIONS: new build flag for disabling CPU-specific function implementations

graebm:
**Issue:**
We have customers encountering mysterious crashes in some of these functions. Give them the ability to disable them while we attempt to diagnose the issue.

**Description of changes:**
New build flag controls whether to use optimized functions that make use of CPU extensions (ex: AVX, SSE).


bretambrose (approved):
I would check aws-checksums and make sure it's not doing any separate feature tests for the crc assembly

graebm:
> I would check aws-checksums and make sure it's not doing any separate feature tests for the crc assembly
I did look through it, pretty sure I got it right. I'm going the same thing as I did here, where only /arch/generic/*.c code is compiled. And at runtime `aws_cpu_has_feature(X)` always returns false, since that function is implemented here in aws-c-common.

> Also, are there Macs with x86
Based on this [release chart](https://en.wikipedia.org/wiki/MacOS_version_history#Releases), 10.14 Mojave was the last version of the OS to support 32bit applications. It's also the oldest OS version that is currently supported. But looking at that chart it seems like they EOL things towards the end of each year, so I doubt it will be supported more than a handful of months longer.

>  Should there be a mac CI build?
I'll add one

Issue #844, opened 2021-10-01
Cmake module install prefix causes packaging headaches


pwillis-els:
As mentioned here https://github.com/awslabs/aws-c-event-stream/issues/15 , the current aws-c-common cmake module paths cause the packages to install and load from non-standard paths on distributions. As a result, the packagers are having to include hacks after building to move the files into the correct paths, so other packages will pick up where the files are.

I am building the AWS package stack for AWS CLI in Alpine right now, and though I'm a total cmake newb, that issue seems to confirm the problem I'm seeing now, which is that I can't force the modules to install to the correct paths, nor include them from the correct paths, without patching the CMakeLists.txt for all the AWS packages.

Maybe I'm wrong and I missed some cmake option that will fix this? In the case of Alpine, the system's modules are installed to `/usr/share/cmake/Modules`, and AWS tools want to install / load from `/usr/lib/cmake`. I can use the following arguments:

```
        cmake -B build \
                -DCMAKE_INSTALL_PREFIX=/usr \
                -DCMAKE_INSTALL_LIBDIR=share \
                "${pkgname}-${pkgver}"
```

To get the files into `/usr/share/cmake`, but I have to move them into `/usr/share/cmake/Modules` before the installed files are packaged. Otherwise you'd have to patch every single app that tried to find them to use the non-standard `/usr/share/cmake` path.

The issue linked at the top explains how to modify CMakeLists.txt to install into a package-specific directory, and also how to specify aws-specific package directories. So, unless you can find a backwards-compatible workaround, this might mean breaking your previous build instructions for your whole stack of tools. But they would install and load according to standard convention...

Anyway, I'm gonna hack around them for now like the other distros, but I'm curious what your thoughts are, or if there's a simpler fix for all this. Thanks!

PR #850, opened 2021-10-08
CMake package should add the modules to CMAKE_MODULE_PATH

r-burns:
(This is the same as https://github.com/awslabs/aws-c-common/pull/587, but rebased and reissued against the new main branch)

Right now aws-c-event-stream needs a lot of boilerplate code to determine the modules installation path while the aws-c-common package should provide it.
This would fix https://github.com/awslabs/aws-c-event-stream/issues/15 simply by moving `find_package(aws-c-common)` before the various module `include`s
Using `configure_package_config_file` makes it easier to have a relocatable package.

*Issue #, if available:*

Closes https://github.com/awslabs/aws-c-common/issues/844

*Description of changes:*


graebm:
Thanks for this great suggestion. Closing this PR to take care of it in a newer one here (along with a bunch of other little fixes): https://github.com/awslabs/aws-c-common/pull/1178

PR #862, opened 2021-11-02
Disable CPU extensions on mingw64 builds on Windows

bretambrose:
Outstanding gcc bug (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54412) can generate misaligned avx2 instructions (not 32 bit aligned) under certain circumstances.  Let's disable arch-specific code in that case until a better solution presents itself.


justinboswell (approved):

TingDaoK (approved):
Good info around

graebm (approved):

PR #868, opened 2021-11-18
Support for Windows server 2008

TingDaoK:
*Issue #, if available:*

- Functions we used not existed in windows server 2008

*Description of changes:*

- Try to load the function from kernel during runtime.
- If not exist, either not improve it or just return error


source/windows/thread.c
```diff
@@ -143,18 +147,46 @@ int aws_thread_init(struct aws_thread *thread, struct aws_allocator *allocator)
...
+typedef DWORD(GetActiveProcessorCount_fn)(WORD);
+static GetActiveProcessorCount_fn *s_GetActiveProcessorCount;
+
+typedef WORD(GetActiveProcessorGroupCount_fn)(void);
+static GetActiveProcessorGroupCount_fn *s_GetActiveProcessorGroupCount;
+
+static void s_check_active_processor_functions(void *user_data) {
+    (void)user_data;
+
+    s_GetActiveProcessorGroupCount = (GetActiveProcessorGroupCount_fn *)GetProcAddress(
```
TingDaoK, line 163: > The GetModuleHandle function returns a handle to a mapped module without incrementing its reference count.

include/aws/common/rw_lock.h
```diff
@@ -54,6 +54,7 @@ AWS_COMMON_API int aws_rw_lock_wlock(struct aws_rw_lock *lock);
  * Attempts to acquire the lock but returns immediately if it can not.
  * While on some platforms such as Windows, this may behave as a reentrant mutex,
  * you should not treat it like one. On platforms it is possible for it to be non-reentrant, it will be.
+ * Note: For windows, minimum support server version is Windows Server 2008 R2 [desktop apps | UWP apps]
```
justinboswell, line 57: Just out of curiosity, why not fallback to ```Enter/LeaveCriticalSection``` which have been around since the 90s?

TingDaoK: hmmm, good call! So, if `TryAcquireSRWLockExclusive` not exist, we could use `TryEnterCriticalSection` for try lock and `Enter/LeaveCriticalSection` for lock and unlock right?

TingDaoK: Use critical section will:
- Will increase the size of lock for modern windows (We don't take allocator for mutex)
- We have static allocator for mutex…

justinboswell (approved):

PR #869, opened 2021-11-22
Use `NULL` in place of `malloc(0)` in CBMC `byte_buf` proofs

SaswatPadhi:
*Issue #, if available:* n/a

*Description of changes:*

This PR avoids a `malloc(0)` call, which is implementation defined per C standard, and uses `NULL` in place of it.

Moreover, this change also makes the proof allocator identical to the code allocator:
https://github.com/awslabs/aws-c-common/blob/3b373b5095c543840a76be0053f2fc1f8326f030/source/byte_buf.c#L21


feliperodri (approved):
LGTM.

PR #874, opened 2021-12-06
avoid overflowing pointers passed to memcpy

qinheping:
After executing the for-loop in ```aws_array_list_mem_swap```, the pointers ```item1``` and ```item2``` could be one past the boundaries. Therefore, both pointers could be outside the address space of the program and result in undefined behavior according to the C99 as follows (7.21.1 par 2):
>Where an argument declared as size_t n specifies the length of the array for a function, n can have the value zero on a call to that function. Unless explicitly stated otherwise in the description of a particular function in this subclause, pointer arguments on such a call shall still have valid values, as described in 7.1.4..  

*Issue: https://github.com/diffblue/cbmc/issues/6472.*

*Description of changes: stop calling ```memcpy``` if the ```remainder``` is zero.*


source/array_list.c
```diff
@@ -184,9 +184,12 @@ static void aws_array_list_mem_swap(void *AWS_RESTRICT item1, void *AWS_RESTRICT
     }
 
     size_t remainder = item_size & (SLICE - 1); /* item_size % SLICE */
```
jeking3, line 186: Why not actually use `item_size % SLICE` here?  It's more understandable.  I know it's not part of the PR, just a comment.

qinheping: I believe it is because bitwise AND is faster than modulo, but I couldn't find a good reference.

graebm: +1 on just using `%` instead of a tricky trick

this operation is just happening once, it's not in some tight loop where the optimization is going to make any difference

PR #881, opened 2021-12-20
Add aws_byte_cursor_parse_uint64

cmello:
The `aws_strutil_read_unsigned_num` function is defined in the `/include/private` folder of `aws-c-http`:

https://github.com/awslabs/aws-c-http/blob/main/include/aws/http/private/strutil.h#L12-L31

We also need the same functionality in `aws-c-s3`.

This commit is the first step of the following refactor:

* Move the `aws_strutil_read_unsigned_num` function to `aws-c-common` (renamed to `aws_byte_cursor_parse_uint64`)
* Delete the function definition in `aws-c-http` and the corresponding copy `aws_s3_strutil_read_unsigned_num` in `aws-c-s3`
* fix callers of the function in `aws-c-http` and `aws-c-s3` so they use the version from `aws-c-common`

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/aws/common/byte_buf.h
+++ b/include/aws/common/byte_buf.h
@@ -901,7 +901,7 @@ AWS_COMMON_API bool aws_isspace(uint8_t ch);
  * "999999999999999999999999999999999999999999" // larger than max u64
  */
 AWS_COMMON_API
-int aws_byte_cursor_parse_uint64(struct aws_byte_cursor cursor, uint64_t *dst);
+int aws_byte_cursor_utf8_parse_u64(struct aws_byte_cursor cursor, uint64_t *dst);
 
 /**
  * Read entire cursor as ASCII/UTF-8 unsigned base-16 number with NO "0x" prefix.
@@ -920,7 +920,7 @@ int aws_byte_cursor_parse_uint64(struct aws_byte_cursor cursor, uint64_t *dst);
  * "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF" // larger than max u64
  */
 AWS_COMMON_API
-int aws_byte_cursor_parse_uint64_hex(struct aws_byte_cursor cursor, uint64_t *dst);
+int aws_byte_cursor_utf8_parse_u64_hex(struct aws_byte_cursor cursor, uint64_t *dst);
 
 AWS_EXTERN_C_END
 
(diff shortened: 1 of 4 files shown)


source/byte_buf.c
```diff
@@ -1633,3 +1633,46 @@ bool aws_isspace(uint8_t ch) {
...
+        const uint8_t c = cursor.ptr[i];
+        const uint8_t cval = hex_to_num_table[c];
+        if (cval >= base) {
+            return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        }
+
+        const uint64_t prev_val = val;
+
+        val *= base;
+        if (val < prev_val) {
```
JonathanHenson, line 1658: i think we have functions for this. something like aws_mul_overflow_check()

JonathanHenson (approved):

source/byte_buf.c
cmello: Changed to `aws_mul_u64_checked` and `aws_add_u64_checked` , thanks a lot!

include/aws/common/byte_buf.h
```diff
@@ -882,6 +882,46 @@ AWS_COMMON_API bool aws_isxdigit(uint8_t ch);
...
+ * "-1" // negative numbers not allowed
+ * "1,000" // only characters 0-9 allowed
+ * "" // blank string not allowed
+ * " 0 " // whitespace not allowed
+ * "0x0" // hex not allowed
+ * "FF" // hex not allowed
+ * "999999999999999999999999999999999999999999" // larger than max u64
+ */
+AWS_COMMON_API
+int aws_byte_cursor_parse_uint64(struct aws_byte_cursor cursor, uint64_t *dst);
```
graebm, line 904: We need the name to be clear

We don't want this to be confused with functions like like `aws_byte_cursor_read_u8()` or `aws_byte_cursor_read_u64()` that reads the bytes as binary. Have something like "str" or "utf8" in the name to be clear we're operation on a byte-cursor that contains a string

also, if you're going to have "uint64" in the name, use "u64" instead to be consistent with other functions in this file.

maybe: `aws_byte_cursor_utf8_parse_u64()`? `aws_byte_cursor_parse_str_to_u64()`?

source/byte_buf.c
graebm: yeah, use `aws_add_u64_checked()` and `aws_mul_u64_checked()` instead of my hacky overflow checks. I don't think those functions existed when I first wrote this?

graebm (approved):

include/aws/common/byte_buf.h
cmello: Renamed to `aws_byte_cursor_utf8_parse_u64` , thanks!

PR #883, opened 2021-12-23
List with map

TingDaoK:
Implement a list with map to support constant time of insert, remove and get random element from the data structure.


include/aws/common/array_list_with_map.h
```diff
@@ -0,0 +1,82 @@
...
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/common/array_list.h>
+#include <aws/common/common.h>
+#include <aws/common/hash_table.h>
+
+struct aws_array_list_with_map {
+    struct aws_array_list list;
+    struct aws_hash_table map; /* map from the element to the index in the array */
```
ilevyor, line 15: comment that the map values are pointers to array list elements

include/aws/common/array_list_with_map.h
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/common/array_list.h>
+#include <aws/common/common.h>
+#include <aws/common/hash_table.h>
+
+struct aws_array_list_with_map {
```
ilevyor, line 13: Maybe some explanation on the relationship between the map and the list? If I understand GfG:
list is a list of elements.
map is a map whose keys are elements, and value are array indexes.

include/aws/common/array_list_with_map.h
```diff
@@ -0,0 +1,82 @@
...
+ * Initialize the array_list with map structure, which support constant time of insert, remove and get random element
+ * from the data structure.
+ *
+ * The underlying hash map will use hash_fn to compute the hash of each element. equals_fn to compute equality of two
+ * keys.
+ *
+ * @param list_with_map             Pointer of structure to initialize with
+ * @param allocator                 Allocator
+ * @param hash_fn                   Compute the hash of each element
+ * @param equals_fn                 Compute equality of two keys
```
ilevyor, line 31: Nit picky, but I like how hash_table.h describes these functions better:
`equals_fn to compute equality of two keys` of each sounds like your iterating through the table.
equals_fun computes the equality of two elements not keys. the fact the the elements are keys in hashtable is an implementation detail.

include/aws/common/array_list_with_map.h
```diff
@@ -0,0 +1,82 @@
...
+ * @param destroy_element_fn        Called when the element is removed
```
ilevyor, line 32: mark as optional

include/aws/common/array_list_with_map.h
```diff
@@ -0,0 +1,82 @@
...
+ * @param item_size                 Size of each element
```
ilevyor, line 33: There's probably a good reason for this since array_list does this, but why do we make the user give us the size? Can't we just calculate?

include/aws/common/array_list_with_map.h
```diff
@@ -0,0 +1,82 @@
...
+ * @param initial_item_allocation   The initial number of item to allocate.
```
ilevyor, line 34: item_allocation sounds kind of weird can we just call this size? From the perspective of the user this is just a set with an initial size, they don't need to care that size isn't the actual size of the list. aws_hash_table size doesn't correspond to the underlying array size either: https://github.com/awslabs/aws-c-common/blob/d9031cf4f02fe4d8c24a53bb6c8df67ce8be9f0c/source/hash_table.c#L233

include/aws/common/array_list_with_map.h
```diff
@@ -0,0 +1,82 @@
...
+    aws_hash_callback_eq_fn *equals_fn,
+    aws_hash_callback_destroy_fn *destroy_element_fn,
+    size_t item_size,
+    size_t initial_item_allocation);
+
+AWS_COMMON_API
+void aws_array_list_with_map_clean_up(struct aws_array_list_with_map *list_with_map);
+
+/**
+ * Insert the element to the end of the array list. A map from the element to the index of it to the hash table.
```
ilevyor, line 51: adds an element to the set, no need to mention array list. What does this function do if an element already exists? Do you want an optional was-created parameter?

https://github.com/awslabs/aws-c-common/blob/d9031cf4f02fe4d8c24a53bb6c8df67ce8be9f0c/include/aws/common/hash_table.h#L269-L270

include/aws/common/array_list_with_map.h
```diff
@@ -0,0 +1,82 @@
...
+ */
+AWS_COMMON_API
+int aws_array_list_with_map_insert(struct aws_array_list_with_map *list_with_map, const void *element);
+
+/**
+ * Find and remove the element from the table. If the element not exist or the table is empty, nothing will happen.
+ * Switch the element with the end of the arraylist if needed.
+ * Remove the end of the arraylist
```
ilevyor, line 59: No need to mention implementation details.
Do you want an optional was_present parameter?
Does the this operation call destroy_fn if out parameter is NULL?
https://github.com/awslabs/aws-c-common/blob/d9031cf4f02fe4d8c24a53bb6c8df67ce8be9f0c/include/aws/common/hash_table.h#L281-L284

include/aws/common/array_list_with_map.h
```diff
@@ -0,0 +1,82 @@
...
+ */
+AWS_COMMON_API
+int aws_array_list_with_map_remove(struct aws_array_list_with_map *list_with_map, const void *element);
+
+/**
+ * Get a random element from the data structure. Fails when the data structure is empty.
+ */
+AWS_COMMON_API
+int aws_array_list_with_map_get_random(struct aws_array_list_with_map *list_with_map, void *out);
```
ilevyor, line 68: Do you also want a `aws_array_list_with_map_remove_random` function?

include/aws/common/array_list_with_map.h
```diff
@@ -0,0 +1,82 @@
...
+
+AWS_COMMON_API
+size_t aws_array_list_with_map_length(struct aws_array_list_with_map *list_with_map);
```
ilevyor, line 71: nitpick, a set has a size not a length.

include/aws/common/array_list_with_map.h
```diff
@@ -0,0 +1,82 @@
...
+void aws_array_list_with_map_clean_up(struct aws_array_list_with_map *list_with_map);
+
+/**
+ * Insert the element to the end of the array list. A map from the element to the index of it to the hash table.
+ */
+AWS_COMMON_API
+int aws_array_list_with_map_insert(struct aws_array_list_with_map *list_with_map, const void *element);
+
+/**
+ * Find and remove the element from the table. If the element not exist or the table is empty, nothing will happen.
```
ilevyor, line 57:
```suggestion
 * Find and remove the element from the table. If the element does not exist, or the table is empty, nothing will happen.
```

include/aws/common/hash_table.h
```diff
@@ -270,7 +270,7 @@ int aws_hash_table_create(
  * element was found, or 1 is a new element was created.
  *
  * Returns AWS_OP_SUCCESS if an item was found or created.
- * Raises AWS_ERROR_OOM if hash table expansion was required and memory
+ * Raises AWS_ERROR_OOM if hash table expansion was required and ran out of memory
```
ilevyor, line 273: Probably out of scope for this pr, but do we even have to worry about this anymore? iirc mike said something about just crashing on OOM now right?

include/aws/common/array_list_with_map.h
ilevyor, line 13: Hide the implementation details using an opaque impl/state struct similar to hash_table:
https://github.com/awslabs/aws-c-common/blob/d9031cf4f02fe4d8c24a53bb6c8df67ce8be9f0c/include/aws/common/hash_table.h#L48-L51

tests/array_list_with_map_test.c
```diff
@@ -0,0 +1,186 @@
...
+        &list_with_map, allocator, s_hash_string_ptr, s_hash_string_ptr_eq, NULL, sizeof(struct aws_string *), 1));
+    ASSERT_SUCCESS(aws_array_list_with_map_insert(&list_with_map, &foobar));
+    ASSERT_SUCCESS(aws_array_list_with_map_insert(&list_with_map, &bar));
+    ASSERT_SUCCESS(aws_array_list_with_map_insert(&list_with_map, &foo));
+
+    /* You cannot have duplicates */
+    ASSERT_FAILS(aws_array_list_with_map_insert(&list_with_map, &foobar));
+
+    /* Check the size */
+    ASSERT_UINT_EQUALS(aws_array_list_with_map_length(&list_with_map), 3);
```
ilevyor, line 52: If you add an iterator, it might be nice to also assert that order of insert, is order of iterator.

tests/array_list_with_map_test.c
```diff
@@ -0,0 +1,186 @@
...
+
+    aws_array_list_with_map_clean_up(&list_with_map);
+    return AWS_OP_SUCCESS;
+}
+
+AWS_TEST_CASE(array_list_with_map_insert_test, s_array_list_with_map_insert_fn)
+
+static int s_array_list_with_map_get_random_fn(struct aws_allocator *allocator, void *ctx) {
```
ilevyor, line 60: I'd want some fuzz testing to make sure our random function gives a roughly uniform output. Not sure how much work this is though.

source/array_list_with_map.c
```diff
@@ -0,0 +1,127 @@
...
+    AWS_PRECONDITION(list_with_map);
+    size_t length = aws_array_list_length(&list_with_map->list);
+    if (length == 0) {
+        return aws_raise_error(AWS_ERROR_LIST_EMPTY);
+    }
+
+    /* use the best of two algorithm to select the connection with the lowest load. */
+    uint64_t random_64_bit_num = 0;
+    aws_device_random_u64(&random_64_bit_num);
+    size_t index = random_64_bit_num % length;
```
DmitriyMusatkin, line 114: nit: the result of mod will be skewed and you would not get ideal random distribution. consider something like http://c-faq.com/lib/randrange.html

tests/array_list_with_map_test.c
```diff
@@ -0,0 +1,186 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/common/array_list_with_map.h>
+#include <aws/common/string.h>
+
+#include <aws/testing/aws_test_harness.h>
+
+static uint64_t s_hash_string_ptr(const void *item) {
```
DmitriyMusatkin, line 11: - check exists false -> insert -> exists true -> remove -> exists false
- check that everything works correctly after underlying mem for both structures is reallocated?

tests/array_list_with_map_test.c
```diff
@@ -0,0 +1,186 @@
...
+        &list_with_map, allocator, s_hash_string_ptr, s_hash_string_ptr_eq, NULL, sizeof(struct aws_string *), 1));
+    ASSERT_SUCCESS(aws_array_list_with_map_insert(&list_with_map, &bar));
+    ASSERT_SUCCESS(aws_array_list_with_map_insert(&list_with_map, &foobar));
+    ASSERT_SUCCESS(aws_array_list_with_map_insert(&list_with_map, &foo));
+
+    ASSERT_SUCCESS(aws_array_list_with_map_remove(&list_with_map, &foo));
+    /* Check the size */
+    ASSERT_UINT_EQUALS(aws_array_list_with_map_length(&list_with_map), 2);
+
+    /* Should be fine to remove something not existing anymore? */
```
ilevyor, line 122: Should succeed and do nothing:
https://github.com/awslabs/aws-c-common/blob/d9031cf4f02fe4d8c24a53bb6c8df67ce8be9f0c/include/aws/common/hash_table.h#L279

source/array_list_with_map.c
```diff
@@ -0,0 +1,127 @@
...
+}
+
+int aws_array_list_with_map_get_random(struct aws_array_list_with_map *list_with_map, void *out) {
+    AWS_PRECONDITION(list_with_map);
+    size_t length = aws_array_list_length(&list_with_map->list);
+    if (length == 0) {
+        return aws_raise_error(AWS_ERROR_LIST_EMPTY);
+    }
+
+    /* use the best of two algorithm to select the connection with the lowest load. */
```
DmitriyMusatkin, line 111: nit: this is callers logic, so probably irrelevant here

source/array_list_with_map.c
```diff
@@ -0,0 +1,127 @@
...
+    AWS_PRECONDITION(element);
+    size_t current_length = aws_array_list_length(&list_with_map->list);
+    if (current_length == 0) {
+        /* Nothing to remove */
+        return AWS_OP_SUCCESS;
+    }
+    /* find the index of element first */
+
+    struct aws_hash_element *find = NULL;
+    /* find and remove the elment from table */
```
DmitriyMusatkin, line 71: nit: typo

tests/array_list_with_map_test.c
```diff
@@ -0,0 +1,186 @@
+/**
```
ilevyor, line 1: some thoughts on tests:
Some more tests on larger inputs, forcing more resizing, maybe add elements in a for loop? Then delete a bunch, and add some more?
Some invariant that should be preserved: 

random_access_set size = array_list len =  hash_table len = array_list last index (ie. no gaps).

Iterating through array list and finding each element should give you back the element.

source/array_list_with_map.c
```diff
@@ -0,0 +1,127 @@
...
+    return AWS_OP_SUCCESS;
+}
+
+size_t aws_array_list_with_map_length(struct aws_array_list_with_map *list_with_map) {
+    return aws_array_list_length(&list_with_map->list);
+}
+
+bool aws_array_list_with_map_exist(struct aws_array_list_with_map *list_with_map, const void *element) {
+    struct aws_hash_element *find = NULL;
+    AWS_FATAL_ASSERT(aws_hash_table_find(&list_with_map->map, element, &find) == AWS_OP_SUCCESS);
```
DmitriyMusatkin, line 125: why is this a fatal assert?

include/aws/common/array_list_with_map.h
TingDaoK: I think two lines after that explains this a bit

source/array_list_with_map.c
```diff
@@ -0,0 +1,127 @@
...
+    if (aws_hash_table_find(&list_with_map->map, element, &find)) {
+        return AWS_OP_ERR;
+    }
+    if (!find) {
+        /* It's removed already */
+        return AWS_OP_SUCCESS;
+    }
+
+    size_t index_to_remove = (size_t)find->value;
+    if (aws_hash_table_remove(&list_with_map->map, element, NULL, NULL)) {
```
ilevyor, line 81: Use `aws_hash_table_remove_element` instead

source/array_list_with_map.c
```diff
@@ -0,0 +1,127 @@
...
+        return AWS_OP_ERR;
+    }
+    /* Nothing else can fail after here. */
+
+    if (index_to_remove != current_length - 1) {
+        /* It's not the last element, we need to swap it with the end of the list and remove the last element */
+        void *last_element = aws_mem_acquire(list_with_map->list.alloc, list_with_map->list.item_size);
```
ilevyor, line 88: use `aws_array_list_swap` instead

https://github.com/awslabs/aws-c-common/blob/d9031cf4f02fe4d8c24a53bb6c8df67ce8be9f0c/include/aws/common/array_list.h#L205-L209

include/aws/common/array_list_with_map.h
TingDaoK: no. I don't see any use case of it and you can get on random and remove it if you really need.

source/array_list_with_map.c
ilevyor, line 88: Minor preference, but I think you could skip the malloc, and just copy the memory address by doing something like this:
```
void *last_element = NULL;
aws_array_list_get_at_ptr(
                &list_with_map->list, &last_element, current_length - 1);
```

source/array_list_with_map.c
TingDaoK: it will only fail when the hash table is not valid any more. https://github.com/awslabs/aws-c-common/blob/main/source/hash_table.c#L412. So, I think a fatal assert will be better here

tests/array_list_with_map_test.c
TingDaoK: I create the structure with 1 element as initial size. So, when 3 elements inserted, both of the underlying mem has been reallocated.

tests/array_list_with_map_test.c
TingDaoK: will add iterator when we need it, now we don't.

include/aws/common/array_list_with_map.h
TingDaoK: I don't see anything weird here. We allocated items for the array list. The hash table doesn't allocate anything related to the element. It stores the pointer. Those are helper as implementation details.

include/aws/common/array_list_with_map.h
TingDaoK: Well, we do tell the implementation to the user. Like the when we initial the data structure, the hash functions are details but user needs to know about. 
I have no strong feeling about the was created, but I thought about to support duplicates in the future if we need. So, a was created will be weird in that case.

include/aws/common/array_list_with_map.h
TingDaoK: ditto

include/aws/common/array_list_with_map.h
TingDaoK: good call, or we can use new/destroy. I don't know which one will be better.

include/aws/common/array_list_with_map.h
ilevyor: Ah I see the name comes from aws_array_list_init_dynamic. I still think size makes more sense since the functionality is that of a set which is more similar to a map than a list so we should use names similar to that of the hashtable, but it's not a huge deal one way or the other.

include/aws/common/array_list_with_map.h
ilevyor: Good point. Your more familiar with usage so I would defer to you, but I feel like handing the pointer to the user is better since we usually prefer to pass pointers than structs. Again, I would defer to your judgment.

source/array_list_with_map.c
DmitriyMusatkin: I looked deeper into aws_device_random and it claims to provide random numbers usable for crypto purposes. In that case, i dont think mod vs div matters that much.
Concern with mod is that a lot of default rand primitives generate numbers that will have poor randomness in the lower bits and taking mod of those will have non-uniform distribution, while crypto safe rand should have even distribution through the bits. So both approaches will have similar distributions. 
Note: unless length is multiple of u64 max all approaches will have some "mod skew", but it will be tiny (something like a couple of elements at the start of the list having higher probability by 1 / 2^~62

ilevyor: We should do fuzz tests and know for sure. Get 1,000,000 random indices, and assert that the distributions is within 3 or 4 standard deviations of normal. Note each time we run CI we would be running this test ~20 times so we need to tolerate a decent amount of variance.

include/aws/common/hash_table.h
TingDaoK: I don't want to change the behavior of others from this pr. But, just to update the doc, which was inaccurate.

source/array_list_with_map.c
DmitriyMusatkin: We already have tests for random number gen https://github.com/awslabs/aws-c-common/blob/057746b2e094f4b7a31743d8ba5a9fd0155f69f3/tests/device_random_test.c#L34. Looks like they bucket generated numbers and check that buckets are uniform. Michael implemented those tests and he chose div approach instead of mod, might be worth checking with him if one is better choice than the other

source/array_list_with_map.c
```diff
@@ -0,0 +1,127 @@
...
+        return;
+    }
+    aws_array_list_clean_up(&list_with_map->list);
+    aws_hash_table_clean_up(&list_with_map->map);
+}
+
+int aws_array_list_with_map_insert(struct aws_array_list_with_map *list_with_map, const void *element) {
+    AWS_PRECONDITION(list_with_map);
+    AWS_PRECONDITION(element);
+    if (aws_array_list_with_map_exist(list_with_map, element)) {
```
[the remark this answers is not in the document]
TingDaoK: If it's "set something", yeah, it makes sense to "you didn't set it, but it's fine". But, if we insert something, "we didn't insert it for you, but it's fine??", it should just fail from my point of view.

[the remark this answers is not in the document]
ilevyor: If the issue is the name, maybe we should rename from insert to something else. Looking at java for inspiration: https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#add(E)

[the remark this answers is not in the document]
TingDaoK: add and insert is bascially the same thing, right? If you didn't add something for real, but the add succeed. I think it's weird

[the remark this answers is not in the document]
ilevyor: Insert implies to me an ordering, but sets don't have an order.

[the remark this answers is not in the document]
ilevyor: Add is like make an element into a set, and then union.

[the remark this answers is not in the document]
TingDaoK: sure, rename it to add makes more sense, but I believe we should still error out for duplicates instead of something like was_created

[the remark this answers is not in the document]
ilevyor: I disagree, I think trying to add the same element to a set twice is perfectly legitimate. We should only error out when something went wrong. @DmitriyMusatkin what do you think?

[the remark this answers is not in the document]
TingDaoK: add thing is not "set twice", that's not hash table. It failed to add a thing, so it's an error to me.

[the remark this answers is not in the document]
ilevyor: Java, CPP, and Dotnet all have a boolean return value for set add type operations. None error out due to failure to add.
https://docs.oracle.com/javase/8/docs/api/java/util/Set.html#add-E-
https://www.cplusplus.com/reference/unordered_set/unordered_set/insert/
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.add?view=net-6.0#system-collections-generic-hashset-1-add(-0)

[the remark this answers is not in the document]
DmitriyMusatkin: What do we expect the caller behavior to be? Do we expect it to be "oh no, everything is on fire fail hard" or "ugh, they can be equal sometimes, so let me just ignore the error". We should probably go with whatever makes user experience easier.

Set is a fairly common ds, and i think common mental model of everyone working with sets would be that it will ignore repeated insert, so leaning more towards option 2. 

On add vs insert, if im not mistaken its more of a legacy thing: c++ initially implemented set as ordered data structure backed by the tree, so it inherited the insert name from the tree. While java and other languages backed set using hash table and went with add

include/aws/common/random_access_set.h
```diff
@@ -0,0 +1,80 @@
...
+
+struct aws_random_access_set_impl;
+
+struct aws_random_access_set {
+    struct aws_random_access_set_impl *impl;
+};
+
+AWS_EXTERN_C_BEGIN
+
+/* TODO: CBMC verification. */
```
ilevyor, line 21: Do we need these to merge to main?

TingDaoK: We should probably address that before merge... But, probably better to have a separate PR. I don't want to prioritize it before the stream manager thought...

TingDaoK:
moved to https://github.com/awslabs/aws-c-http/pull/361

Issue #884, opened 2022-01-28
GCC ASan compatibility


jeremy-coulon:
I looks like ASan support in aws-c-common only works for clang.

I have faced this surprising issue when trying to use ASan with gcc: https://github.com/awslabs/aws-c-http/issues/362

The fix is quite easy. Instead of giving a blacklist as a compilation flag (which is not supported by gcc), blacklist the function directly in code.

I found the macro `AWS_SUPPRESS_ASAN` here: https://github.com/awslabs/aws-c-common/blob/main/include/aws/common/macros.h#L78

It doesn't work with gcc. The recommanded way to define this macro is:
```c
#if defined(__clang__) || defined (__GNUC__)
# define AWS_SUPPRESS_ASAN __attribute__((no_sanitize_address))
#else
# define AWS_SUPPRESS_ASAN
#endif
```
See https://github.com/google/sanitizers/wiki/AddressSanitizer#turning-off-instrumentation

Then you can apply this macro to `hashlittle2()` function.

DmitriyMusatkin:
resolved by https://github.com/awslabs/aws-c-common/commit/e86e1b6f4622ff8f09d8219d00d79bf8025deaae

PR #885, opened 2022-01-28
Asan gcc

jeremy-coulon:
*Issue #, if available:*
#884 

*Description of changes:*
GCC ASan compatibility


bretambrose:
It seems like there's more needed here.  I remembered this PR and came back to it as I'm hoping gcc based asan is easier to get working with a JNI extension than clang, where I'm having no luck.  It looks like the basic feature test still fails, as an empty -fsanitize= is an error to gcc (at least 11, which is what I'm trying with).  I will try and see if I can update the feature test safely while also incorporating this PR.

DmitriyMusatkin:
Didnt realize you already had a PR for this a year ago. Yes, im working on a similar change that will also support msvc and disabling tsan

DmitriyMusatkin:
addressed in https://github.com/awslabs/aws-c-common/commit/e86e1b6f4622ff8f09d8219d00d79bf8025deaae

PR #889, opened 2022-03-02
Add cJSON module

TwistedTwigleg:
*Description of changes:*

Adds a JSON module to the common repository so it's possible for all CRT repositories to use JSON without having to use cJSON directly or use any cJSON types directly. This makes it easier to replace cJSON should we need to down the road, as well as reduce duplicate copies of cJSON from being embedded throughout the SDK.

This PR is needed for the following:
* aws-c-iot changes: https://github.com/awslabs/aws-c-iot/pull/66
* aws-c-auth changes: https://github.com/awslabs/aws-c-auth/pull/155

____________

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.*


include/aws/common/external/cJSON.h
```diff
@@ -0,0 +1,301 @@
...
+
+#define CJSON_CDECL __cdecl
+#define CJSON_STDCALL __stdcall
+
+/* export symbols by default, this is necessary for copy pasting the C and header file */
+#if !defined(CJSON_HIDE_SYMBOLS) && !defined(CJSON_IMPORT_SYMBOLS) && !defined(CJSON_EXPORT_SYMBOLS)
+#define CJSON_EXPORT_SYMBOLS
+#endif
+
+#if defined(CJSON_HIDE_SYMBOLS)
```
bretambrose, line 75: It's critical that aws-c-common be configured so that CJson symbols are private.

include/aws/common/external/cJSON.h
```diff
@@ -0,0 +1,301 @@
...
+#define CJSON_PUBLIC(type)   __attribute__((visibility("default"))) type
+#else
+#define CJSON_PUBLIC(type) type
+#endif
+#endif
+
+/* project version */
+#define CJSON_VERSION_MAJOR 1
+#define CJSON_VERSION_MINOR 7
+#define CJSON_VERSION_PATCH 12
```
bretambrose, line 96: Can we use the latest version?  If this came from aws-c-auth, you'll want to port over the safety changes that were applied to it in order to resolve appsec code scanning issues.

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+/**
+ * Creates a new string aws_json_node with the given string and returns a pointer to it.
+ *
+ * Note: You will need to free the memory for the aws_json_node using aws_json_delete on the aws_json_node or
+ * on the object/array containing the aws_json_node.
+ * @param string The string you want to store in the aws_json_node
+ * @return A new string aws_json_node
+ */
+AWS_COMMON_API
+struct aws_json_node *aws_json_string_new(char *string);
```
bretambrose, line 34: This may be a bit painful (especially for the input parameters), but there shouldn't be any C-strings in the public API, either as parameters or return values.  Use byte cursors instead and for parameters that are passed directly to CJSON, you'll need to create an aws_string, pass in the bytes, and then destroy the string.

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+char *aws_json_string_get(struct aws_json_node *node);
+
+/**
+ * Gets the number of a number aws_json_node.
+ * @param node The number aws_json_node.
+ * @return The number of the number aws_json_node, otherwise NULL.
+ */
+AWS_COMMON_API
+double *aws_json_number_get(struct aws_json_node *node);
+
```
bretambrose, line 136: I feel like we're missing an API that returns the type of a node.

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+ * Checks if there is a aws_json_node at the given key.
+ * @param object The object aws_json_node you want to check a key in.
+ * @param key The key that you want to check. Is case sensitive.
+ * @return AWS_OP_SUCCESS if a aws_json_node is found, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_object_has(struct aws_json_node *object, char *key);
+
+/**
+ * Removes the aws_json_node at the given key.
```
bretambrose, line 187: It's not clear if either of these clean up the memory of the removed object.

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+int aws_json_array_add(struct aws_json_node *array, struct aws_json_node *node);
+
+/**
+ * Returns the aws_json_node at the given index in the array aws_json_node.
+ * @param array The array aws_json_node.
+ * @param index The index of the aws_json_node you want to access.
+ * @return A pointer to the aws_json_node at the given index in the array, otherwise NULL.
+ */
+AWS_COMMON_API
+struct aws_json_node *aws_json_array_get(struct aws_json_node *array, int index);
```
bretambrose, line 224: use size_t for indexing and size values and sanity check the case internally

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+ * Returns whether the given aws_json_node is located within the array aws_json_node.
+ * @param array The array aws_json_node.
+ * @param node The aws_json_node to check for.
+ * @return AWS_OP_SUCCESS if the aws_json_node exists in the array, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_array_has(struct aws_json_node *array, struct aws_json_node *node);
+
+/**
+ * Removes the aws_json_node at the given index in the array aws_json_node.
```
bretambrose, line 244: Like objects, not clear what the memory contract is for the removed object.

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+// ====================
+// Utility
+
+/**
+ * Returns an unformatted JSON string representation of the aws_json_node.
+ * @param node The aws_json_node to format.
+ * @return A string containing the JSON.
+ */
+AWS_COMMON_API
+char *aws_json_to_string(struct aws_json_node *node);
```
bretambrose, line 358: I would have all of the _to_string* write to byte buffers instead.

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+
+/**
+ * Fills the passed in string with a JSON string representation of the aws_json_node.
+ * @param node The aws_json_node to format.
+ * @param output The string to place the JSON into.
+ * @param length The length of the aws_json_node string.
+ * @param format A boolean to determine whether to format the result.
+ */
+AWS_COMMON_API
+void aws_json_to_string_preallocated(struct aws_json_node *node, char *output, int length, bool format);
```
bretambrose, line 376: I feel like this could be removed.  If it's needed, add it later.

source/json.c
```diff
@@ -0,0 +1,358 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include "aws/common/json.h"
+
+#include "aws/common/external/cJSON.h"
+
+struct aws_json_node *aws_json_node_new(void) {
+    return (void *)(uintptr_t)cJSON_CreateObject();
```
bretambrose, line 11: I'm not sure why the uintptr_t case is necessary here.

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+AWS_COMMON_API
+int aws_json_boolean_set(struct aws_json_node *node, bool boolean);
+
+/**
+ * Gets the string of a string aws_json_node.
+ * @param node The string aws_json_node.
+ * @return The string of the string aws_json_node, otherwise NULL.
+ */
+AWS_COMMON_API
+char *aws_json_string_get(struct aws_json_node *node);
```
bretambrose, line 127: Let's do a pass through the whole API for const.

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+int aws_json_array_get_count(struct aws_json_node *array);
+
+/**
+ * Returns whether the given aws_json_node is located within the array aws_json_node.
+ * @param array The array aws_json_node.
+ * @param node The aws_json_node to check for.
+ * @return AWS_OP_SUCCESS if the aws_json_node exists in the array, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_array_has(struct aws_json_node *array, struct aws_json_node *node);
```
bretambrose, line 241: Seems unnecessary for a basic API.

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+int aws_json_object_remove(struct aws_json_node *object, char *key);
+
+/**
+ * Removes the aws_json_node from the aws_json_object using a pointer to the aws_json_node.
+ * @param object The object aws_json_node you want to remove a aws_json_node in.
+ * @param node The aws_json_node you want to remove from the object aws_json_node.
+ * @return AWS_OP_SUCCESS if the aws_json_node was removed, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_object_remove_node(struct aws_json_node *object, struct aws_json_node *node);
```
bretambrose, line 202: Seems unnecessary for a basic API.

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+int aws_json_array_remove(struct aws_json_node *array, int index);
+
+/**
+ * Removes the aws_json_node in the array aws_json_node using a pointer to the aws_json_node.
+ * @param array The array aws_json_node.
+ * @param node The aws_json_node you want to remove from the array.
+ * @return AWS_OP_SUCCESS if the aws_json_node was removed from the array, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_array_remove_node(struct aws_json_node *array, struct aws_json_node *node);
```
bretambrose, line 259: Seems unnecessary for a basic API

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+AWS_COMMON_API
+struct aws_json_node *aws_json_from_string(char *string);
+
+/**
+ * Determines if the aws_json_node is a valid aws_json_node.
+ * @param node The aws_json_node to check.
+ * @return AWS_OP_SUCCESS if the aws_json_node is valid, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_is_valid(struct aws_json_node *node);
```
bretambrose, line 392: What would make something invalid?

include/aws/common/json.h
```diff
@@ -0,0 +1,355 @@
...
+ * @return True if the boolean was successfully set, otherwise false.
+ */
+bool aws_json_set_boolean(struct aws_json_node *node, bool boolean);
+
+/**
+ * Gets the string of a string aws_json_node.
+ * @param node The string aws_json_node.
+ * @return The string of the string aws_json_node, otherwise NULL.
+ */
+char *aws_json_node_get_string(struct aws_json_node *node);
```
graebm: [trivial] naming inconsistency between setters and getters. the setters have `_node_` in them but the getters don't.

the getter here is named better than the setter, since our naming conventions call for "classname_functionname" pattern

but we have something that's kiiiiind of like subclasssing, where boolean is a "subclass" of node so I could also buy into a name like:
`aws_json_boolean_get()` `aws_json_boolean_set()`

either way

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+
+/**
+ * Creates a new aws_json_node and returns a pointer to it.
+ *
+ * Note: You will need to free the memory for the aws_json_node using aws_json_delete on the aws_json_node or
+ * on the object/array containing the aws_json_node.
+ * @return A new aws_json_node
+ */
+AWS_COMMON_API
+struct aws_json_node *aws_json_node_new(void);
```
graebm, line 23: [trivial] the offical json terminology is ["value"](https://www.json.org/json-en.html). we might want to use that instead of "node"

include/aws/common/json.h
```diff
@@ -0,0 +1,395 @@
...
+// Value setters and getters
+
+/**
+ * Sets the string value of a string aws_json_node.
+ * @param node The string aws_json_node.
+ * @param string The new string value.
+ * @return AWS_OP_SUCCESS if the string was successfully set, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_string_set(struct aws_json_node *node, char *string);
```
graebm, line 101: we might not even want to offer setters for basic values, just set the value initially in the new() function

this just needs to be the most basic possible API

include/aws/common/json.h
graebm: arguable, in the spirit of a minimal API we could just call the getter for the type we expect and bounce when an error-code is returned

include/aws/common/external/cJSON.h
TwistedTwigleg: Ah okay! I missed that but will add it in the next commit.
Thanks!

include/aws/common/external/cJSON.h
graebm: It looks like you pulled the `main` branch. There have been a few commits since `v1.7.15` was released: https://github.com/DaveGamble/cJSON/releases/tag/v1.7.15

Probably better to pull it at the actual `v1.7.15` tag, so we know what we have.

TwistedTwigleg: Will do! 👍

include/aws/common/json.h
```diff
@@ -0,0 +1,358 @@
...
+AWS_COMMON_API
+struct aws_json_value *aws_json_array_get(const struct aws_json_value *array, const size_t index);
+
+/**
+ * Returns the number of items in the array aws_json_value.
+ * @param array The array aws_json_value.
+ * @return The number of items in the array_json_node.
+ */
+AWS_COMMON_API
+int aws_json_array_get_count(const struct aws_json_value *array);
```
bretambrose, line 216: size_t

include/aws/common/json.h
```diff
@@ -0,0 +1,358 @@
...
+ * Adds a aws_json_value to the given array aws_json_value.
+ *
+ * Note that the aws_json_value will be destroyed when the aws_json_value array is destroyed
+ * by calling "aws_json_destroy()"
+ * @param array The array aws_json_value you want to add an aws_json_value to.
+ * @param node The aws_json_value you want to add.
+ * @return AWS_OP_SUCCESS if adding the aws_json_value was successful, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_array_add(const struct aws_json_value *array, const struct aws_json_value *node);
```
bretambrose, line 199: similar function name and const comments apply to many of the array APIs.

include/aws/common/json.h
```diff
@@ -0,0 +1,358 @@
...
+ * attached aws_json_values as well.
+ *
+ * For example, if you called "aws_json_array_add(b, a)" to add an object "a" to an array "b", if you call
+ * "aws_json_destroy(b)" then it will also free "a" automatically. All children/attached aws_json_values are freed
+ * when the parent/root aws_json_value is destroyed.
+ * @param node The aws_json_value to destroy.
+ * @return AWS_OP_SUCCESS if the destroy was successful, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_destroy(const struct aws_json_value *node);
```
bretambrose, line 307: function name and const removal

include/aws/common/json.h
```diff
@@ -0,0 +1,358 @@
...
+
+// ====================
+// Memory Management
+
+/**
+ * Initializes the JSON module for use.
+ * @param allocator The allocator to use for creating aws_json_value structs.
+ */
+AWS_COMMON_API
+void aws_json_module_init(const struct aws_allocator *allocator);
```
bretambrose, line 288: const removal

include/aws/common/json.h
```diff
@@ -0,0 +1,358 @@
...
+    const struct aws_byte_cursor *cursor,
+    const struct aws_allocator *allocator);
+
+/**
+ * Determines if the aws_json_value is a valid aws_json_value.
+ * @param node The aws_json_value to check.
+ * @return AWS_OP_SUCCESS if the aws_json_value is valid, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_is_valid(const struct aws_json_value *node);
```
bretambrose, line 355: bool, aws_json_value_is_valid()

include/aws/common/json.h
```diff
@@ -0,0 +1,358 @@
...
+int aws_json_to_string_formatted(const struct aws_json_value *node, struct aws_byte_buf *output,
+                                 struct aws_allocator *allocator);
+
+/**
+ * Parses the JSON string and returns a aws_json_value containing the root of the JSON.
+ * @param string The string containing the JSON.
+ * @return The root aws_json_value of the JSON.
+ */
+AWS_COMMON_API
+struct aws_json_value *aws_json_from_string(
```
bretambrose, line 345: remove const from allocator, rename as aws_json_value_new_from_string(), allocator as first param

include/aws/common/json.h
```diff
@@ -0,0 +1,325 @@
...
+
+/**
+ * Checks if there is a aws_json_value at the given key.
+ * @param object The object aws_json_value you want to check a key in.
+ * @param key The key that you want to check. Is case sensitive.
+ * @return AWS_OP_SUCCESS if a aws_json_value is found, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_object_has(const struct aws_json_value *object, const struct aws_byte_cursor *cursor,
+                        const struct aws_allocator *allocator);
```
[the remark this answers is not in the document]
graebm: since we don't actually create an aws_json_value struct that stores its allocator, we could have these getters use whatever allocator was passed to initialize cjson

include/aws/common/json.h
```diff
@@ -0,0 +1,358 @@
...
+ * Returns an unformatted JSON string representation of the aws_json_value.
+ *
+ * Note: When you are finished with the aws_byte_buf, you must call "aws_byte_buf_clean_up_secure" to free
+ * the memory used, as it will NOT be called automatically.
+ * @param node The aws_json_value to format.
+ * @param output The destination for the JSON string
+ * @return A string containing the JSON.
+ */
+AWS_COMMON_API
+int aws_json_to_string(const struct aws_json_value *node, struct aws_byte_buf *output,
```
bretambrose, line 323: These shouldn't take an allocator.  Use the byte_buf APIs to either write or append the value into the buffer.

include/aws/common/json.h
```diff
@@ -0,0 +1,325 @@
...
+/**
+ * Returns the aws_json_value at the given case insensitive key.
+ * @param object The object aws_json_value you want to get the node from.
+ * @param key The key that the aws_json_value is at. Is not case sensitive.
+ * @return The aws_json_value at the given key, otherwise NULL.
+ */
+AWS_COMMON_API
+struct aws_json_value *aws_json_object_get_insensitive(const struct aws_json_value *object,
+                                                       const struct aws_byte_cursor *cursor,
+                                                       const struct aws_allocator *allocator);
```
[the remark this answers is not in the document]
TwistedTwigleg: Works for me. 👍 
I think it's only used in a couple places and it should be easy to replicate the behavior via a helper function if needed.

include/aws/common/json.h
```diff
@@ -0,0 +1,325 @@
...
+AWS_COMMON_API
+struct aws_json_value *aws_json_from_string(const struct aws_byte_cursor *cursor, const struct aws_allocator *allocator);
+
+/**
+ * Determines if the aws_json_value is a valid aws_json_value.
+ * @param node The aws_json_value to check.
+ * @return AWS_OP_SUCCESS if the aws_json_value is valid, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_is_valid(const struct aws_json_value *node);
```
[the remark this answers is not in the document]
TwistedTwigleg: I don't think it is possible anymore. The only case where I think it is possible now is if you try to pass a `void *` that was not created using the JSON API.

include/aws/common/json.h
```diff
@@ -0,0 +1,325 @@
...
+int aws_json_array_get_count(const struct aws_json_value *array);
+
+/**
+ * Removes the aws_json_value at the given index in the array aws_json_value.
+ * @param array The array aws_json_value.
+ * @param index The index containing the aws_json_value you want to remove.
+ * @return AWS_OP_SUCCESS if the aws_json_value at the index was removed, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_array_remove(const struct aws_json_value *array, const size_t index);
```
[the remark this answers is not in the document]
TwistedTwigleg: I'm not sure. The case I can see for having it is for data manipulation, like if you need to update a value in the array. Outside of that though, I do not think it would be needed.

include/aws/common/json.h
```diff
@@ -0,0 +1,325 @@
...
+                        const struct aws_allocator *allocator);
+
+/**
+ * Removes the aws_json_value at the given key.
+ * @param object The object aws_json_value you want to remove a aws_json_value in.
+ * @param key The key that the aws_json_value is at. Is case sensitive.
+ * @return AWS_OP_SUCCESS if the aws_json_value was removed, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_object_remove(const struct aws_json_value *object, const struct aws_byte_cursor *cursor,
```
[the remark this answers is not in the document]
TwistedTwigleg: I'm not sure. The case I can see for having it is for data manipulation, like if you need to update a value in the object. Outside of that though, I do not think it would be needed.

include/aws/common/json.h
```diff
@@ -0,0 +1,358 @@
...
+// ====================
+// Value getters
+
+/**
+ * Gets the string of a string aws_json_value.
+ * @param node The string aws_json_value.
+ * @return The string of the string aws_json_value, otherwise NULL.
+ */
+AWS_COMMON_API
+int aws_json_value_get_string(const struct aws_json_value *node, struct aws_byte_cursor *output);
```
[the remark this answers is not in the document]
TwistedTwigleg: Maybe just `aws_json_get_string`?

include/aws/common/json.h
```diff
@@ -0,0 +1,358 @@
...
+    const struct aws_allocator *allocator);
+
+/**
+ * Removes the aws_json_value at the given key.
+ * @param object The object aws_json_value you want to remove a aws_json_value in.
+ * @param key The key that the aws_json_value is at. Is case sensitive.
+ * @return AWS_OP_SUCCESS if the aws_json_value was removed, otherwise AWS_OP_ERR.
+ */
+AWS_COMMON_API
+int aws_json_object_remove(
```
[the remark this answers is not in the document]
TwistedTwigleg: I'll change the return to something else 👍

TwistedTwigleg:
Thanks guys! I'll look at making the code review changes soon

include/aws/common/json.h
```diff
@@ -0,0 +1,325 @@
...
+
+// ====================
+// Memory Management
+
+/**
+ * Initializes the JSON module for use.
+ * @param allocator The allocator to use for creating aws_json_value structs.
+ */
+AWS_COMMON_API
+void aws_json_module_init(const struct aws_allocator *allocator);
```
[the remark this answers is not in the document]
TwistedTwigleg: I was trying to achieve this and I do not think it is currently possible, at least not in any way I was able to get working.
I can move some of the initialization code to aws_common_library_init, but I'd need to import `cJSON.h` into `common.c` to make a cJSON_Hooks object, which is required to initialize cJSON. Additionally, I'm not sure how to make it where I can then reference the allocator, which is needed for the getters now that an allocator is not being passed.

I don't mind changing it, but as it is currently, I do not see how to make the modification without exposing cJSON. I did remove several instances where it was called though, so the only places it is called is in `common.c` and the JSON test file.

TwistedTwigleg:
Thanks for the reviews! I made the changes, fixed the clang-tidy issues, and got it compiling on all Windows platforms successfully. 👍

include/aws/common/json.h
```diff
@@ -0,0 +1,342 @@
...
+/**
+ * Adds a aws_json_value to a object aws_json_value.
+ *
+ * Note that the aws_json_value will be destroyed when the aws_json_value object is destroyed
+ * by calling "aws_json_destroy()"
+ * @param object The object aws_json_value you want to add a value to.
+ * @param key The key to add the aws_json_value at.
+ * @param value The aws_json_value you want to add.
+ * @return AWS_OP_SUCCESS if adding was successful.
+ *          Will return AWS_ERROR_INVALID_ARGUMENT if the object passed is invalid or
```
bretambrose, line 132: error codes should not be returned.  This is either an incorrect description or the implementation needs to change.

include/aws/common/json.h
```diff
@@ -0,0 +1,342 @@
...
+    const struct aws_byte_cursor cursor);
+
+/**
+ * Checks if there is a aws_json_value at the given key.
+ * @param object The object aws_json_value you want to check a key in.
+ * @param key The key that you want to check. Is case sensitive.
+ * @return True if a aws_json_value is found.
+ */
+AWS_COMMON_API
+bool aws_json_value_check_has_in_object(const struct aws_json_value *object, const struct aws_byte_cursor cursor);
```
bretambrose, line 159: "object_has_key" seems a better suffix

source/json.c
```diff
@@ -0,0 +1,306 @@
...
+struct aws_json_value *aws_json_new_string(struct aws_allocator *allocator, const struct aws_byte_cursor cursor) {
+    struct aws_string *tmp = aws_string_new_from_cursor((struct aws_allocator *)allocator, &cursor);
+    void *ret_val = cJSON_CreateString(aws_string_c_str(tmp));
+    aws_string_destroy_secure(tmp);
+    return ret_val;
+}
+
+struct aws_json_value *aws_json_new_number(struct aws_allocator *allocator, double number) {
+    (void)(allocator); // prevent warnings over unused parameter
+    return (void *)(uintptr_t)cJSON_CreateNumber(number);
```
bretambrose, line 25: I'm still confused why uintptr_t is used here.

source/json.c
```diff
@@ -0,0 +1,306 @@
...
+}
+
+struct aws_json_value *aws_json_new_object(struct aws_allocator *allocator) {
+    (void)(allocator); // prevent warnings over unused parameter
+    return (void *)(uintptr_t)cJSON_CreateObject();
+}
+
+int aws_json_value_get_string(const struct aws_json_value *value, struct aws_byte_cursor *output) {
+    struct cJSON *cjson = (struct cJSON *)value;
+    if (cJSON_IsString(cjson)) {
```
bretambrose, line 50: If you have a function that returns AWS_OP_ERR/AWS_OP_SUCCESS, and you're returning AWS_OP_ERR, then either you need to aws_raise_error() or be certain that the right error has been raised by what you've called and are reacting to.  Since CJSON does not use aws_raise_error() you'll need to do that here.  I would switch things around like this:

```
if (!cJSON_IsString(cjson)) {
   return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
}

*output = aws_byte_cursor(...);
return AWS_OP_SUCCESS;
```

Similar for all the other functions with this pattern: raise error if it hasn't been done and prefer early out on errors until you're in the success state.

source/json.c
```diff
@@ -0,0 +1,306 @@
...
+    }
+    return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+}
+
+bool aws_json_value_is_string(const struct aws_json_value *value) {
+    struct cJSON *cjson = (struct cJSON *)value;
+    if (cjson == NULL) {
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+    if (cJSON_IsString(cjson) == true) {
```
bretambrose, line 183: return cJSON_IsString(...);

source/json.c
TwistedTwigleg: It is to prevent a compiler warning on Windows that otherwise makes the CI fail, if I recall correctly. I'll see if I can remove it since the code has changed a bit since it was added. 👍

TwistedTwigleg:
Thanks for the review! I adjusted the code accordingly 👍

source/json.c
```diff
@@ -0,0 +1,359 @@
...
+        return aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+    }
+
+    char *tmp = cJSON_Print(cjson);
+    if (tmp == NULL) {
+        return AWS_OP_ERR;
+    }
+
+    // clean the buffer before re-assigning
+    if (aws_byte_buf_is_valid(output)) {
```
bretambrose, line 345: Same here

bretambrose (commented):
Mostly some minor function naming tweaks (aws_json_value_... vs aws_json_...) and some implementation tweaks.

include/aws/common/json.h
```diff
@@ -24,7 +24,7 @@ struct aws_json_value;
  * @return A new string aws_json_value
  */
 AWS_COMMON_API
-struct aws_json_value *aws_json_new_string(struct aws_allocator *allocator, struct aws_byte_cursor string);
+struct aws_json_value *aws_json_string_new(struct aws_allocator *allocator, struct aws_byte_cursor string);
```
bretambrose, line 27: My suggstion was intended to be `aws_json_value_new_*` ie `aws_json_value_new_string` `aws_json_value_new_number` etc...

source/json.c
```diff
@@ -277,11 +269,11 @@ bool aws_json_value_is_object(const struct aws_json_value *value) {
     return cJSON_IsObject(cjson);
 }
 
-void *aws_cJSON_alloc(size_t sz) {
+static void *aws_cJSON_alloc(size_t sz) {
```
bretambrose, line 272: s_ prefix to function names here

source/json.c
```diff
@@ -321,9 +313,7 @@ int aws_json_value_to_string(const struct aws_json_value *value, struct aws_byte
     }
 
     // clean the buffer before re-assigning
-    if (aws_byte_buf_is_valid(output)) {
-        aws_byte_buf_clean_up(output);
-    }
+    aws_byte_buf_clean_up(output);
```
bretambrose, line 316: Ok, looking at this now, I've changed my mind on the byte_buf expectations.  The contract should be "pass in a zeroed or uninitialized byte buf only" and we should just overwrite output and not call clean up at all.  It makes no sense to ask people to init a buffer only to immediately destroy its contents and then overwrite them.

Actually I think I have a stronger preference:

```aws_byte_buf_append_json_string(struct aws_byte_buf *buffer, const struct aws_json_value *value)```

and

```aws_byte_buf_append_json_string_formatted(struct aws_byte_buf *buffer, const struct aws_json_value *value)```

which take an already-initialized byte buf and then append the json string to it.

We don't need to do any clean up and it can be used as a sub-part in something larger because we're not stomping the buffer.

For an example of something similar, where we have API functions targeting an "associated" type in the header definition of a related type see https://github.com/awslabs/aws-c-common/blob/main/include/aws/common/string.h#L315 where we have an aws_byte_buf API in string.h

bretambrose (approved):

TwistedTwigleg:
Thanks for the reviews! Merging into `main`...

PR #891, opened 2022-03-12
logging: fix race condition in changing the logger

grrtrr:
This race condition was found via clang TSAN analyzer.

`aws_logger_set/get` operate on a global variable, but the operation on the shared global variable is not thread-safe.

Use a `mutex` to avoid the race condition.


source/logging.c
```diff
@@ -270,17 +270,26 @@ int aws_logger_init_from_external(
...
 
 struct aws_logger *aws_logger_get(void) {
-    return s_root_logger_ptr;
+    struct aws_logger *result;
+
+    aws_mutex_lock(&s_root_logger_lock);
+    result = s_root_logger_ptr;
+    aws_mutex_unlock(&s_root_logger_lock);
+
+    return result;
```
hsaliak, line 292: This returns a pointer to the aws_logger(). The pointer can still access the underlying data in such a way that introduces a data race.
```
struct aws_logger * logger = aws_logger_get();
logger->foo = xx;
```
Since this can still race, AWS needs to provide a better solution here, whereby the API is changed to return a copy.

waahm7:
Hi, 

Thank you for the contribution. This API is not thread safe and should only be called once. Please feel free to reopen the PR if there is anything else.

https://github.com/awslabs/aws-c-common/blob/4b4675c0670a0a55ddfd37cb9eda3712eba31fc4/include/aws/common/logging.h#L222

grrtrr:
> This API is not thread safe and must only be called once. Please feel free to reopen the PR if there is anything else.

This is a very old PR. The condition occurred when `Aws::InitAPI` and  `Aws::ShutdownAPI` of `aws-sdk-cpp` were called from different threads. We ran into a similar problem (memory leak) with that in https://github.com/aws/s2n-tls/issues/3525. 

To avoid that problem, we have been using a dedicated initializer thread, which takes care of calling `Aws::InitAPI` and  `Aws::ShutdownAPI` (details are in https://github.com/aws/s2n-tls/pull/3530#issuecomment-1368249088). It may be that this also fixed the race condition we experienced when using `aws-c-common` via the `aws-sdk-cpp`.

Should the problem re-occur, I will create an issue.

PR #892, opened 2022-03-15
file utils: fix TODO in obtaining the user's home directory

grrtrr:
Currently `aws_get_home_directory` only works if `$HOME` is set.
This fails in environments that do not set `HOME`.

One such case is `bazel` tests, which break if this is not supported.
Use `getpwuid`, as employed already by the main AWS SDK.


source/posix/file.c
```diff
@@ -226,14 +227,20 @@ char aws_get_platform_directory_separator(void) {
...
-    /* ToDo: check getpwuid_r if environment check fails */
     struct aws_string *home_env_var_value = NULL;
-    if (aws_get_environment_value(allocator, s_home_env_var, &home_env_var_value) == 0 && home_env_var_value != NULL) {
-        return home_env_var_value;
-    }
 
-    return NULL;
+    if (aws_get_environment_value(allocator, s_home_env_var, &home_env_var_value) != 0 ||
+        home_env_var_value == NULL || *home_env_var_value->bytes == '\0') {
+        char buf[16384];
```
waahm7, line 234: We should not allocate a large buffer on stack. We can either use [sdk-cpp's approach](https://github.com/aws/aws-sdk-cpp/blob/bb1fdce01cc7e8ae2fe7162f24c8836e9d3ab0a2/aws-cpp-sdk-core/source/platform/linux-shared/FileSystem.cpp#L172) and allocate 4096, or something like [dotnet runtime](https://github.com/dotnet/runtime/issues/24966) where we allocate a really small buffer (256) on stack to cater 99% of the cases, and fallback on a heap allocated buffer.

@graebm What do you think?

waahm7 (changes requested):

source/posix/file.c
grrtrr: Fair enough. According to the [man page](https://linux.die.net/man/3/getpwuid_r), the correct way would be:
1. Call `sysconf(_SC_GETPW_R_SIZE_MAX)` to determine the [initial size](https://man7.org/linux/man-pages/man3/sysconf.3p.html) of the buffer, using a default in case the call fails.
2. Allocate a buffer of the size determined in (1).
3. Call `getpwuid_r` and use values if successful.
4. If `getpwuid_r` fails with `ERANGE`, de-allocate buffer allocated in (2), re-allocate a larger one, and continue with (3).

This leads to [complicated code](https://stackoverflow.com/questions/47462890/how-to-properly-set-buffer-and-bufsize-with-getpwuid-r). The suggestion to use 4k is reasonable and also agrees with `aws-sdk-cpp`. I will use that.

waahm7 (approved):

graebm:
Thanks for contributing this!

Sorry if it's rude to put out another PR, but it can be faster than taking a PR from a fork, since it's a hassle to get CI to run on PRs from forks.

PR #895, opened 2022-03-29
Append pthread to platform libs if cmake supplied Threads is not good enough

serzhick:
*Issue #, if available:*
aws-c-common has an explicit dependency on pthread's pthread_mutexattr_settype [here in mutex.c ](https://github.com/awslabs/aws-c-common/blob/057746b2e094f4b7a31743d8ba5a9fd0155f69f3/source/posix/mutex.c#L26).
However, aws-c-common cmake script does not instruct cmake to *explicitly* link with pthread.

Appending fsanitize=address to CMAKE_CXX_FLAGS results in GLIBC library to provide ***some*** pthread API but not all. Therefore, `find_package(Threads REQUIRED)` starts to "find" fake pthread APIs in GLIBC and stops to set `CMAKE_THREAD_LIBS_INIT` (which is typically expected to contain Threads lib implementation (which is typically pthread)). In addition, if user sets OpenSSL as a crypto lib implementation, aws-lc is not being built too (aws-lc has a hard set lining dependency on pthread, aws-lc hides the issue being described/fixed in this pull request).

The end result is that aws-c-common fails to link (in case of fsanitize= and OpenSSL crypto) because pthread_mutexattr_settype symbol is not found.

*Description of changes:*
Link explicitly with pthread unless cmake found Threads library provides necessary symbol (i.e. still provide user with a theoretical ability to use custom Threads lib).


CMakeLists.txt
```diff
@@ -85,7 +85,15 @@ else ()
 
     if (UNIX OR APPLE)
         find_package(Threads REQUIRED)
-    endif ()
+
+        if (NOT WIN32 AND NOT ANDROID AND NOT CMAKE_THREAD_LIBS_INIT)
```
sdavtaker, line 89: I think the NOT WIN32 is redundant, the live above is `if UNIX OR APPLE`

sdavtaker (commented):
Looks good.

DmitriyMusatkin (approved):

PR #898, opened 2022-04-11
Added metric reporting functions

TwistedTwigleg:
*Description of changes:*

Adds metric report functions that are platform specific that can be used in the Device Defender custom metrics. Needed for the following PR: https://github.com/aws/aws-iot-device-sdk-cpp-v2/pull/401

_________


bretambrose (commented):
I wouldn't make these "custom_metric" I'd just make them top-level functions, likely adding to either process.* or system_info.*

Also, restricted to posix-only, does the implementation still need to check defines?  Does it work on a mac?

TwistedTwigleg:
> I wouldn't make these "custom_metric" I'd just make them top-level functions, likely adding to either process.* or system_info.*

Will do 👍 

> Also, restricted to posix-only, does the implementation still need to check defines? Does it work on a mac?

Unfortunately, as `sysinfo.h` isn't available on Mac. There is a [similar function](https://stackoverflow.com/questions/12523704/mac-os-x-equivalent-header-file-for-sysinfo-h-in-linux) that provides some of the information used, but since Device Defender doesn't support Mac I haven't added the support for it.

source/posix/system_info.c
```diff
@@ -445,3 +450,118 @@ enum aws_platform_os aws_get_platform_build_os(void) {
...
+        total_system < s_cpu_last_total_system || total_idle < s_cpu_last_total_idle) {
+        *output = 0;
+    } else {
+        total = (total_user - s_cpu_last_total_user) + (total_user_low - s_cpu_last_total_user_low) +
+                (total_system - s_cpu_last_total_system);
+        percent = total;
+        total += total_idle - s_cpu_last_total_idle;
+        percent = (percent / total) * 100;
+
+        // If percent is negative, then there was an error (overflow?)
```
bretambrose, line 508: Would be better to detect previous failures directly using the appropriate math functions (aws_*_u64_checked)

source/posix/system_info.c
```diff
@@ -445,3 +450,118 @@ enum aws_platform_os aws_get_platform_build_os(void) {
...
+    uint64_t total_user, total_user_low, total_system, total_idle, total;
+    s_getCurrentCpuUsage(&total_user, &total_user_low, &total_system, &total_idle);
+    double percent;
+
+    // Overflow detection
+    if (total_user < s_cpu_last_total_user || total_user_low < s_cpu_last_total_user_low ||
+        total_system < s_cpu_last_total_system || total_idle < s_cpu_last_total_idle) {
+        *output = 0;
+    } else {
+        total = (total_user - s_cpu_last_total_user) + (total_user_low - s_cpu_last_total_user_low) +
```
bretambrose, line 502: Would rather see some locals with good names here rather than reusing total/perecent multiple times.

source/posix/system_info.c
```diff
@@ -445,3 +450,118 @@ enum aws_platform_os aws_get_platform_build_os(void) {
...
+
+int aws_get_memory_usage(double *output) {
+// Get the Memory usage from Linux
+#if defined(__linux__) || defined(__unix__)
+    struct sysinfo memory_info;
+    sysinfo(&memory_info);
+    uint64_t physical_memory_used = memory_info.totalram - memory_info.freeram;
+    physical_memory_used *= memory_info.mem_unit;
+    // Return data in Kilobytes
+    physical_memory_used = physical_memory_used / (1024);
```
bretambrose, line 544: Could use named constant here

include/aws/common/system_info.h
```diff
@@ -100,6 +100,29 @@ void aws_backtrace_print(FILE *fp, void *call_site_data);
...
+ * Sets the passed double value as the memory usage in kilobytes. Only supported on Linux currently.
+ */
+AWS_COMMON_API
+int aws_get_memory_usage(double *output);
+
+/**
+ * Sets the passed double value as the process count. Only supported on Linux currently.
+ */
+AWS_COMMON_API
+int aws_get_process_count(double *output);
```
bretambrose, line 124: Can we make both of these uint64s?  If we have to convert to double later because of JSON, that's fine, but a general query in aws-c-common shouldn't be fractional

bretambrose (commented):
You'll find older things that don't follow this pattern, but currently new features tend to use the following pattern:

(all structures are private, either in a source file or a private header)

```
struct aws_blah {
   /* data used by all subclasses */

   /* pointer to vtable */

   void *impl; /* points to the start of the subclass */
};

struct aws_blah_subclass {
   struct aws_allocator *allocator;

   struct aws_blah base;

   /* subclass specific data */
};
```

your public operations, other than the constructors, defer to the vtable.  This includes destroy.

Your constructor functions (aws_blah_subclass_new()) return a pointer to the aws_blah inside the allocated aws_blah_subclass structure.

Your vtable-referenced subclass functions (which can be static and hidden) always convert the aws_blah that they are passed into an aws_blah_subclass by casting aws_blah->impl.

I'll send you a link to a recent example (that I can't put in a public comment) shortly.  It uses acquire/release rather than destroy, but other than that the pattern is the same.

source/cpu_usage_sampler.c
```diff
@@ -0,0 +1,238 @@
...
+    }
+    aws_mem_release(sampler->allocator, sampler);
+}
+
+static struct aws_cpu_sampler_vtable aws_cpu_sampler_vtable_default = {
+    .aws_get_cpu_sample_fn = NULL,
+    .aws_cpu_sampler_destroy = aws_cpu_sampler_destroy_default,
+};
+
+/*********************************************************************************************************************
```
bretambrose, line 40: You're on the right track in terms of data structures.  The missing component is that you want implementations in each of source/windows/ (empty, do nothing), source/darwin (empty, do nothing), and source/linux (implementation with no conditional compilation).  Unfortunately, we don't have darwin and linux folders in aws-c-common the same way we do in aws-c-io.  Seeing what people think right now.

source/cpu_usage_sampler.c
```diff
@@ -0,0 +1,238 @@
...
+
+    int return_result = AWS_OP_ERR;
+    uint64_t total_user, total_user_low, total_system, total_idle, total;
+    s_get_cpu_usage_linux(&total_user, &total_user_low, &total_system, &total_idle);
+    // total_combined needs to be double to allow for fractions
+    double percent, total_combined;
+
+    // If the value has overflown and wrapped around, swap the variables
+    uint64_t tmp;
+    if (total_user < sampler_linux->cpu_last_total_user) {
```
bretambrose, line 91: let's make a helper compute-difference function that does this overflow check and adjust for arbitrary (start, end) uint64 pairs.

Also I don't think we want to swap.  The overflow case should be handled by:

```
if (overflow) {
    diff = (AWS_UINT64_MAX - start) + end;
} else {
    diff = end - start;
}
```

source/cpu_usage_sampler.c
```diff
@@ -0,0 +1,238 @@
...
+    uint64_t *total_user_low,
+    uint64_t *total_system,
+    uint64_t *total_idle) {
+
+    FILE *file;
+    int matched_results;
+    file = fopen("/proc/stat", "r");
+    matched_results = fscanf(
+        file,
+        "cpu %llu %llu %llu %llu",
```
bretambrose, line 69: I think we want to use `SCNu64` here rather than `llu`.  

It's a macro so you need to reference it outside the string constant:

```
fscanf(file, "cpu %" SCNu64 " %" SCNu64 " %" SCNu64" %" SCNu64, ....); 
```

See how we use PRIu64 elsewhere in common.

https://en.cppreference.com/w/c/types/integer

source/linux/cpu_usage_sampler.c
```diff
@@ -0,0 +1,196 @@
...
+}
+
+static int aws_get_cpu_sample_fn_linux_get_uint64_delta(uint64_t *first, uint64_t *second, uint64_t *output) {
+    int64_t ret_val = AWS_OP_SUCCESS;
+    if (*first > *second) {
+        ret_val = aws_sub_u64_checked(*first, *second, output);
+    } else {
+        ret_val = aws_add_u64_checked((UINT64_MAX - *second), *first, output);
+    }
+    return ret_val;
```
bretambrose, line 72: In fact, the whole function can't fail now, so let's ditch the return value and then remove all of those if-statements in the caller.

bretambrose (approved):

TwistedTwigleg:
Thanks for the reviews! Merging into `main`...

source/cpu_usage_sampler.c
```diff
@@ -0,0 +1,33 @@
...
+}
+
+int aws_cpu_sampler_get_sample(struct aws_cpu_sampler *sampler, double *output) {
+    if (sampler == NULL) {
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        return AWS_OP_ERR;
+    }
+    if (sampler->vtable->aws_get_cpu_sample_fn == NULL) {
+        aws_raise_error(AWS_ERROR_PLATFORM_NOT_SUPPORTED);
+        return AWS_OP_ERR;
```
graebm, line 30: don't bother checking if the vtable functions are set or not
a valid sampler must have all its vtable functions set

include/aws/common/cpu_usage_sampler.h
```diff
@@ -0,0 +1,52 @@
+#ifndef AWS_COMMON_CPU_USAGE_SAMPLER_H
+#define AWS_COMMON_CPU_USAGE_SAMPLER_H
```
graebm, line 2: trivial/debatable: this doesn't need to be its own header, the aws_cpu_sampler stuff would fit in fine with the other CPU functions in system_info.h

include/aws/common/cpu_usage_sampler.h
```diff
@@ -0,0 +1,52 @@
...
+ * Frees the memory used by the CPU sampler.
+ */
+AWS_COMMON_API
+void aws_cpu_sampler_destroy(struct aws_cpu_sampler *sampler);
+
+/**
+ * Gets the CPU usage and populates the given double, output, with the value. The value
+ * returned is a percentage from 0.0 to 100.0.
+ *
+ * Will return AWS_OP_SUCCESS if polling the CPU was successful. AWS_OP_ERR will be returned
```
graebm, line 42: in the comments, mention that it's getting the usage since the last sample was taken

include/aws/common/cpu_usage_sampler.h
```diff
@@ -0,0 +1,52 @@
+#ifndef AWS_COMMON_CPU_USAGE_SAMPLER_H
+#define AWS_COMMON_CPU_USAGE_SAMPLER_H
+
+/**
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0.
+ */
+
+#include <aws/common/common.h>
+#include <aws/common/math.h>
```
graebm, line 10: there's no need to include math.h in this file

include/aws/common/cpu_usage_sampler.h
```diff
@@ -0,0 +1,52 @@
...
+ * to free the memory even for unsupported platforms.
+ */
+AWS_COMMON_API
+struct aws_cpu_sampler *aws_cpu_sampler_new(struct aws_allocator *allocator);
+
+/**
+ * Frees the memory used by the CPU sampler.
+ */
+AWS_COMMON_API
+void aws_cpu_sampler_destroy(struct aws_cpu_sampler *sampler);
```
graebm, line 36: will this class be bound to higher-level languages? if so do ref-counted acquire()/release() instead of destroy()

if there are no plans then this is fine, we can always add them later if needed

source/linux/cpu_usage_sampler.c
```diff
@@ -0,0 +1,140 @@
...
+}
+
+static int aws_get_cpu_sample_fn_linux(struct aws_cpu_sampler *sampler, double *output) {
+    struct aws_cpu_sampler_linux *sampler_linux = sampler->impl;
+
+    int return_result = AWS_OP_ERR;
+    uint64_t total_user, total_user_low, total_system, total_idle, total;
+    s_get_cpu_usage_linux(&total_user, &total_user_low, &total_system, &total_idle);
+    // total_combined needs to be double to allow for fractions
+    double percent, total_combined;
```
graebm, line 63: stop declaring variables early unless it's technically necessary

source/linux/cpu_usage_sampler.c
```diff
@@ -0,0 +1,140 @@
...
+        *output = 0;
+        return_result = AWS_OP_ERR;
+        goto cleanup;
+    }
+
+    percent = (total_combined / total) * 100;
+
+    // If negative, there was an error (overflow?)
+    if (percent < 0) {
+        *output = 0;
```
graebm, line 86: Just clamp it to 0. Floating point math is weird and it's actually not hard to go sliiiiiightly below 0 (not that it can happen with the exact math you have here, but in general)

Likewise, be sure to clamp it to the max (I suggested 1.0, but if you stick with 100.0 clamp it to that)

include/aws/common/system_info.h
```diff
@@ -100,6 +100,18 @@ void aws_backtrace_print(FILE *fp, void *call_site_data);
 AWS_COMMON_API
 void aws_backtrace_log(int log_level);
 
+/**
+ * Sets the passed int value as the memory usage in kilobytes. Only supported on Linux currently.
+ */
+AWS_COMMON_API
+int aws_get_memory_usage(int64_t *output);
```
graebm, line 107: I would report bytes, not KB

maybe Device Defender wants KB, but bytes are more useful

graebm: aws_get_memory_usage() sounds like it's querying THIS PROCESS's memory usage

I would name this something like aws_system_memory_usage() to be clear it's a global system query

include/aws/common/cpu_usage_sampler.h
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Creates a new CPU sampler using the provided allocator, or will return NULL if there is an error.
+ *
+ * Note: On unsupported platforms, the CPU sampler returned will return AWS_OP_ERR when calling
+ * aws_cpu_sampler_get_sample. You will still need to call aws_cpu_sampler_clean_up when finished
+ * to free the memory even for unsupported platforms.
+ */
+AWS_COMMON_API
+struct aws_cpu_sampler *aws_cpu_sampler_new(struct aws_allocator *allocator);
```
graebm, line 30: I would name this something like aws_system_cpu_sampler to be clear it's querying the system, and not this process or this library

source/linux/cpu_usage_sampler.c
```diff
@@ -0,0 +1,140 @@
...
+    if (output_linux == NULL) {
+        return NULL;
+    }
+    output_linux->base.allocator = allocator;
+    output_linux->base.vtable = &aws_cpu_sampler_vtable_linux;
+    output_linux->base.impl = output_linux;
+
+    // CPU reporting is done via deltas, so we need to cache the initial CPU values
+    double tmp = 0;
+    aws_get_cpu_sample_fn_linux(&output_linux->base, &tmp);
```
graebm, line 137: you could check for errors from the initial sample to see if our sampling code actually works on this platform.

source/posix/system_info.c
```diff
@@ -445,3 +450,40 @@ enum aws_platform_os aws_get_platform_build_os(void) {
...
+#if defined(__linux__) || defined(__unix__)
+    struct sysinfo memory_info;
+    sysinfo(&memory_info);
+    uint64_t physical_memory_used = memory_info.totalram - memory_info.freeram;
+    physical_memory_used *= memory_info.mem_unit;
+    // Return data in Kilobytes
+    physical_memory_used = physical_memory_used / S_BYTES_TO_KILO_BYTES;
+    *output = physical_memory_used;
+    return AWS_OP_SUCCESS;
+#endif
```
graebm, line 469: trivial: `#else` instead of just #endif here. I could see some compiler complaining that the rest of the code is unreachable

source/windows/system_info.c
```diff
@@ -299,4 +299,19 @@ void aws_backtrace_log() {
     AWS_LOGF_TRACE(
         AWS_LS_COMMON_GENERAL, "aws_backtrace_log: backtrace requested, but logging is unsupported on this platform");
 }
+
+int aws_get_memory_usage(int64_t *output) {
+    // Windows is currently not supported.
+    *output = 0;
+    aws_raise_error(AWS_ERROR_PLATFORM_NOT_SUPPORTED);
```
graebm, line 306: surely windows offers something like this ...

googling found: [GlobalMemoryStatusEx()](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-globalmemorystatusex)

I'm hesitant to add simple functions to aws-c-common that we didn't implement for common platforms.

source/windows/system_info.c
```diff
@@ -299,4 +299,19 @@ void aws_backtrace_log() {
...
+    return AWS_OP_ERR;
+}
+
+int aws_get_process_count(int64_t *output) {
+    // Windows is currently not supported.
+    *output = 0;
```
graebm, line 312: [EnumProcesses()](https://docs.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-enumprocesses)

include/aws/common/system_info.h
```diff
@@ -100,6 +100,18 @@ void aws_backtrace_print(FILE *fp, void *call_site_data);
...
+
+/**
+ * Sets the passed int value as the process count. Only supported on Linux currently.
+ */
+AWS_COMMON_API
+int aws_get_process_count(int64_t *output);
```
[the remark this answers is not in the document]
graebm: also: aws_system_process_count() would be more clear that we're not talking about this application or library

source/linux/cpu_usage_sampler.c
```diff
@@ -0,0 +1,140 @@
...
+    if (matched_results == EOF || matched_results != 4) {
+        aws_raise_error(AWS_ERROR_INVALID_STATE);
+    }
+}
+
+static void aws_get_cpu_sample_fn_linux_get_uint64_delta(uint64_t first, uint64_t second, uint64_t *output) {
+    if (first > second) {
+        aws_sub_u64_checked(first, second, output);
+    } else {
+        aws_add_u64_checked((UINT64_MAX - second), first, output);
```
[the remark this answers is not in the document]
bretambrose: I think you can theoretically overflow the after vs. the before,which this is checking.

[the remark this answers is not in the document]
graebm: the `_checked()` functions literally do nothing (don't set the result) if there's going to be overflow. Using these functions without checking if they succeeded or failed is ... very confusing

the `_saturated()` functions were meant for exactly this use case, we should be using them instead

PR #901, opened 2022-04-21
Remove DeviceDefender custom metric reporting functions

TwistedTwigleg:
*Description of changes:*

~~Adjusts the metric reporting functions added in this PR, addressing the code review changes: https://github.com/awslabs/aws-c-common/pull/898~~

~~Adjusted to better fit code style, added Windows support for process and memory statistics, general code improvements~~

**EDIT**: Previously this PR was to make some changes to extend/enhance the metric reporting functions I introduced earlier in `aws-c-common`, however they are no longer needed for the task I was working on and since nothing else currently needs them, this PR is instead to remove that functionality.

**NOTE: The comments prior to the comment saying** "UPDATE - Removed metric reporting functions" are made in regards to the prior purpose of this PR, not the current purpose.

_________


source/linux/cpu_usage_sampler.c
```diff
@@ -53,33 +61,41 @@ static void aws_get_cpu_sample_fn_linux_get_uint64_delta(uint64_t first, uint64_
...
-        *output = 0;
-        return_result = AWS_OP_ERR;
-        goto cleanup;
+    double percent = 0;
+    if (total != 0) {
+        percent = (total_combined / total);
     }
 
-    percent = (total_combined / total) * 100;
+    // Clamp to min/max
```
sbSteveK, line 93: You may only need to check for > 1 since immediately following there's a < 0 check that will result in an error

sbSteveK (approved):

include/aws/common/private/cpu_usage_sampler_private.h
```diff
@@ -12,16 +12,16 @@
 /**
  * The VTable for the CPU sampler in cpu_usage_sampler.h
  */
-struct aws_cpu_sampler_vtable {
-    int (*aws_get_cpu_sample_fn)(struct aws_cpu_sampler *sampler, double *output);
-    void (*aws_cpu_sampler_destroy)(struct aws_cpu_sampler *sampler);
+struct aws_system_cpu_sampler_vtable {
+    int (*aws_get_cpu_sample_fn)(struct aws_system_cpu_sampler *sampler, double *output);
+    void (*aws_system_cpu_sampler_destroy)(struct aws_system_cpu_sampler *sampler);
```
graebm, line 17: trivial: just name these vtable members like:
- ~aws_get_cpu_sample_fn~ -> `get_sample`
- ~aws_system_cpu_sampler_destroy~ -> `destroy`

If we're `typedef`ing a callback we do the full aws_ name and put _fn at the end. But for vtable member names there's no need for all the fanfare

source/linux/cpu_usage_sampler.c
```diff
@@ -53,39 +61,40 @@ static void aws_get_cpu_sample_fn_linux_get_uint64_delta(uint64_t first, uint64_
...
+    struct aws_system_cpu_sampler_linux *sampler_linux = sampler->impl;
 
     int return_result = AWS_OP_ERR;
-    uint64_t total_user, total_user_low, total_system, total_idle, total;
-    s_get_cpu_usage_linux(&total_user, &total_user_low, &total_system, &total_idle);
-    // total_combined needs to be double to allow for fractions
-    double percent, total_combined;
-
+    // Have to define these here for cleanup
+    uint64_t total_user = 0, total_user_low = 0, total_system = 0, total_idle = 0;
```
graebm, line 69: the way it is now, if this function fails, we end up storing `sampler_linux->cpu_last_total_user = 0` before returning.

It's generally a best-practice to leave the state of something unchanged if there's an error. The way it is now, if there's a 1-time failure the stored value will be zeroed out, and the next sample to succeed is going to look really weird because it will be a delta from 0.

You probably don't want to update ``sampler_linux->cpu_last_total_user` and friends unless the function has completely succeeded. You might be able to get away from using `goto` at all in this function, since it turns out there's nothing really to clean up, and just go with early-return on failure

source/posix/system_info.c
```diff
@@ -451,39 +456,65 @@ enum aws_platform_os aws_get_platform_build_os(void) {
...
     sysinfo(&system_info);
-    *output = (int64_t)system_info.procs;
+    *output = (uint64_t)system_info.procs;
     return AWS_OP_SUCCESS;
-#endif
+
+#elif defined(__APPLE__)
+
+    pid_t pid_array[2048];
+    int pid_bytes = proc_listpids(PROC_ALL_PIDS, 0, pid_array, sizeof(pid_array));
```
graebm, line 507: Cool thing: a lot of C libs that have functions like this, where you need to pass in a block of memory for it to fill out, will TELL you how much memory is needed if you pass in NULL

so we can remove the useless 2KB stack variable and just do:
~pid_t pid_array[2048];~
int pid_bytes = proc_listpids(PROC_ALL_PIDS, 0, NULL, 0);
...
return pid_bytes / sizeof(pid_t);

I suspected this was a thing, but couldn't find actual documentation on this function, but found a [blog post](https://tramdas.wordpress.com/2015/10/21/libproc-for-process-listing-on-osx/) where someone is calling it this way.

source/posix/system_info.c
```diff
@@ -451,39 +456,65 @@ enum aws_platform_os aws_get_platform_build_os(void) {
...
+    if (pid_bytes == 0) {
```
graebm, line 508: probably want to check for values less than 0 too

I can't find actual documentation for this function, but here's the [header](https://opensource.apple.com/source/xnu/xnu-2422.1.72/libsyscall/wrappers/libproc/libproc.h.auto.html) online and a nearby function has docs that say it returns -1 on error

source/windows/system_info.c
```diff
@@ -300,18 +300,30 @@ void aws_backtrace_log() {
...
-int aws_get_process_count(int64_t *output) {
-    // Windows is currently not supported.
-    *output = 0;
-    aws_raise_error(AWS_ERROR_PLATFORM_NOT_SUPPORTED);
-    return AWS_OP_ERR;
+int aws_get_system_process_count(uint64_t *output) {
+    DWORD processes_array[1024], processes_array_size, processes_count;
+    if (EnumProcesses(processes_array, sizeof(processes_array), &processes_array_size) == 0) {
+        // Try a larger array
+        DWORD processes_array_big[2048];
```
graebm, line 319: if this function fails, retring with a bigger array won't fix anything, the [docs](https://docs.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-enumprocesses) say this function totally expects that the array might be too small, and it will just result in `processes_array_size` being equal to `sizeof(processes_array`

> To determine how many processes were enumerated, divide the lpcbNeeded value by sizeof(DWORD). There is no indication given when the buffer is too small to store all process identifiers. Therefore, if lpcbNeeded equals cb, consider retrying the call with a larger array.

I'm personally sad that this function doesn't have a mode where you pass NULL and it tells you what's needed

source/windows/system_info.c
```diff
@@ -300,18 +300,30 @@ void aws_backtrace_log() {
...
+    return AWS_OP_SUCCESS;
 }
 
-int aws_get_process_count(int64_t *output) {
-    // Windows is currently not supported.
-    *output = 0;
-    aws_raise_error(AWS_ERROR_PLATFORM_NOT_SUPPORTED);
-    return AWS_OP_ERR;
+int aws_get_system_process_count(uint64_t *output) {
+    DWORD processes_array[1024], processes_array_size, processes_count;
```
graebm, line 316: `processes_count` is never used.

source/windows/system_info.c
```diff
@@ -300,18 +300,30 @@ void aws_backtrace_log() {
...
+    if (EnumProcesses(processes_array, sizeof(processes_array), &processes_array_size) == 0) {
+        // Try a larger array
+        DWORD processes_array_big[2048];
+        if EnumProcesses(processes_array_big, sizeof(processes_array_big), &processes_array_size) == 0) {
+                // Error twice - something is wrong!
+                return aws_raise_error(AWS_ERROR_INVALID_STATE);
+            }
+    }
+    *output = (double)(processes_array_size / sizeof(DWORD));
```
graebm, line 325: don't drag floating-point types into whole-number math

`output` is uint64_t, so if you need to do any casting, cast to that

include/aws/common/system_info.h
```diff
@@ -104,13 +104,13 @@ void aws_backtrace_log(int log_level);
  * Sets the passed int value as the memory usage in kilobytes. Only supported on Linux currently.
  */
 AWS_COMMON_API
-int aws_get_memory_usage(int64_t *output);
+int aws_get_system_memory_usage(uint64_t *output);
```
graebm, line 107: update comments, usage is now in bytes

CMakeLists.txt
```diff
@@ -101,22 +101,19 @@ else ()
...
+        list (APPEND AWS_COMMON_OS_SRC "source/darwin/*.c") # OS specific includes
     elseif (${CMAKE_SYSTEM_NAME} STREQUAL "Linux") # Android does not link to libpthread nor librt, so this is fine
         list(APPEND PLATFORM_LIBS dl m Threads::Threads rt)
+        list (APPEND AWS_COMMON_OS_SRC "source/linux/*.c") # OS specific includes
     elseif(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
         list(APPEND PLATFORM_LIBS dl m thr execinfo)
+        list (APPEND AWS_COMMON_OS_SRC "source/linux/*.c") # OS specific includes
     elseif(CMAKE_SYSTEM_NAME STREQUAL "NetBSD")
         list(APPEND PLATFORM_LIBS dl m Threads::Threads execinfo)
+        list (APPEND AWS_COMMON_OS_SRC "source/linux/*.c") # OS specific includes
```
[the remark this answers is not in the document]
TwistedTwigleg: Should I look into add `BSD` support? Or should I just leave it as an unsupported platform for now?

source/linux/cpu_usage_sampler.c
```diff
@@ -53,39 +61,40 @@ static void aws_get_cpu_sample_fn_linux_get_uint64_delta(uint64_t first, uint64_
...
     aws_get_cpu_sample_fn_linux_get_uint64_delta(
         total_user_low, sampler_linux->cpu_last_total_user_low, &total_user_low_delta);
     aws_get_cpu_sample_fn_linux_get_uint64_delta(
         total_system, sampler_linux->cpu_last_total_system, &total_system_delta);
     aws_get_cpu_sample_fn_linux_get_uint64_delta(total_idle, sampler_linux->cpu_last_total_idle, &total_idle_delta);
 
-    total_combined = (double)(total_user_delta) + (double)(total_user_low_delta) + (double)(total_system_delta);
-    total = total_combined + (double)(total_idle_delta);
+    double total_combined = (double)(total_user_delta) + (double)(total_user_low_delta) + (double)(total_system_delta);
+    uint64_t total = total_combined + (double)(total_idle_delta);
```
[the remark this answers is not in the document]
TwistedTwigleg: By accident. It is supposed to be a double 😅 
Will keep it a double this time.

TwistedTwigleg:
**UPDATE - Removed metric reporting functions**. This is because, upon further discussion and review of the task for Device Defender custom metrics, the CPU, process, and memory reporting functions are not needed to be implemented in `aws-c-common`. Therefore, since they are no longer needed, this **PR has been adjusted to remove these functions rather than  extend/enhance them.**

TwistedTwigleg:
Thanks! Merging into `main`...

PR #904, opened 2022-04-25
Apple Universal binary support

graebm:
work in progres...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,6 +147,26 @@ jobs:
         chmod a+x builder
         ./builder build -p ${{ env.PACKAGE_NAME }} --cmake-extra=-DUSE_CPU_EXTENSIONS=OFF
 
+  # build universal binary that works on both x86_64 and arm64
+  osx-universal2-static-libs:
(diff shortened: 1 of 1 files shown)


graebm:
closing.

This already compiles for arm64. If you want a [Universal Binary](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) you must build the x86_64 library, then build the arm64 library, then glue them together into a Universal Binary using `lipo`.

This PR was an attempt to get a Universal Binary in a single cmake build by configuring with `DCMAKE_OSX_ARCHITECTURES=arm64;x86_64`. This really confuses our configuration scripts because they do lots checks assuming that they are checking a single platform/arch/compiler. For example "does this compiler accept the `-mavx2` flag?". This flag is specific to the x86_64 architecture. The compiler will not accept the flag when compiling for both x86_64 and arm64 at the same time.

I looked to see how XCode tries to handle stuff like this, and XCode also seems to just compile the x86_64 and arm64 stuff separately, then glue it together. I'm following XCode's lead on this. We do not support building multiple architectures simultaneously.

PR #905, opened 2022-04-30
aws_byte_cursor_starts_with(input, prefix)

graebm:
more string helper functions


source/byte_buf.c
```diff
@@ -558,6 +558,42 @@ bool aws_byte_cursor_eq_c_str_ignore_case(const struct aws_byte_cursor *const cu
...
+
+    if (input->len < prefix->len) {
+        return false;
+    }
+
+    struct aws_byte_cursor start = {.ptr = input->ptr, .len = prefix->len};
+    bool rv = aws_byte_cursor_eq(&start, prefix);
+
+    AWS_POSTCONDITION(aws_byte_cursor_is_valid(input));
+    AWS_POSTCONDITION(aws_byte_cursor_is_valid(prefix));
```
TingDaoK, line 574: trivial/question: they are both const, do we still need to check they are valid afterward?

TingDaoK (approved):
Question and 🚢

source/byte_buf.c
TingDaoK: ignore it. you can still change the content inside...

graebm: I was just doing checks the same way that other const functions in this file were doing it 🤷‍♀️

PR #906, opened 2022-05-02
Exclude libs

DmitriyMusatkin:
Exclude libcrypto symbols from test runners. This is mitigation to avoid tests breaking when trying to load libs that were linked against other versions of libcrypto.


graebm:
Used this technique in a PR on a different branch
https://github.com/awslabs/aws-c-common/pull/925

PR #911, opened 2022-05-17
Adjust logging function to fix segfault if filename and file are not set

TwistedTwigleg:
*Description of changes:*

Adjusted `aws_logger_init_noalloc` to no longer assume a file name has  been set. If no file or filename is set, it will now use `stderr` ~~and log a warning~~. This fixes an issue we are seeing in some testing.

Edit: Now it just defaults to `stderr`, as logging a warning was causing issues on Windows. I also updated the header to note that it will default to `stderr` if `filename` and `file` are both not set.

___________

### DIFF
--- a/source/logging.c
+++ b/source/logging.c
@@ -547,8 +547,16 @@ int aws_logger_init_noalloc(
         impl->file = options->file;
         impl->should_close = false;
     } else { /* _MSC_VER */
-        impl->file = aws_fopen(options->filename, "w");
-        impl->should_close = true;
+        if (options->filename != NULL) {
+            impl->file = aws_fopen(options->filename, "w");
+            impl->should_close = true;
+        } else {
+            impl->file = stderr;
+            impl->should_close = false;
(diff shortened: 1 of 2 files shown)


source/logging.c
```diff
@@ -547,8 +547,13 @@ int aws_logger_init_noalloc(
         impl->file = options->file;
         impl->should_close = false;
     } else { /* _MSC_VER */
-        impl->file = aws_fopen(options->filename, "w");
-        impl->should_close = true;
+        if (options->filename != NULL) {
```
graebm, line 550: unrelated: I have no idea what this `/* _MSC_VER */` is about

graebm (approved):

source/logging.c
TwistedTwigleg: Yeah, I just left it because it was there. I'm guessing that maybe it was to fix a MSVC issue at one point?

TwistedTwigleg:
Thanks! Merging into `main`...

PR #915, opened 2022-06-20
Make sure file paths are not empty strings to avoid segfault on Windows

TwistedTwigleg:
*Description of changes:*

Adds some conditions to check to make sure empty strings are not processed if they are passed to file operating functions. This prevents segfaults on Java and C++ that occur if you pass an empty string for the key or certificate path when creating an MQTT client.

__________

### DIFF
--- a/source/file.c
+++ b/source/file.c
@@ -12,6 +12,11 @@
 #include <errno.h>
 
 FILE *aws_fopen(const char *file_path, const char *mode) {
+    if (!file_path || strlen(file_path) == 0) {
+        AWS_LOGF_ERROR(AWS_LS_COMMON_IO, "static: Failed to open file %s", file_path);
+        return NULL;
+    }
+
     struct aws_string *file_path_str = aws_string_new_from_c_str(aws_default_allocator(), file_path);
     struct aws_string *mode_str = aws_string_new_from_c_str(aws_default_allocator(), mode);
 
@@ -23,6 +28,11 @@ FILE *aws_fopen(const char *file_path, const char *mode) {
 }
 
 int aws_byte_buf_init_from_file(struct aws_byte_buf *out_buf, struct aws_allocator *alloc, const char *filename) {
+    if (!filename || strlen(filename) == 0) {
+        AWS_LOGF_ERROR(AWS_LS_COMMON_IO, "static: Failed to open file %s", filename);
+        return aws_raise_error(AWS_ERROR_FILE_INVALID_PATH);
+    }
+
     AWS_ZERO_STRUCT(*out_buf);
     FILE *fp = aws_fopen(filename, "rb");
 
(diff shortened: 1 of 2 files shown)


source/file.c
```diff
@@ -23,6 +28,11 @@ FILE *aws_fopen(const char *file_path, const char *mode) {
 }
 
 int aws_byte_buf_init_from_file(struct aws_byte_buf *out_buf, struct aws_allocator *alloc, const char *filename) {
+    if (!filename || strlen(filename) == 0) {
```
bretambrose, line 31: This seems redundant given the addition to aws_fopen

source/windows/file.c
```diff
@@ -440,6 +468,10 @@ struct aws_string *aws_get_home_directory(struct aws_allocator *allocator) {
 }
 
 bool aws_path_exists(const struct aws_string *path) {
+    if (!path || aws_string_eq_c_str(path, "")) {
```
bretambrose, line 471: these empty string checks can just be "string->len == 0"

source/windows/file.c
```diff
@@ -14,6 +14,10 @@
 #include <windows.h>
```
bretambrose, line 15: I'd make a static s_is_non_empty(...) that checks for non-null and string len > 0

source/file.c
TwistedTwigleg: I modified this so now it checks to see if `errno` is `0` after calling `aws_fopen`, which should mean that the file returned is `null` and, if it is, then it will return `AWS_ERROR_FILE_INVALID_PATH` rather than `aws_translate_and_raise_io_error(errno)`. 👍

source/windows/file.c
```diff
@@ -13,7 +13,19 @@
 #include <stdio.h>
 #include <windows.h>
 
+static bool s_is_string_non_empty(const struct aws_string *str) {
+    return str != NULL && str->len > 0;
+}
+
+static bool s_is_wstring_non_empty(const struct aws_wstring *str) {
```
sbSteveK, line 20: ooc, this looks to be identical to the above s_is_string_non_empty(). Is there a reason to have a separate function for aws_wstring? Or can we use the above for both without issue?

source/windows/file.c
```diff
@@ -13,7 +13,19 @@
 #include <stdio.h>
 #include <windows.h>
 
+static bool s_is_string_non_empty(const struct aws_string *str) {
```
sbSteveK, line 16: Absolutely personal preference and trivial but since every instance of this being used is "!s_is_string_non_empty" I feel like this should be "s_is_string_empty" to make the checks a positive one instead of a double negative.

source/windows/file.c
sbSteveK: Ah, gotcha. That makes sense =P

TwistedTwigleg:
Thanks! Merging into `main`..

PR #920, opened 2022-07-20
hide json module init and cleanup

TingDaoK:
`aws_json_module_init` should only be called once and called with the lib initial, so, hide it from other lib.

### DIFF
--- a/include/aws/common/json.h
+++ b/include/aws/common/json.h
@@ -6,10 +6,10 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
+#include <aws/common/byte_buf.h>
 #include <aws/common/common.h>
 
 struct aws_json_value;
-struct aws_byte_buf;
 
 // ====================
 // Create and pass type
--- a/include/aws/common/json.h
+++ b/include/aws/common/json.h
@@ -6,10 +6,10 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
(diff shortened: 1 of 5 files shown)


DmitriyMusatkin (approved):

include/aws/common/json.h
```diff
@@ -6,10 +6,10 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
-#include <aws/common/byte_buf.h>
 #include <aws/common/common.h>
 
 struct aws_json_value;
+struct aws_byte_buf;
```
graebm, line 13: can't remove `#include <aws/common/byte_buf.h>`

we pass `struct aws_byte_cursor` by value in this file, so we need to know its actual size

include/aws/common/json.h
```diff
@@ -268,19 +268,6 @@ bool aws_json_value_is_object(const struct aws_json_value *value);
...
- * @param allocator The allocator to use for creating aws_json_value structs.
- */
-AWS_COMMON_API
-void aws_json_module_init(struct aws_allocator *allocator);
-
-/**
- * Cleans up the JSON module. Should be called when finished using the module.
- */
-AWS_COMMON_API
-void aws_json_module_cleanup(void);
```
graebm, line 282: make sure we cut a new release of aws-c-s3 as well, that removes its call to this function

graebm (approved):
fix & ship

PR #924, opened 2022-08-15
Fall back for load target

TingDaoK:
*Issue #, if available:*

- We load target based on the `BUILD_SHARED_LIBS`, which will lead to a problem when the lib was built with `BUILD_SHARED_LIBS` on, and the app built with `BUILD_SHARED_LIBS` off and the app will fail to load the path.
- related https://github.com/aws4embeddedlinux/meta-aws/issues/178

*Description of changes:*

- Double fallback for now, we will introduce new cmake config to let user decide to enforce the shared/static lib to load.


DmitriyMusatkin (approved):

CMakeLists.txt
```diff
@@ -277,6 +277,7 @@ list(APPEND EXPORT_MODULES
     "cmake/AwsSIMD.cmake"
     "cmake/AwsFindPackage.cmake"
     "cmake/AwsFeatureTests.cmake"
+    "cmake/AwsCrtLoadTarget.cmake"
```
graebm, line 280: I'm worried that an application with a dead-simple CMakeLists.txt file won't find this file

test it out

graebm (approved):

PR #937, opened 2022-10-01
Set thread names, for debugging

graebm:
Also, rename a few choice functions where people often think our code is deadlocked, but it's not.

Using CamelCase names like "AwsHostResolver" because 15 characters is the max for Linux, so we can't do "aws_host_resolver"


source/posix/thread.c
```diff
@@ -336,22 +375,22 @@ int aws_thread_launch(
...
+            case EPERM:
+                return aws_raise_error(AWS_ERROR_THREAD_NO_PERMISSIONS);
+            case ENOMEM:
+                return aws_raise_error(AWS_ERROR_OOM);
+            default:
+                return aws_raise_error(AWS_ERROR_UNKNOWN);
+        }
     }
-
     return AWS_OP_SUCCESS;
```
graebm, line 394: Fix some fishy error handling.
The previous code forgot to free memory, and could report success if the error wasn't one of the 4 it expected

source/windows/thread.c
```diff
@@ -8,6 +8,7 @@
 #include <aws/common/linked_list.h>
 #include <aws/common/logging.h>
 #include <aws/common/private/thread_shared.h>
+#include <aws/common/string.h>
 
 #include <Windows.h>
```
sbSteveK, line 14: If aws_thread_options s_default_options is used from below and its .name isn't set, it might contain garbage as the thread name. Should we zero it out since it's not a pointer?

sbSteveK (approved):

source/log_channel.c
```diff
@@ -127,7 +127,19 @@ static bool s_background_wait(void *context) {
...
-static void s_background_thread_writer(void *thread_data) {
+/**
+ * This is where the background thread spends 99.999% of its time.
+ * We broke this out into its own function so that the stacktrace clearly shows
+ * what this thread is doing. We've had a lot of cases where users think this
+ * thread is deadlocked because it's stuck here. We want it to be clear
+ * that it's doing nothing on purpose. It's waiting for log messages...
+ */
+AWS_NO_INLINE
+static void aws_logger_listen_for_messages(struct aws_log_background_channel *impl) {
```
bretambrose, line 138: maybe add background to name

source/log_channel.c
```diff
@@ -127,7 +127,19 @@ static bool s_background_wait(void *context) {
...
+    aws_condition_variable_wait_pred(&impl->pending_line_signal, &impl->sync, s_background_wait, impl);
+}
+
+static void aws_logger_thread(void *thread_data) {
```
bretambrose, line 142: background

source/log_channel.c
```diff
@@ -217,10 +230,10 @@ int aws_log_channel_init_background(
     /*
      * Logging thread should need very little stack, but let's defer this to later
      */
-    struct aws_thread_options thread_options = {.stack_size = 0};
+    struct aws_thread_options thread_options = *aws_default_thread_options();
+    thread_options.name = aws_byte_cursor_from_c_str("AwsLogger");
```
bretambrose, line 234: Background

bretambrose (approved):

source/log_channel.c
graebm: Linux has a 15 char limit on thread names :(
Keep in mind this IS the name of a thread, and only shows up in a debugger when listing threads by name, so it's sort of implicit that it's in the background

Here's me brainstorming some more names
```
123456789012345
AwsBkGrndLogger
AwsBGLogger
```

I'm open to suggestions... bit I still like "AwsLogger" best

PR #939, opened 2022-10-05
fix parsing of ipv6 literals in uris

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*

Fix uri parsing for uris that include ipv6 literal.


source/uri.c
```diff
@@ -350,7 +350,21 @@ static void s_parse_authority(struct uri_parser *parser, struct aws_byte_cursor
...
+                parser->state = ERROR;
+                aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
+                return;
+            }
+        }
+
+        uint8_t *port_delim = memchr(port_search_start, ':', authority_parse_csr.len);
 
         if (!port_delim) {
             parser->uri->port = 0;
```
graebm, line 370: commenting on code below this...

if the URI has brackets in it, should the `parser->uri->host_name` also have the brackets in it? Or are we supposed to strip them out when users query later what the host_name is?

source/uri.c
```diff
@@ -350,7 +350,21 @@ static void s_parse_authority(struct uri_parser *parser, struct aws_byte_cursor
...
             }
         }
-        uint8_t *port_delim = memchr(authority_parse_csr.ptr, ':', authority_parse_csr.len);
+
+        /* RFC-3986 section 3.2: host identified by IPv6 literal address is
+         * enclosed within square brackets. We must ignore any colons within
+         * IPv6 literals and only search for port delimiter after closing bracket.*/
+        uint8_t *port_search_start = authority_parse_csr.ptr;
+        if (authority_parse_csr.ptr[0] == '[') {
+            port_search_start = memchr(authority_parse_csr.ptr, ']', authority_parse_csr.len);
```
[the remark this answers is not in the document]
DmitriyMusatkin: not legal.
uri spec calls for wrapping ipv6 in brackets and transforming % to %25. ipv6 on its own does not have brackets as valid chars

source/uri.c
```diff
@@ -350,7 +350,21 @@ static void s_parse_authority(struct uri_parser *parser, struct aws_byte_cursor
                 parser->uri->user = userinfo_parse_csr;
             }
         }
-        uint8_t *port_delim = memchr(authority_parse_csr.ptr, ':', authority_parse_csr.len);
+
+        /* RFC-3986 section 3.2: host identified by IPv6 literal address is
+         * enclosed within square brackets. We must ignore any colons within
+         * IPv6 literals and only search for port delimiter after closing bracket.*/
+        uint8_t *port_search_start = authority_parse_csr.ptr;
+        if (authority_parse_csr.ptr[0] == '[') {
```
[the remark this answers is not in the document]
DmitriyMusatkin: fair point. will add checks

source/uri.c
DmitriyMusatkin: my feeling on this is that we should preserve ipv6 encoding in host_name. undoing uri should be a separate concern, but lets discuss

DmitriyMusatkin: looking at how other people implement it. its pretty common to keep brackets on authority, but strip them on host_name, thoughts?

graebm: I want to do whatever the spec says

graebm: [Spec says](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2)
```
   authority     = [ userinfo "@" ] host [ ":" port ]
   userinfo      = *( unreserved / pct-encoded / sub-delims / ":" )
   host          = IP-literal / IPv4address / reg-name
   port          = *DIGIT

   IP-literal    = "[" ( IPv6address / IPvFuture  ) "]"
```

graebm: and we're doing that!
yay

source/uri.c
```diff
@@ -350,7 +350,21 @@ static void s_parse_authority(struct uri_parser *parser, struct aws_byte_cursor
...
+        if (authority_parse_csr.len > 0 && authority_parse_csr.ptr[0] == '[') {
+            port_search_start = memchr(authority_parse_csr.ptr, ']', authority_parse_csr.len);
+            if (!port_search_start) {
+                parser->state = ERROR;
+                aws_raise_error(AWS_ERROR_MALFORMED_INPUT_STRING);
+                return;
+            }
+        }
+
+        uint8_t *port_delim = memchr(port_search_start, ':', authority_parse_csr.len);
```
graebm, line 367: the length here isn't valid anymore
if:
`authority_parse_csr = "[1]:" // len=4`
then `port_search_start` = ":" but you're saying the length is 4

DmitriyMusatkin: fixed

graebm (approved):

PR #940, opened 2022-10-11
Json duplicate

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Add json duplicate api

### DIFF
--- a/source/json.c
+++ b/source/json.c
@@ -294,7 +294,8 @@ bool aws_json_value_compare(const struct aws_json_value *a, const struct aws_jso
 struct aws_json_value *aws_json_value_duplicate(const struct aws_json_value *value) {
     struct cJSON *cjson = (struct cJSON *)value;
     if (cJSON_IsInvalid(cjson)) {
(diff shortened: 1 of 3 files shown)


waahm7 (approved):

source/json.c
```diff
@@ -291,6 +291,16 @@ bool aws_json_value_compare(const struct aws_json_value *a, const struct aws_jso
...
 }
 
+struct aws_json_value *aws_json_value_duplicate(const struct aws_json_value *value) {
+    struct cJSON *cjson = (struct cJSON *)value;
+    if (cJSON_IsInvalid(cjson)) {
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        return NULL;
+    }
+
+    return (void *)cJSON_Duplicate(cjson, true);
```
waahm7, line 301: This function can also fail. I think we should do aws_raise_error if it fails.

DmitriyMusatkin: fair point. it returns null on failure. let me add a check for that

PR #943, opened 2022-10-19
clarify aws_byte_cursor_next_split comments and add tests

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
No functional changes. Doc and test updates


include/aws/common/byte_buf.h
```diff
@@ -229,16 +229,24 @@ AWS_COMMON_API
...
- * No copies, no buffer allocations. Iterates over input_str, and returns the next substring between split_on instances.
+ * No copies, no buffer allocations. Iterates over input_str, and returns the
+ * next substring between split_on instances relative to previous substr.
+ * Function assumes that input_str implicitly starts and ends with split_on
+ * char, so first substr will be from input_str start to first occurrence of
+ * split_on and last substr will be from last occurrence of split_on until the
+ * end of string.
+ *
+ * Returns true if substr has value and false if not (substr will be zeroed in
+ * that case).
```
graebm, line 240: "substr has value" is a bit ambiguous, kind of sounds like the substr.len will never be `len==0`

maybe something like:
> Returns true each time substr is set to a split. When false is returned, there are no further splits (substr will be zeroed out in that case). Every string will split at least once.

worth showing a code example?
```
struct aws_byte_cursor substr = {0};
while (aws_byte_cursor_next_split(&input_str, ';', &substr)) {
    // ...use substr...
}
```

tests/split_test.c
```diff
@@ -282,3 +282,50 @@ static int s_test_char_split_output_too_small_fn(struct aws_allocator *allocator
 
     return 0;
 }
+
+AWS_TEST_CASE(test_byte_cursor_next_split, s_test_byte_cursor_next_split)
+static int s_test_byte_cursor_next_split(struct aws_allocator *allocator, void *ctx) {
```
graebm, line 287: 😅 I thought we had explicit tests for this function

I guess we *implicitly* tested it since it's powering the other split functions.

This is a good addition

graebm (approved):

include/aws/common/byte_buf.h
```diff
@@ -229,16 +229,24 @@ AWS_COMMON_API
 bool aws_byte_buf_eq_c_str_ignore_case(const struct aws_byte_buf *const buf, const char *const c_str);
 
 /**
- * No copies, no buffer allocations. Iterates over input_str, and returns the next substring between split_on instances.
+ * No copies, no buffer allocations. Iterates over input_str, and returns the
+ * next substring between split_on instances relative to previous substr.
+ * Function assumes that input_str implicitly starts and ends with split_on
+ * char, so first substr will be from input_str start to first occurrence of
+ * split_on and last substr will be from last occurrence of split_on until the
+ * end of string.
```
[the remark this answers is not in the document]
DmitriyMusatkin: Yeah, it was not initially clear to me that its a strtok kinda function from reading the comments. I've reworded my comments a bit to make it clearer.

PR #969, opened 2023-01-11
Add getter for thread name

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Add getter for thread name


SergeyRyabinin (approved):

cmake/AwsThreadName.cmake
```diff
@@ -52,7 +54,18 @@ function(aws_set_thread_name_method target)
...
+    endif()
+
+    check_c_source_compiles("
+        ${c_source_start}
+        char name[16] = {0};
+        pthread_getname_np(thread_id, name, 16);
+        ${c_source_end}
+        " PTHREAD_HAS_GETNAME)
+    if (PTHREAD_HAS_GETNAME)
+        target_compile_definitions(${target} PRIVATE -DAWS_PTHREAD_HAS_GETNAME)
```
graebm, line 67: oh glorious, I googled pthread_setname_np and found docs pages for some less-common systems where it takes 2 args: [IBM i???](https://www.ibm.com/docs/en/i/7.2?topic=ssw_ibm_i_72/apis/pthread_getname_np.html) [MKS???](https://www.mkssoftware.com/docs/man3/pthread_getname_np.3.asp)

maybe stick with the _TAKES_3_ARGS naming scheme
your call if you want to support the _TAKES_2_ARGS case

include/aws/common/thread.h
```diff
@@ -237,6 +238,14 @@ AWS_COMMON_API void aws_thread_increment_unjoined_count(void);
  */
 AWS_COMMON_API void aws_thread_decrement_unjoined_count(void);
 
+/**
+ * Gets name of the current thread.
+ * Caller is responsible for destroying returned string.
+ * If thread does not have a name, NULL is returned.
+ * If call failed NULL is returned and aws error is raised.
+ */
+AWS_COMMON_API struct aws_string *aws_thread_current_name(struct aws_allocator *allocator);
```
graebm, line 247: trivial: if you want to disambiguate between "has no data" and "error occurred", then change the signature to:
`int aws_thread_current_name(struct aws_allocator *, struct aws_string **out_name);`

that would be consistent with our existing APIs

to my knowledge, we have no APIs that clear aws_last_error() and then ask the user to check it themselves to see if something went wrong

source/posix/thread.c
```diff
@@ -460,3 +460,16 @@ int aws_thread_current_at_exit(aws_thread_atexit_fn *callback, void *user_data)
     tl_wrapper->atexit = cb;
     return AWS_OP_SUCCESS;
 }
+
+struct aws_string *aws_thread_current_name(struct aws_allocator *allocator) {
+#ifdef AWS_PTHREAD_HAS_GETNAME
+    char name[16] = {0};
+    if (pthread_getname_np(aws_thread_current_thread_id(), name, 16)) {
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
```
graebm, line 468: there's no way OUR user could have passed an invalid arg to this function, so this error isn't helping them

I'd have AWS_ERROR_SYS_CALL_FAILURE be "we don't know what went wrong" error

source/posix/thread.c
```diff
@@ -460,3 +460,16 @@ int aws_thread_current_at_exit(aws_thread_atexit_fn *callback, void *user_data)
     tl_wrapper->atexit = cb;
     return AWS_OP_SUCCESS;
 }
+
+struct aws_string *aws_thread_current_name(struct aws_allocator *allocator) {
+#ifdef AWS_PTHREAD_HAS_GETNAME
+    char name[16] = {0};
```
graebm, line 466:
```suggestion
    char name[256] = {0};
```
pthread_getname_np() errors with ERANGE if the buffers not big enough for the whole name

16 is the size for Linux, but I know on Mac it can get way bigger, and I believe Mac gives the main thread a pretty long name, at least it looks long in XCode.

source/windows/thread.c
```diff
@@ -407,3 +412,19 @@ int aws_thread_current_at_exit(aws_thread_atexit_fn *callback, void *user_data)
...
+    if (s_GetThreadDescription) {
+
+        PWSTR wname = NULL;
+        if (SUCCEEDED(s_GetThreadDescription(GetCurrentThread(), &wname))) {
+            struct aws_string *name = aws_string_convert_from_wchar_c_str(allocator, wname);
+            LocalFree(wname);
+            return name;
+        }
+
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
```
graebm, line 426: Again, I'd do AWS_ERROR_SYS_CALL_FAILURE

source/posix/thread.c
```diff
@@ -460,3 +460,16 @@ int aws_thread_current_at_exit(aws_thread_atexit_fn *callback, void *user_data)
...
+    if (pthread_getname_np(aws_thread_current_thread_id(), name, 16)) {
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        return NULL;
+    }
+    return aws_string_new_from_c_str(allocator, name);
+#else
+    return NULL;
```
graebm, line 473: aws_raise_error(AWS_ERROR_PLATFORM_NOT_SUPPORTED) is the "Feature not supported on this platform" error

source/windows/thread.c
```diff
@@ -407,3 +412,19 @@ int aws_thread_current_at_exit(aws_thread_atexit_fn *callback, void *user_data)
...
+    }
+
+    return NULL;
```
graebm, line 429: similar, AWS_ERROR_PLATFORM_NOT_SUPPORTED

tests/thread_test.c
```diff
@@ -47,6 +50,12 @@ static int s_test_thread_creation_join_fn(struct aws_allocator *allocator, void
         aws_thread_get_detach_state(&thread),
         "thread state should have returned JOIN_COMPLETED");
 
+#ifdef AWS_PTHREAD_HAS_GETNAME
+    ASSERT_CURSOR_VALUE_STRING_EQUALS(
+        aws_byte_cursor_from_c_str("MyThreadName"), test_data.thread_name, "thread name equals");
+#endif
```
graebm, line 56: Instead of using the ifdef, you could just call get_name(), and assert that either:
- error is PLATFORM_NOT_SUPPORTED
- OR name is the same one we passed in

tests/thread_test.c
```diff
@@ -47,6 +50,12 @@ static int s_test_thread_creation_join_fn(struct aws_allocator *allocator, void
         aws_thread_get_detach_state(&thread),
         "thread state should have returned JOIN_COMPLETED");
 
+#ifdef AWS_PTHREAD_HAS_GETNAME
+    ASSERT_CURSOR_VALUE_STRING_EQUALS(
```
graebm, line 54: debatable: we could add a new test, explicitly for setname/getname

I'd tacked the setname stuff into this existing test because there was no way to actually test the results before, I just wanted the code path the get exercised, but now we CAN check the results

include/aws/common/thread.h
```diff
@@ -237,6 +238,14 @@ AWS_COMMON_API void aws_thread_increment_unjoined_count(void);
  */
 AWS_COMMON_API void aws_thread_decrement_unjoined_count(void);
 
+/**
+ * Gets name of the current thread.
```
graebm, line 242: While we're doing this, let's have the function take a `aws_thread_id_t` instead of only being for the current thread.

That's the API for every lower-level get-thread-name function we're calling into

You could still offer the _current_name function as a convenience thing, or not

cmake/AwsThreadName.cmake
```diff
@@ -41,7 +43,7 @@ function(aws_set_thread_name_method target)
         PTHREAD_SETNAME_TAKES_2ARGS)
     if (PTHREAD_SETNAME_TAKES_2ARGS)
         target_compile_definitions(${target} PRIVATE -DAWS_PTHREAD_SETNAME_TAKES_2ARGS)
-        return()
+        #return()
```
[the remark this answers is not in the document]
DmitriyMusatkin: Ugh, i completely forgot i commented it out while debugging things. will fix

include/aws/common/thread.h
DmitriyMusatkin: Im not opposed to expose thread id version as well, but there are a couple reasons i decided not to add it to pr:
1. we currently dont have apis that take thread id, but we have couple apis that work on current thread. use case for this is to get thread name in the callback, so i didnt want to overcomplicate
2. on windows we would need to open/close handle to use thread id. minor thing to implement, but seemed like overkill

include/aws/common/thread.h
DmitriyMusatkin: i've debated on this a little bit. i originally wrote it as int version, but i cant think of a case where user would care if its error or missing name. and passing out_name is a bit uglier to use. maybe we should just treat both cases the same and return NULL ptr.

I vaguely remember seeing this pattern in our code base (maybe in s3). But i do agree its ugly and probably should not be used here

cmake/AwsThreadName.cmake
DmitriyMusatkin: yeah. missed those 2. switched to 2/3 arg approach.

tests/thread_test.c
DmitriyMusatkin: its currently in a test that does not do much besides joining and checking thread id, so i decided to just keep it in the same test

graebm (approved):

PR #970, opened 2023-01-11
aws_fopen_safe() raises aws-error on failure

graebm:
**Issue:**
`aws_fopen_safe()` and `aws_fopen()` did not raise an aws-error-code on failure. It expected the caller to check `errno`. This is super inconsistent, none of our other functions expect the caller to check `errno`. Also, the windows implementation never set `errno` in any circumstance.

**Description of changes:**
* `aws_fopen_safe()` and `aws_fopen()` always raise an aws-error-code if something goes wrong.
* aws_logging_init_noalloc() checks if file open failed. It wasn't checking before.

### DIFF
--- a/source/file.c
+++ b/source/file.c
@@ -13,6 +13,7 @@
 
 FILE *aws_fopen(const char *file_path, const char *mode) {
     if (!file_path || strlen(file_path) == 0) {
+        AWS_LOGF_ERROR(AWS_LS_COMMON_IO, "static: Failed to open file. path is empty");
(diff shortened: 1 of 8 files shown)


source/file.c
```diff
@@ -13,7 +13,8 @@
 
 FILE *aws_fopen(const char *file_path, const char *mode) {
     if (!file_path || strlen(file_path) == 0) {
```
bretambrose, line 15: Do we want to check mode too?

bretambrose (approved):

PR #971, opened 2023-01-13
Always pull out value of errno right after sys call

bretambrose:
A pass through all uses of errno.  There were a number of places where a system call was made and then the errno check for it occurred after other things that might side-affect the value of errno (like logging).

This PR establishes a new pattern for system call error checking.  The following approaches are acceptable:
```
int result = system_call(...);
int errno_value = errno;
if (result) {
   ... // do stuff with errno_value
}
```

or if there's a concern with thread local access performance?:

```
int result = system_call(...);
if (result) {
   int errno_value = errno;
   ... // do stuff with errno_value
}
```

Applying this pattern universally sometimes leads to what looks like trivial separation:

```
int result = system_call(...);
if (result) {
   int errno_value = errno;
   return aws_raise_error(translate_io_error(errno_value));
}
```

but it's important to follow the pattern everywhere since skipping it in the above case makes it far easier for someone to
accidentally break things in the future by inserting logic between the system call and the aws_raise_error() call, ala:

```
int result = system_call(...);
if (result) {
   AWS_LOGF_ERROR(...);
   return aws_raise_error(translate_io_error(errno));
}
``` 

### DIFF
--- a/source/file.c
+++ b/source/file.c
@@ -41,7 +41,7 @@ int aws_byte_buf_init_from_file(struct aws_byte_buf *out_buf, struct aws_allocat
 
     if (fp) {
         if (fseek(fp, 0L, SEEK_END)) {
-            int errno_value = errno;
+            int errno_value = errno; /* Always cache errno before potential side-effect */
             AWS_LOGF_ERROR(AWS_LS_COMMON_IO, "static: Failed to seek file %s with errno %d", filename, errno_value);
             fclose(fp);
             return aws_translate_and_raise_io_error(errno_value);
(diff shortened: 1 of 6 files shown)


source/file.c
```diff
@@ -41,9 +41,10 @@ int aws_byte_buf_init_from_file(struct aws_byte_buf *out_buf, struct aws_allocat
 
     if (fp) {
         if (fseek(fp, 0L, SEEK_END)) {
-            AWS_LOGF_ERROR(AWS_LS_COMMON_IO, "static: Failed to seek file %s with errno %d", filename, errno);
+            int errno_value = errno;
```
graebm, line 44: trivial: annoying, but explain the subtle bug for any future programmers who come in here. I copy/pasted this everywhere when I did the windows GetLastError() pass
```suggestion
            int errno_value = errno; /* logging may reset error, so cache it */
```

graebm (approved):

Issue #968, opened 2023-01-10
Invalid _mulx_u32 CMake check


fweimer-rh:
This build-time check is invalid:

https://github.com/awslabs/aws-c-common/blob/7959065399a86be95073085be381eedfcbe93538/cmake/AwsFeatureTests.cmake#L36..L43

* GCC defines `_mulx_u32` in a header file so the test always fails.
* A build-time test for `_mulx_u32` does not tell us anything about run-time availability.

Found as part of:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

TingDaoK:
The issue has been resolved by https://github.com/awslabs/aws-c-common/pull/974

PR #974, opened 2023-01-19
use BMI2 cpuid instead for `_mulx_u32`

TingDaoK:
*Issue #, if available:*

- https://github.com/awslabs/aws-c-common/issues/968
- As the check always fails, we never ever compile `math.msvc.inl`
- It also uncovers bug of implementation there.

*Description of changes:*

- Instead of using build-time check, use runtime check with CPUID.
- According to [here](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mulx_u32&ig_expand=5244) CPUID Flags for `_mulx_u32` is BMI2
- According to [here](https://en.wikipedia.org/wiki/CPUID) BMI2 `CPUID.(EAX=07H, ECX=0H):EBX.BMI2[bit 8]==1`
- According to [here](https://learn.microsoft.com/en-us/cpp/intrinsics/emul-emulu?view=msvc-170), use `__emulu` instead when `_mulx_u32` is unavailable.
- Uncovers the bug on ctz and clz impl, fixed it and added tests

###NOTES:

- On 32 bit windows, following functions need to be replaced. We cannot find any good replacement, fallback on 32 bit windows as basically half of the optimization is not available.
```
D:\a\aws-c-common\aws-c-common\aws-c-common\include\aws/common/math.msvc.inl(26): warning C4013: '_umul128' undefined; assuming extern returning int [D:\a\aws-c-common\aws-c-common\aws-c-common\build\aws-c-common\aws-c-common.vcxproj]
D:\a\aws-c-common\aws-c-common\aws-c-common\include\aws/common/math.msvc.inl(87): warning C4013: '_addcarry_u64' undefined; assuming extern returning int [D:\a\aws-c-common\aws-c-common\aws-c-common\build\aws-c-common\aws-c-common.vcxproj]
D:\a\aws-c-common\aws-c-common\aws-c-common\include\aws/common/math.msvc.inl(147): warning C4013: '_BitScanReverse64' undefined; assuming extern returning int [D:\a\aws-c-common\aws-c-common\aws-c-common\build\aws-c-common\aws-c-common.vcxproj]
D:\a\aws-c-common\aws-c-common\aws-c-common\include\aws/common/math.msvc.inl(182): warning C4013: '_BitScanForward64' undefined; assuming extern returning int [D:\a\aws-c-common\aws-c-common\aws-c-common\build\aws-c-common\aws-c-common.vcxproj]
```
- For windows 2019 with MSVC 15, `_addcarry_u32` and `_addcarry_u64` has a weird bug that if the first number is 0, the returned carry out overflow flag will ALWAYS be set, which makes me Worried, as it will affect all our downstream and who knows what else weird bug is out there.


include/aws/common/math.msvc.inl
```diff
@@ -44,21 +44,36 @@ AWS_STATIC_IMPL int aws_mul_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
  * Multiplies a * b. If the result overflows, returns 2^32 - 1.
  */
 AWS_STATIC_IMPL uint32_t aws_mul_u32_saturating(uint32_t a, uint32_t b) {
-    uint32_t out;
-    uint32_t ret_val = _mulx_u32(a, b, &out);
-    return (out == 0) ? ret_val : UINT32_MAX;
+    if (aws_cpu_has_feature(AWS_CPU_FEATURE_BMI2)) {
+        uint32_t out;
+        uint32_t ret_val = _mulx_u32(a, b, &out);
```
DmitriyMusatkin, line 49: is out here just whether or not overflow occured?

include/aws/common/math.msvc.inl
```diff
@@ -44,21 +44,36 @@ AWS_STATIC_IMPL int aws_mul_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
...
+        return (out == 0) ? ret_val : UINT32_MAX;
+    } else {
+        /* If BMI2 unavailable, use __emulu instead */
+        uint64_t result = __emulu(a, b);
```
DmitriyMusatkin, line 53: is it guaranteed to be available everywhere?

DmitriyMusatkin: seems like its base x86 intrinsic judging by https://learn.microsoft.com/en-us/cpp/intrinsics/x64-amd64-intrinsics-list?view=msvc-160

tests/cpuid_test.c
```diff
@@ -24,6 +24,7 @@ static int s_cpuid_test_fn(struct aws_allocator *allocator, void *ctx) {
     aws_cpu_has_feature(AWS_CPU_FEATURE_SSE_4_2);
     aws_cpu_has_feature(AWS_CPU_FEATURE_AVX2);
     aws_cpu_has_feature(AWS_CPU_FEATURE_ARM_CRC);
```
DmitriyMusatkin, line 26: should we be asserting in this test in any way? how do we know it works?

include/aws/common/math.msvc.inl
TingDaoK: It stores the high 32 bit from [here](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mulx_u32&ig_expand=5244).

> Multiply unsigned 32-bit integers a and b, store the low 32-bits of the result in dst, and store the high 32-bits in hi

So, yeah, basically if it's non-zero, overflow happened.

tests/cpuid_test.c
TingDaoK: We cannot assert on it, it has different result on different platform...

Just run them and not crash?

DmitriyMusatkin: how do we know if we are actually picking up the feature on the platforms where its supported? wasnt the original issue because of incorrect check. 
Does not have to be addressed in this pr, but something we need to think about

include/aws/common/math.msvc.inl
DmitriyMusatkin: nit: but can we rename out to overflow or something?

include/aws/common/math.msvc.inl
DmitriyMusatkin: why intel only? it should be available on all processors that support x84 or x64 instruction sets

DmitriyMusatkin (approved):

include/aws/common/math.msvc.inl
```diff
@@ -67,8 +83,17 @@ AWS_STATIC_IMPL int aws_mul_u32_checked(uint32_t a, uint32_t b, uint32_t *r) {
  * If a + b overflows, returns AWS_OP_ERR; otherwise adds
  * a + b, returns the result in *r, and returns AWS_OP_SUCCESS.
  */
-AWS_STATIC_IMPL int aws_add_u64_checked(uint32_t a, uint32_t b, uint32_t *r) {
-    if (_addcarry_u64(0, a, b, *r)) {
+AWS_STATIC_IMPL int aws_add_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
+    uint8_t c_in = 0u;
+    if (a == 0) {
+        /* Fallback for any one is zero, as a weird bug happened if a=0, returned carry out will always be 1 at certain
+         * platform (MSVC 15) */
```
graebm, line 90: I have doubts that this is a compiler bug ???

I fed some similar code into Compiler Explorer and I get identical assembly from mvsc [2015](https://gcc.godbolt.org/z/8MMxsTMe3) and [2017](https://gcc.godbolt.org/z/oo7d79eGs)

include/aws/common/math.gcc_builtin.inl
```diff
@@ -48,18 +60,30 @@ AWS_STATIC_IMPL size_t aws_clz_size(size_t n) {
  * Search from the LSB to MSB, looking for a 1
  */
 AWS_STATIC_IMPL size_t aws_ctz_u32(uint32_t n) {
+    if (n == 0) {
+        return sizeof(n) * 8;
+    }
```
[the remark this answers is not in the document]
graebm: after thinking about this, I am wrong.

undefined behavior is bad. MS did the right thing, detecting it and returning true/false. But we emulated the GCC API, which has no way to report error. So yeah, agree this change is righteous

.github/workflows/ci.yml
```diff
@@ -113,6 +113,28 @@ jobs:
...
+    runs-on: windows-2019 # windows-2019 is last env with Visual Studio 2015 (v14.0)
+    strategy:
+      matrix:
+        arch: [x86, x64]
+    steps:
+    - name: Build ${{ env.PACKAGE_NAME }} + consumers
+      run: |
+        python -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder.pyz')"
+        python builder.pyz build -p ${{ env.PACKAGE_NAME }} --target windows-${{ matrix.arch }} --compiler msvc-15
+
```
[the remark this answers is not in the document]
TingDaoK: Yeah, it does.
> For windows 2019 with MSVC 15, _addcarry_u32 and _addcarry_u64 has a weird bug that if the first number is 0, the returned carry out overflow flag will ALWAYS be set, which makes me Worried, as it will affect all our downstream and who knows what else weird bug is out there.

include/aws/common/math.msvc.inl
TingDaoK: Hmmm, yeah, that's weird... 

But the two CI runs on the same type of image, the only difference is the compiler version. I don't know why...

include/aws/common/math.inl
```diff
@@ -27,7 +27,8 @@ AWS_EXTERN_C_BEGIN
 #    include <aws/common/math.gcc_x64_asm.inl>
 #elif defined(__aarch64__) && defined(AWS_HAVE_GCC_INLINE_ASM)
 #    include <aws/common/math.gcc_arm64_asm.inl>
-#elif defined(AWS_HAVE_MSVC_MULX)
+#elif defined(AWS_HAVE_MSVC_INTRINSICS_X64)
+/* If targeted at 32bit doesn't support 64bit intrinsics, and no replacement was implemented by MSVC, fallback */
```
graebm, line 31: trivial: comment seems unnecessary, the ifdef is clear enough
```suggestion
```

include/aws/common/math.msvc.inl
```diff
@@ -40,102 +41,176 @@ AWS_STATIC_IMPL int aws_mul_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
     return AWS_OP_SUCCESS;
 }
 
+static uint32_t (*s_mul_u32_saturating_fn_ptr)(uint32_t a, uint32_t b) = 0;
```
graebm, line 44: trivial: 
```suggestion
static uint32_t (*s_mul_u32_saturating_fn_ptr)(uint32_t a, uint32_t b) = NULL;
```

include/aws/common/math.msvc.inl
```diff
@@ -40,102 +41,176 @@ AWS_STATIC_IMPL int aws_mul_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
...
+            s_mul_u32_saturating_fn_ptr = s_mulx_u32_saturating;
+        } else {
+            /* If BMI2 unavailable, use __emulu instead */
+            s_mul_u32_saturating_fn_ptr = s_emulu_saturating;
+        }
+    }
+    return s_mul_u32_saturating_fn_ptr(a, b);
+}
+
+static int (*s_mul_u32_checked_fn_ptr)(uint32_t a, uint32_t b, uint32_t *r) = 0;
```
graebm, line 71: trivial:
```suggestion
static int (*s_mul_u32_checked_fn_ptr)(uint32_t a, uint32_t b, uint32_t *r) = NULL;
```

include/aws/common/math.msvc.inl
```diff
@@ -40,102 +41,176 @@ AWS_STATIC_IMPL int aws_mul_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
...
  * If a + b overflows, returns AWS_OP_ERR; otherwise adds
  * a + b, returns the result in *r, and returns AWS_OP_SUCCESS.
  */
-AWS_STATIC_IMPL int aws_add_u64_checked(uint32_t a, uint32_t b, uint32_t *r) {
-    if (_addcarry_u64(0, a, b, *r)) {
+AWS_STATIC_IMPL int aws_add_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
+#if !defined(_MSC_VER) || _MSC_VER < 1920
+    /* Fallback MSVC 2017 and older, _addcarry doesn't work correctly for those compiler */
+    if ((b > 0) && (a > (UINT64_MAX - b)))
+        return aws_raise_error(AWS_ERROR_OVERFLOW_DETECTED);
```
graebm, line 115: trivial:
```suggestion
    if ((b > 0) && (a > (UINT64_MAX - b))) {
        return aws_raise_error(AWS_ERROR_OVERFLOW_DETECTED);
    }
```

include/aws/common/math.msvc.inl
```diff
@@ -40,102 +41,176 @@ AWS_STATIC_IMPL int aws_mul_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
...
  */
-AWS_STATIC_IMPL uint64_t aws_add_u64_saturating(uint32_t a, uint32_t b) {
-    uint32_t res;
-
-    if (_addcarry_u64(0, a, b, &res)) {
+AWS_STATIC_IMPL uint64_t aws_add_u64_saturating(uint64_t a, uint64_t b) {
+#if !defined(_MSC_VER) || _MSC_VER < 1920
+    /* Fallback MSVC 2017 and older, _addcarry doesn't work correctly for those compiler */
+    if ((b > 0) && (a > (UINT64_MAX - b)))
+        return UINT64_MAX;
```
graebm, line 133: trivial:
```suggestion
    if ((b > 0) && (a > (UINT64_MAX - b))) {
        return UINT64_MAX;
    }
```

include/aws/common/math.msvc.inl
```diff
@@ -40,102 +41,176 @@ AWS_STATIC_IMPL int aws_mul_u64_checked(uint64_t a, uint64_t b, uint64_t *r) {
...
 /**
  * If a + b overflows, returns AWS_OP_ERR; otherwise adds
  * a + b, returns the result in *r, and returns AWS_OP_SUCCESS.
  */
 AWS_STATIC_IMPL int aws_add_u32_checked(uint32_t a, uint32_t b, uint32_t *r) {
-  if(_addcarry_u32(0, a, b, *r){
+#if !defined(_MSC_VER) || _MSC_VER < 1920
+    /* Fallback MSVC 2017 and older, _addcarry doesn't work correctly for those compiler */
+    if ((b > 0) && (a > (UINT32_MAX - b)))
+        return aws_raise_error(AWS_ERROR_OVERFLOW_DETECTED);
```
graebm, line 152:
```suggestion
    if ((b > 0) && (a > (UINT32_MAX - b))) {
        return aws_raise_error(AWS_ERROR_OVERFLOW_DETECTED);
    }
```

tests/CMakeLists.txt
```diff
@@ -511,19 +514,19 @@ add_test_case(test_uri_decode)
 
 generate_test_driver(${PROJECT_NAME}-tests)
 
-if (NOT MSVC AND NOT LEGACY_COMPILER_SUPPORT)
-    #we have some tests here that purposely overflow
+if(NOT MSVC AND NOT LEGACY_COMPILER_SUPPORT)
+    # we have some tests here that purposely overflow
```
graebm, line 518: what plugin is enforcing all this whitespace consistency?

I like it

.github/workflows/ci.yml
```diff
@@ -113,6 +113,28 @@ jobs:
         python -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder.pyz')"
         python builder.pyz build -p ${{ env.PACKAGE_NAME }}
 
+  windows-vc16:
+    runs-on: windows-2022 # latest
+    strategy:
+      matrix:
+        arch: [x86, x64]
```
graebm, line 120: I've look at all the build output and certify that these builds area actually doing what they claim. amazing

graebm: don't forget to mark these required when you merge

graebm (approved):
fix & ship

tests/CMakeLists.txt
TingDaoK: I don't think I used anything special for cmake file format, it's probably just the default one with VScode. But, I do use [cbrevik.toggle-format-on-save](https://marketplace.visualstudio.com/items?itemName=cbrevik.toggle-format-on-save) to automatic format files.

PR #978, opened 2023-01-25
Assert that each error-info-list starts with correct value

graebm:
**Issue:**
aws-c-iot is misreporting its errors, because we forgot to put the first error-code into the error-info-list.

This slipped past our existing checks on the correctness of error-info-lists.

**Description of changes:**
Check that the error-info-list starts with the correct value.


source/error.c
```diff
@@ -145,8 +145,20 @@ void aws_register_error_info(const struct aws_error_info_list *error_info) {
     }
 
 #if DEBUG_BUILD
+    /* Assert that first error has the right value */
+    const int expected_first_code = slot_index << AWS_ERROR_ENUM_STRIDE_BITS;
+    if (error_info->error_list[0].error_code != expected_first_code) {
```
sbSteveK, line 150: Solid check. 

Very Trivial:
Might want to specify in the log that the numbers are error codes for clarity. 
I wonder if there's a better way to indicate that the literal name being listed is associated with the second error code. It's clear what's going on looking at the code and with some context but could be slightly ambiguous at a glance. Maybe you could add (MISSING ERROR) or something after the first code. I dunno. Nitpicking here.

sbSteveK (approved):

PR #980, opened 2023-01-26
Added Utf8 validation callback to allow extra validation

xiazhvera:
*Issue #, if available:*

*Description of changes:*


source/encoding.c
```diff
@@ -499,6 +506,13 @@ int aws_utf8_validator_update(struct aws_utf8_validator *validator, struct aws_b
                 }
             }
         }
+
+        // Extra parse for caller defined validation callback.
```
bretambrose, line 510: Why wouldn't this go up in the if statement from 488-500?

bretambrose (commented):
Let's also add testing.

source/encoding.c
xiazhvera: Because it need also take care of the case for 1 byte codepoint (ln 459 - 463). Or I can add it to both place instead of here so we can reduce the branching.

include/aws/common/encoding.h
```diff
@@ -138,6 +138,14 @@ enum aws_text_encoding {
     AWS_TEXT_ASCII,
 };
 
+/**
+ * Callbacks to add another parse to customized validation for UTF-8 code point.
+ * @param codepoint the codepoint for validation. We assume that the codepoint
+ *                  would already be respecting RFC-3629.
+ * @return true if the codepoint is valid, otherwise false
+ */
+typedef bool(aws_utf8_customzied_validator_callback_fn)(const uint32_t codepoint);
```
bretambrose, line 147: 'customized'

can also drop the const on the parameter

include/aws/common/encoding.h
```diff
@@ -156,6 +164,16 @@ AWS_STATIC_IMPL bool aws_text_is_utf8(const uint8_t *bytes, size_t size);
...
 
+/**
+ * Scans every byte, and returns true if it is valid UTF8/ASCII as defined in RFC-3629 and
+ * user's customized validation callback. The function will return true for empty string
+ * regardless of the callback functions.
+ * The text does not need to begin with a UTF8 BOM.
+ */
+AWS_COMMON_API bool aws_text_is_valid_utf8_with_callback(
+    struct aws_byte_cursor bytes,
+    aws_utf8_customzied_validator_callback_fn callback);
```
bretambrose, line 175: I'd call this parameter validator maybe

Also, while it doesn't actually change anything underneath (functions are passed as pointers) I'd explicitly mark the function as a pointer.

include/aws/common/encoding.h
```diff
@@ -174,6 +192,15 @@ AWS_COMMON_API void aws_utf8_validator_reset(struct aws_utf8_validator *validato
  */
 AWS_COMMON_API int aws_utf8_validator_update(struct aws_utf8_validator *validator, struct aws_byte_cursor bytes);
 
+/**
+ * Update the validator with more bytes of text, using extra validation callback.
+ * Raises AWS_ERROR_INVALID_UTF8 if invalid UTF8 is encountered.
```
bretambrose, line 197: Do we want to raise a separate error if validation fails (but it's valid utf-8)?

include/aws/common/encoding.h
xiazhvera: "validator" is already used for the utf8 validator in the function body. I will use validator_fn instead.

tests/encoding_test.c
```diff
@@ -1368,7 +1368,7 @@ static int s_text_is_valid_utf8_callback(struct aws_allocator *allocator, void *
         } else {
             printf("valid example should still failed by the callback[%zu]: %s\n", i, example.name);
             ASSERT_FALSE(aws_text_is_valid_utf8_with_callback(example.text, s_utf8_validation_callback));
-            ASSERT_ERROR(AWS_ERROR_UTF8_EXTRA_VALIDATION_FAILED, aws_last_error());
+            ASSERT_TRUE(AWS_ERROR_UTF8_EXTRA_VALIDATION_FAILED == aws_last_error());
```
bretambrose, line 1371: You could also use ASSERT_INT_EQUALS

bretambrose (approved):

include/aws/common/encoding.h
```diff
@@ -138,6 +138,14 @@ enum aws_text_encoding {
     AWS_TEXT_ASCII,
 };
 
+/**
+ * Callbacks to add another parse to customized validation for UTF-8 code point.
+ * @param codepoint the codepoint for validation. We assume that the codepoint
+ *                  would already be respecting RFC-3629.
+ * @return true if the codepoint is valid, otherwise false
+ */
+typedef bool(aws_utf8_customized_validator_callback_fn)(uint32_t codepoint);
```
graebm, line 147:
```suggestion
/**
 * Callback invoked for each Unicode codepoint in a UTF-8 string.
 * Use this callback to store codepoints as they're decoded.
 * or to perform additional validation. RFC-3629 is already enforced,
 * which forbids codepoints between U+D800 and U+DFFF,
 * but you may need to forbid codepoints like U+0000.
 *
 * Return AWS_OP_SUCCESS to continue processing the string,
 * Raise an error (i.e. AWS_ERROR_INVALID_UTF8) to stop processing
 * the string and report failure.
 */
typedef int(aws_utf8_on_codepoint_fn)(uint32_t codepoint, void *user_data);
```

1) All callbacks must take a `void *`. If you leave this out, the callback can't meet a lot of use cases (i.e. calling back into another language)
2) The `void *` also opens this up to being used as a "decoder" if anyone ever needed to store the codepoints as their decoded.
3) let's return `int` instead of `bool` so someone can report their own special errors if they need to (like, if they're decoding a string, and it passes some max length, they'd want to raise a specific error about that, not just INVALID_UTF8)

include/aws/common/encoding.h
```diff
@@ -156,6 +164,16 @@ AWS_STATIC_IMPL bool aws_text_is_utf8(const uint8_t *bytes, size_t size);
...
 
+/**
+ * Scans every byte, and returns true if it is valid UTF8/ASCII as defined in RFC-3629 and
+ * user's customized validation callback. The function will return true for empty string
+ * regardless of the callback functions.
+ * The text does not need to begin with a UTF8 BOM.
+ */
+AWS_COMMON_API bool aws_text_is_valid_utf8_with_callback(
+    struct aws_byte_cursor bytes,
+    aws_utf8_customized_validator_callback_fn *validator_fn);
```
graebm, line 175: Let's just make this a general purpose "decode" function, instead of a "validate" function.

`int aws_decode_utf8(struct aws_byte_cursor bytes, aws_on_utf8_codepoint_fn *on_codepoint, void *user_data);`

That makes more sense, people can use it to check length, or validate codepoints, or store codepoints... whatever they want!

include/aws/common/encoding.h
```diff
@@ -171,9 +189,28 @@ AWS_COMMON_API void aws_utf8_validator_reset(struct aws_utf8_validator *validato
...
+ *
+ * Please note, there could be bytes remaining in the validator. In that case the
+ * update function will not report the invalidation and waiting for more input.
+ * You can manually check the validator->remaining or calling aws_utf8_validator_finalize
+ * to make sure there are no bytes left in the validator.
+ */
+AWS_COMMON_API int aws_utf8_validator_update_with_callback(
+    struct aws_utf8_validator *validator,
+    struct aws_byte_cursor bytes,
+    aws_utf8_customized_validator_callback_fn *validator_fn);
```
graebm, line 212: I prefer all options to be passed when something is created.

maybe aws_utf8_validator_new_with_callback(...) instead of an alternate update function

include/aws/common/error.h
```diff
@@ -198,6 +198,7 @@ enum aws_common_error {
     AWS_ERROR_DIRECTORY_NOT_EMPTY,
     AWS_ERROR_PLATFORM_NOT_SUPPORTED,
     AWS_ERROR_INVALID_UTF8,
+    AWS_ERROR_UTF8_EXTRA_VALIDATION_FAILED,
```
graebm, line 201: don't need this if we just let the callback raise whatever error it wants

MQTT could reuse the AWS_ERROR_INVALID_UTF8, because that still makes sense. Or it could raise its own special error with "MQTT5" in the name. Or it could raise an error about string being too long, or whatever!

graebm (commented):
I realized, when I wrote the validator, that it could also serve as a "decoder" if I added a callback to it. But I didn't, to keep it simple. But now we're adding that complexity! So let's just go all the way and support the decoding use case.

PR #983, opened 2023-02-15
Add aws_min_non_0_64()

sbSteveK:
helper function aws_min_non_0_64 used in multiple places moved to common.

### DIFF
--- a/include/aws/common/math.h
+++ b/include/aws/common/math.h
@@ -186,6 +186,7 @@ AWS_STATIC_IMPL uint32_t aws_max_u32(uint32_t a, uint32_t b);
 AWS_STATIC_IMPL int32_t aws_min_i32(int32_t a, int32_t b);
 AWS_STATIC_IMPL int32_t aws_max_i32(int32_t a, int32_t b);
 AWS_STATIC_IMPL uint64_t aws_min_u64(uint64_t a, uint64_t b);
+AWS_STATIC_IMPL uint64_t aws_min_non_0_u64(uint64_t a, uint64_t b);
 AWS_STATIC_IMPL uint64_t aws_max_u64(uint64_t a, uint64_t b);
 AWS_STATIC_IMPL int64_t aws_min_i64(int64_t a, int64_t b);
 AWS_STATIC_IMPL int64_t aws_max_i64(int64_t a, int64_t b);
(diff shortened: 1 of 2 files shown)


include/aws/common/math.h
```diff
@@ -186,6 +186,7 @@ AWS_STATIC_IMPL uint32_t aws_max_u32(uint32_t a, uint32_t b);
 AWS_STATIC_IMPL int32_t aws_min_i32(int32_t a, int32_t b);
 AWS_STATIC_IMPL int32_t aws_max_i32(int32_t a, int32_t b);
 AWS_STATIC_IMPL uint64_t aws_min_u64(uint64_t a, uint64_t b);
+AWS_STATIC_IMPL uint64_t aws_min_non_0_u64(uint64_t a, uint64_t b);
```
graebm, line 189: something like [clamp()](https://en.cppreference.com/w/cpp/algorithm/clamp) seems more generically useful?

```
uint64_t aws_clamp_u64(uint64_t val, uint64_t minimum, uint64_t maximum);
```

using it would look like:
```
next_service_time = aws_clamp_u64(operation_processing_time, 1, next_service_time);
```
which is equivalent to the proposed:
```
next_service_time = aws_min_non_0_64(operation_processing_time, next_service_time);
```

bretambrose: They're not equivalent

bretambrose: For that matter, maybe we need a serious name change that reflects that null/none property.

bretambrose: This all comes from the mqtt5 client architecture for composing next service time calculations. 0 means no-affect, non-zero means a time constraint.  We want the smallest non-zero time constraint as our next scheduled time.

graebm: maybe this should just stay an MQTT5 helper function, it's kinda weird as a generic math operation.
Looking at this header file, which is just a function name with no docs, assumed it meant "min, but higher than 0".

bretambrose: We're using it in secure tunneling too now.

bretambrose: aws_min_u64_nullable might be closer to a name improvement.

graebm: OK. But add docs to the header file explaining what it does

PR #985, opened 2023-02-16
Add CBMC proof-running GitHub Action

karkhaz:
This commit adds a GitHub Action that runs the CBMC proofs upon every push and pull request. This is intended to replace the current CBMC CI.

### DIFF
--- a/.github/workflows/proof_ci.yaml
+++ b/.github/workflows/proof_ci.yaml
@@ -1,0 +1,188 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+# CBMC starter kit 2.9
+name: Run CBMC proofs
+on:
+  push:
+    branches-ignore:
+      - gh-pages
+  pull_request:
+    branches-ignore:
+      - gh-pages
+  workflow_dispatch:
+
+# USAGE
+#
+# If you need to use different versions for tools like CBMC, modify this file:
+# .github/workflows/proof_ci_resources/config.yaml
+#
+# If you want the CI to use a different GitHub-hosted runner (which must still
+# be running Ubuntu 20.04), modify the value of this key:
+# jobs.run_cbmc_proofs.runs-on
+
+jobs:
+  run_cbmc_proofs:
+    runs-on: cbmc_ubuntu-latest_64-core
+    name: run_cbmc_proofs
+    permissions:
+      contents: read
+      id-token: write
+      pull-requests: read
+    steps:
+      - name: Check out repository and submodules recursively
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+      - name: Parse config file
+        run: |
+          CONFIG_FILE='.github/workflows/proof_ci_resources/config.yaml'
+          for setting in cadical-tag cbmc-version cbmc-viewer-version kissat-tag litani-version proofs-dir run-cbmc-proofs-command; do
+            VAR=$(echo $setting | tr "[:lower:]" "[:upper:]" | tr - _)
+            echo "${VAR}"=$(yq .$setting $CONFIG_FILE) >> $GITHUB_ENV
+          done
+      - name: Ensure CBMC, CBMC viewer, Litani versions have been specified
+        shell: bash
+        run: |
+          should_exit=false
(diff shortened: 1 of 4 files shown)


.github/workflows/proof_ci.yaml
```diff
@@ -0,0 +1,188 @@
...
+# SPDX-License-Identifier: MIT-0
+# CBMC starter kit 2.9
+name: Run CBMC proofs
+on:
+  push:
+    branches-ignore:
+      - gh-pages
+  pull_request:
+    branches-ignore:
+      - gh-pages
```
graebm, line 11: trivial: to my knowledge, the common-runtime team doesn't have a `gh-pages` branch in any of our repos
but if you just want this here for consistency with how it's set up in other repos, that's fine too

.github/workflows/proof_ci.yaml
```diff
@@ -0,0 +1,188 @@
...
+# If you need to use different versions for tools like CBMC, modify this file:
+# .github/workflows/proof_ci_resources/config.yaml
+#
+# If you want the CI to use a different GitHub-hosted runner (which must still
+# be running Ubuntu 20.04), modify the value of this key:
+# jobs.run_cbmc_proofs.runs-on
+
+jobs:
+  run_cbmc_proofs:
+    runs-on: cbmc_ubuntu-latest_64-core
```
graebm, line 25: just curious, where does this customer runner `cbmc_ubuntu-latest_64-core` come from?

.github/workflows/proof_ci.yaml
```diff
@@ -0,0 +1,188 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+# CBMC starter kit 2.9
+name: Run CBMC proofs
```
graebm, line 4: Should I remove the old "CBMC Proofs" Codebuild jobs once this is merged?

graebm (approved):

.github/workflows/proof_ci.yaml
karkhaz: We do have a [central definition](https://github.com/model-checking/cbmc-starter-kit/blob/master/src/cbmc_starter_kit/template-for-ci-workflow/proof_ci.yaml) of this file that would be copied into this repository whenever we update the starter kit, so I suggest we keep this here for idempocy

.github/workflows/proof_ci.yaml
karkhaz: This is a [Large Runner](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners) pool that we have access to. This pool is shared with all other CBMC-using repositories in the awslabs GitHub organisation and has a certain concurrency limit. I'll ping you offline with more details.

.github/workflows/proof_ci.yaml
karkhaz: I suggest we keep them running in parallel for a little while to bake and ensure that the new one works smoothly. I'll be working on migrating the data from the old accounts. Once that's done, I can ping you to set the new check as "Required" and remove the old webhook.

karkhaz: To clarify, by "a little while" I expect ~3 weeks or so to do the data transfer.

.github/workflows/proof_ci_resources/config.yaml
```diff
@@ -0,0 +1,7 @@
+cadical-tag: latest
+cbmc-version: latest
+cbmc-viewer-version: latest
+kissat-tag: latest
+litani-version: latest
+proofs-dir: verification/cbmc/proofs
+run-cbmc-proofs-command: ./run-cbmc-proofs.py
```
feliperodri, line 7: starter-kit is not part of this list, so I assume we will continue updating it as we always do: submitting a PR after applying the script the update it. Is that correct?

feliperodri (approved):
Thank you @karkhaz!

.github/workflows/proof_ci_resources/config.yaml
karkhaz: This file is considered part of the starter kit, it's in the same repo :) so whenever we update the starter kit, the relevant files will change anyway

PR #998, opened 2023-03-16
Fix warnings in public headers

graebm:
**Issue #:** https://github.com/awslabs/aws-crt-cpp/issues/330

Users should be able to crank warnings up to max in their own applications. But they can't easily do that if our public headers generate warnings.

(we do not promise to fix pedantic warnings in our .c code, only our public headers)

**Description of Changes:**
- Update header checker to error on any warning, and use pedantic warning levels
- new AWS_PUSH_SANE_WARNING_LEVEL/AWS_POP_SANE_WARNING_LEVEL macros let us disable ridiculous warnings within our own headers
    - Add these to start and end of every public header.


cmake/AwsCheckHeaders.cmake
```diff
@@ -40,20 +40,29 @@ function(aws_check_headers target)
             LINKER_LANGUAGE CXX
             CXX_STANDARD 11
             CXX_STANDARD_REQUIRED 0
-            C_STANDARD 90
+            C_STANDARD 99
```
JonathanHenson, line 43: If these are for cpp c99 is too modern right?

include/aws/common/array_list.h
```diff
@@ -10,6 +10,8 @@
 
 #include <stdlib.h>
 
+AWS_EXTERN_C_BEGIN
```
JonathanHenson, line 13: Won’t this export symbols that don’t need to be exported?

include/aws/common/atomics.h
```diff
@@ -8,6 +8,8 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
+AWS_EXTERN_C_BEGIN
```
JonathanHenson, line 11: Why are we exporting above the functions?

include/aws/common/array_list.h
graebm: yeah, after reading the docs I'm still not sure.

I'll run tests to see if this results in any more symbols being exposed...

cmake/AwsCheckHeaders.cmake
graebm: When it's C_STANDARD 90 and warnings are pedantic, we get slammed with stuff like "variadic macros are a C99 feature" and "bool is a c99 feature"

Remember how we separately generate a .cpp file and a .c file? So we stress including the header from C and we stress including it from C++. The .cpp files are passing so it seems ok?

include/aws/common/atomics.h
graebm: resolving this. we're having this discussion on another comment...

include/aws/common/array_list.h
DmitriyMusatkin: extern just tells linker that symbol will be available later somewhere else. and extern C does the same thing, but tells c++ compiler not to mangle the symbols it is looking for.

From what i remember from c spec, function declarations have implicit extern on them, so extern c would only change mangling on them.
But global vars might change semantic a bit if they are now under extern. ex if you declared int in header before without extern each compilation unit would get its own copy, but with extern the variable will be shared between all compilation units. But i dont think we ever use variables in this way. 

extern itself does not affect visibility of the symbol in the final dynamic lib

cmake/AwsCheckHeaders.cmake
JonathanHenson: I forgot that the simple act of using a c++ compiler will be sufficient to check those headers. So I'm fine with this.

include/aws/common/array_list.h
graebm: I ended up leaving the AWS_EXTERN_C_BEGIN where they were

include/aws/common/array_list.inl
```diff
@@ -96,18 +96,6 @@ bool aws_array_list_is_valid(const struct aws_array_list *AWS_RESTRICT list) {
     return required_size_is_valid && current_size_is_valid && data_is_valid && item_size_is_valid;
 }
 
-AWS_STATIC_IMPL
-void aws_array_list_debug_print(const struct aws_array_list *list) {
```
graebm, line 100: this function was unused, so I removed it

include/aws/common/array_list.inl
```diff
@@ -404,15 +392,6 @@ int aws_array_list_set_at(struct aws_array_list *AWS_RESTRICT list, const void *
     return AWS_OP_SUCCESS;
 }
 
-AWS_STATIC_IMPL
-void aws_array_list_sort(struct aws_array_list *AWS_RESTRICT list, aws_array_list_comparator_fn *compare_fn) {
-    AWS_PRECONDITION(aws_array_list_is_valid(list));
-    if (list->data) {
-        qsort(list->data, aws_array_list_length(list), list->item_size, compare_fn);
```
graebm, line 411: moved this implementation to another file: `.inl` -> `.c`
and changed it so it wasn't static/inline anymore
it was causing warnings because it called qsort() and couldn't be inlined

include/aws/testing/aws_test_harness.h
```diff
@@ -94,11 +98,11 @@ static int total_skip;
...
         return SUCCESS;                                                                                                \
     } while (0)
-#define PRINT_FAIL_INTERNAL(...)                                                                                       \
-    CUNIT_FAILURE_MESSAGE(__FUNCTION__, __FILE__, __LINE__, ##__VA_ARGS__, (const char *)NULL)
+
+#define PRINT_FAIL_INTERNAL(...) CUNIT_FAILURE_MESSAGE(__func__, __FILE__, __LINE__, ##__VA_ARGS__, (const char *)NULL)
 
 #define PRINT_FAIL_INTERNAL0(...)                                                                                      \
-    s_cunit_failure_message0(FAIL_PREFIX, __FUNCTION__, __FILE__, __LINE__, ##__VA_ARGS__, (const char *)NULL)
+    s_cunit_failure_message0(FAIL_PREFIX, __func__, __FILE__, __LINE__, ##__VA_ARGS__, (const char *)NULL)
```
graebm, line 105: `__FUNCTION__` is an extension, but `__func__` is official c99

.github/workflows/ci.yml
```diff
@@ -6,9 +6,9 @@ on:
       - 'main'
 
 env:
-  BUILDER_VERSION: v0.9.40
```
DmitriyMusatkin, line 9: dont forget to revert builder before merging

DmitriyMusatkin (approved):

waahm7 (commented):
We are missing promise.h, package.h, platform.h, stdbool.h, predicates.h, and stdint.h. The script is missing files that don’t contain extern C. `aws_promise.h` header should contain extern C headers as well I think.

waahm7 (approved):

PR #1005, opened 2023-03-28
Add function for writing N bytes of randomness

graebm:
**Issue:**
Currently, you need to [jump through hoops](https://github.com/awslabs/aws-c-cal/pull/136/files#diff-a8020fba42739fe7951879a98656c225b7d153f35c08a7f97af037146782c0bcR105-R113) if you only want to write N bytes to a buffer, because our existing `aws_device_random_buffer(aws_byte_buf *dst)` function simply fills the buffer to capacity.

**Description of Changes**:
- New function for appending N bytes `aws_device_random_buffer_append(aws_byte_buf *dst, size_t n)`
- Fix bug on windows when writing more than UINT32_MAX bytes
- Prevent errors when writing large buffers on Posix


source/posix/device_random.c
```diff
@@ -35,23 +35,26 @@ static void s_init_rand(void *user_data) {
     }
 }
 
-static int s_fallback_device_random_buffer(struct aws_byte_buf *output) {
+int aws_device_random_buffer_append(struct aws_byte_buf *output, size_t n) {
```
TingDaoK, line 38: should we add a `AWS_POSTCONDITION(aws_byte_buf_is_valid(output));`

TingDaoK (approved):

tests/device_random_test.c
```diff
@@ -130,3 +130,47 @@ static int s_device_rand_buffer_distribution_fn(struct aws_allocator *allocator,
...
 AWS_TEST_CASE(device_rand_buffer_distribution, s_device_rand_buffer_distribution_fn)
+
+static int s_device_rand_buffer_append_distribution_fn(struct aws_allocator *allocator, void *ctx) {
+    (void)allocator;
+    (void)ctx;
+
+    /* Create array full of zeroes, but only partially fill it with randomness */
+    uint8_t array[DISTRIBUTION_PUT_COUNT + 100] = {0};
+    struct aws_byte_buf buf = aws_byte_buf_from_empty_array(array, sizeof(array));
+    ASSERT_SUCCESS(aws_device_random_buffer_append(&buf, DISTRIBUTION_PUT_COUNT));
```
waahm7, line 141: trivial: maybe we should add a test for n > UINT32_MAX

waahm7 (approved):

tests/device_random_test.c
graebm: nice, this unveiled some failure
not a BUG per-se, because we'd properly report the failure
but on my x64 Mac, the old code would fail if size exceeded INT32_MAX
probably other POSIX platforms would fail similarly
I changed it work work on the data in smaller chunks

graebm: aaaaand I took the test out.
to push the boundaries it required 64bit machines to make a 4+GB allocation
that would make it impossible to run the unit tests on something like a raspberry pi 3

PR #1012, opened 2023-04-07
Add a way to skip a test

xiazhvera:
*Issue #, if available:*

*Description of changes:*
The test case can be skipped if it returned AWS_OP_SKIP. 

### DIFF
--- a/cmake/AwsTestHarness.cmake
+++ b/cmake/AwsTestHarness.cmake
@@ -10,7 +10,8 @@ option(ENABLE_NET_TESTS "Run tests requiring an internet connection." ON)
 define_property(GLOBAL PROPERTY AWS_TEST_CASES BRIEF_DOCS "Test Cases" FULL_DOCS "Test Cases")
 set(AWS_TEST_CASES "" CACHE INTERNAL "Test cases valid for this configuration")
 
-# The return value for the skipped test cases. Refer to the return code defined in aws_test_harness.h
+# The return value for the skipped test cases. Refer to the return code defined in aws_test_harness.h:
+# #define SKIP (1)
 set(SKIP_RETURN_CODE_VALUE 1)
 
 # Registers a test case by name (the first argument to the AWS_TEST_CASE macro in aws_test_harness.h)
--- a/cmake/AwsTestHarness.cmake
+++ b/cmake/AwsTestHarness.cmake
@@ -10,6 +10,9 @@ option(ENABLE_NET_TESTS "Run tests requiring an internet connection." ON)
 define_property(GLOBAL PROPERTY AWS_TEST_CASES BRIEF_DOCS "Test Cases" FULL_DOCS "Test Cases")
 set(AWS_TEST_CASES "" CACHE INTERNAL "Test cases valid for this configuration")
 
+# The return value for the skipped test cases. Refer to the return code defined in aws_test_harness.h
+set(SKIP_RETURN_CODE_VALUE 1)
+
 # Registers a test case by name (the first argument to the AWS_TEST_CASE macro in aws_test_harness.h)
 macro(add_test_case name)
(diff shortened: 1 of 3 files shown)


include/aws/testing/aws_test_harness.h
```diff
@@ -62,9 +62,24 @@ static int s_cunit_failure_message0(
     s_cunit_failure_message0(FAIL_PREFIX, func, file, line, format, #__VA_ARGS__)
 
 static int total_failures;
+static int total_skip;
```
xiazhvera, line 65: The value reserved to record the total skipped tests count. Current not in use.

include/aws/common/error.h
```diff
@@ -14,6 +14,7 @@
 
 #define AWS_OP_SUCCESS (0)
 #define AWS_OP_ERR (-1)
+#define AWS_OP_SKIP (1)
```
DmitriyMusatkin, line 17: probably not the best place to define a test only return code. can we put it in some test related header?

include/aws/testing/aws_test_harness.h
```diff
@@ -434,7 +449,7 @@ static inline int s_aws_run_test_case(struct aws_test_harness *harness) {
         test_res |= harness->on_after(allocator, setup_res, harness->ctx);
     }
 
-    if (test_res != AWS_OP_SUCCESS) {
+    if (test_res == AWS_OP_ERR) {
```
DmitriyMusatkin, line 452: should we define anything outside success and skip as error? its corner case, but if test returns something weird outside the defined constants we should probably fail

cmake/AwsTestHarness.cmake
```diff
@@ -87,6 +90,7 @@ function(generate_cpp_test_driver driver_exe_name)
 
     foreach(name IN LISTS TEST_CASES)
         add_test(${name} ${driver_exe_name} "${name}")
+        set_tests_properties("${name}" PROPERTIES SKIP_RETURN_CODE ${SKIP_RETURN_CODE_VALUE})
```
DmitriyMusatkin, line 93: do we need this if harness is checking return code of the test point?

xiazhvera: Yes. The ctest still need to know which error code is defined as "skipped" from the harness result.

include/aws/common/error.h
xiazhvera: I can move the AWS_OP_SKIP into aws_test_harness.h. However, `AWS_OP_SUCCESS` and `AWS_OP_ERR ` has been used in all the current tests, it makes more sense to have them together. 
Ideally, we should define test specified codes for succes/err/skip in aws_test_harness.h. However, that will be a significant API change that affects all aws-c tests, I dont know if we want to do it now?

DmitriyMusatkin: AWS_OP_SUCCESS and AWS_OP_ERR are also used extensively outside of tests. My concern is we are adding a test specific constant into a common header. In my opinion we should not be polluting this header with test only stuff and putting just the skip macro in test harness is ok as compromise, but we can have a team wide discussion on this.

include/aws/testing/aws_test_harness.h
```diff
@@ -62,9 +67,24 @@ static int s_cunit_failure_message0(
     s_cunit_failure_message0(FAIL_PREFIX, func, file, line, format, #__VA_ARGS__)
 
 static int total_failures;
+static int total_skip;
 
 #define SUCCESS (0)
 #define FAILURE (-1)
+#define SKIP (1)
```
xiazhvera, line 74: The skip code returned to cmake has to be in range [0,255]. Refer to cmake doc: https://cmake.org/cmake/help/latest/prop_test/SKIP_RETURN_CODE.html

DmitriyMusatkin (approved):

PR #1019, opened 2023-04-24
Tolerate unintialized list in aws_array_list_length

waahm7:
*Issue #, if available:*

*Description of changes:*
- Return 0 if the array list is not initilized. 
Follow up of https://github.com/awslabs/aws-c-io/pull/567#discussion_r1174203785

### DIFF
--- a/include/aws/testing/aws_test_harness.h
+++ b/include/aws/testing/aws_test_harness.h
@@ -404,7 +404,7 @@ static inline int s_aws_run_test_case(struct aws_test_harness *harness) {
     AWS_ASSERT(harness->run);
 /*
  * MSVC compiler has a weird interactive pop-up in debug whenever 'abort()' is called, which can be triggered
- * by hitting any aws_assert or aws_pre_condition, causing the CI to hang. So disable the pop up in tests.
+ * by hitting any aws_assert or aws_pre_condition, causing the CI to hang. So disable the pop-up in tests.
  */
 #ifdef _MSC_VER
     _set_abort_behavior(0, _WRITE_ABORT_MSG | _CALL_REPORTFAULT);
--- a/include/aws/testing/aws_test_harness.h
+++ b/include/aws/testing/aws_test_harness.h
@@ -404,6 +404,9 @@ static inline int s_aws_run_test_case(struct aws_test_harness *harness) {
     AWS_ASSERT(harness->run);
 
 #if defined(_WIN32)
+#    if defined(AWS_WIN_ABORT_POP_UP_DISABLE)
+    _set_abort_behavior(0, _WRITE_ABORT_MSG | _CALL_REPORTFAULT);
+#    endif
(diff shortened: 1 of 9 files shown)


include/aws/testing/aws_test_harness.h
```diff
@@ -404,6 +404,9 @@ static inline int s_aws_run_test_case(struct aws_test_harness *harness) {
     AWS_ASSERT(harness->run);
 
 #if defined(_WIN32)
+#    if defined(AWS_WIN_ABORT_POP_UP_DISABLE)
+    _set_abort_behavior(0, _WRITE_ABORT_MSG | _CALL_REPORTFAULT);
+#    endif
```
graebm, line 409:
```suggestion
    _set_abort_behavior(0, _WRITE_ABORT_MSG | _CALL_REPORTFAULT);
```

I vote that we simply always do this.

Get rid of the AWS_WIN_ABORT_POP_UP_DISABLE cmake option, and cmakedefine, and --cmake-extra, and comment explaining why we need it in every ci.yml...

If we always set it, you'll still break in a debugger, if you have a debugger attached, see the implementation of [aws_fatal_assert()](https://github.com/awslabs/aws-c-common/blob/b0fe4ee08f754784cf39918e53f922bbd751941b/source/assert.c#L14), which calls `aws_debug_break()` which breaks into debugger if a debugger is attached, before calling abort().

graebm: The only advantage of Windows' weird abort() popup is that it helps you break into a debugger. But no other platform does this, which is why we wrote our own logic to break into the debugger, which makes the popup totally useless for our tests...

always suppressing it will make tests on windows work like they do on all our other platforms

include/aws/testing/aws_test_harness.h
```diff
@@ -402,6 +402,13 @@ static void s_print_stack_trace(int sig, siginfo_t *sig_info, void *user_data) {
 
 static inline int s_aws_run_test_case(struct aws_test_harness *harness) {
     AWS_ASSERT(harness->run);
+/*
+ * MSVC compiler has a weird interactive pop-up in debug whenever 'abort()' is called, which can be triggered
+ * by hitting any aws_assert or aws_pre_condition, causing the CI to hang. So disable the pop up in tests.
+ */
+#ifdef _MSC_VER
+    _set_abort_behavior(0, _WRITE_ABORT_MSG | _CALL_REPORTFAULT);
+#endif
```
waahm7, line 411: During my local testing, I found out that `MinGW` fails to link with `_set_abort_behavior` and doesn't have the pop-up behavior.

include/aws/testing/aws_test_harness.h
waahm7: Thanks, updated.

graebm (approved):

PR #1040, opened 2023-07-06
New API to wrap an initialized array with an array list

bretambrose:

### DIFF
--- a/include/aws/common/array_list.h
+++ b/include/aws/common/array_list.h
@@ -63,7 +63,9 @@ void aws_array_list_init_static(
  * Initializes an array list with a preallocated array of *already-initialized* elements. item_count is the number of
  * elements in the array, and item_size is the size in bytes of each element.
  *
- * Primary use is to be able to treat an already initialized C array as an array list.
+ * Once initialized, nothing further can be added to the list, since it will be full and cannot resize.
+ *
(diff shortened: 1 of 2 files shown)


include/aws/common/array_list.h
```diff
@@ -59,6 +59,19 @@ void aws_array_list_init_static(
     size_t item_count,
     size_t item_size);
 
+/**
+ * Initializes an array list with a preallocated array of *already-initialized* elements. item_count is the number of
+ * elements in the array, and item_size is the size in bytes of each element.
```
graebm, line 64: Add a comment about how no further items can be added

include/aws/common/array_list.h
```diff
@@ -59,6 +59,19 @@ void aws_array_list_init_static(
...
+ *
+ * Primary use is to be able to treat an already initialized C array as an array list.
```
graebm, line 66: seriously: would it make more sense to just use the C array as it is? what benefit do you get by forcing people to use the array-list API to access the data? It checks your bounds, but we seldom write code anymore that checks whether of the functions returned an error

bretambrose: It's not my choice. The mqtt311 API throws around array lists in places and it's useful in tests.

graebm (approved):

PR #1043, opened 2023-07-13
Simplify error handling in XML API - BREAKING CHANGE

graebm:
**Issue:**
It's hard to report errors with the current API. Errors are being accidentally ignored, and some errors are never checked (perhaps because it was too much effort?).

**Diagnosis:**
The current callback returns `bool` of whether to continue parsing, rather than our typical `int/AWS_OP_SUCCESS/aws_raise_error()` [error handling](https://github.com/awslabs/aws-c-common/blob/b1ef77c1790a8776b58ae5ae57f90e9534b78991/README.md#error-handling).

This seems like a simple design. But the inconsistency in return type leads to errors being [mistakenly swallowed](https://github.com/awslabs/aws-c-s3/pull/330#discussion_r1256612397). And it makes it hard when you do want to "bubble up" an error from the callback. Callbacks needs to store a custom `error_code` in their `user_data` to report an error. Most callbacks never bothered to do this, maybe because it was extra work?

**Description of changes:**
- XML traversal callback returns `int` instead of `bool`.
    - If a callback fails, the whole parse() fails.
    - You can no longer stop parsing without causing failure. But in nearly all use-cases we were stopping due to error. I found 1 case where we stopped because we found what we were looking for. But in this case, it didn't really hurt to continue parsing. The reduced complexity seemed worth the change.
- Remove ~aws_xml_parser_new(), aws_xml_parser_destroy(), aws_xml_parser_parse()~, replace with  `aws_xml_parse()`.
    - The new() and destroy() calls were unnecessary. Removing them simplifies use.
- `aws_xml_node_get_name()` just returns `aws_byte_cursor()`, instead of ~int~
    - This can't fail. Changing the signature simplifies use.
- Raise `AWS_ERROR_INVALID_XML` instead of ~AWS_ERROR_MALFORMED_INPUT_STRING~
    - This change is more wishy-washy. It seemed useful to get this new, more specific, error code if it bubbling up from deep within some larger operation, like an S3 meta-request.

**API BREAK:**
We don't know any external uses of this API, so it seems safe to change. The API is only intended for internal use by the aws-c libraries, which are being fixed up now. This API was quickly written as private code in aws-c-auth (https://github.com/awslabs/aws-c-auth/pull/40), then moved to public in aws-c-common (https://github.com/awslabs/aws-c-common/pull/674) when aws-c-s3 also needed to parse XML. The fact that it was originally private is why this API didn't get more scrutiny originally.


include/aws/common/xml_parser.h
```diff
@@ -24,62 +23,51 @@ struct aws_xml_attribute {
...
  * 2. call aws_xml_node_traverse() on the node to descend into the node with a new callback and user_data.
  * 3. call aws_xml_node_as_body() to retrieve the contents of the node as text.
  *
+ * You MUST NOT call both aws_xml_node_traverse() and aws_xml_node_as_body() on the same node.
+ *
  * return true to continue the parsing operation.
  */
-typedef bool(
-    aws_xml_parser_on_node_encountered_fn)(struct aws_xml_parser *parser, struct aws_xml_node *node, void *user_data);
+typedef int(aws_xml_parser_on_node_encountered_fn)(struct aws_xml_node *node, void *user_data);
```
TingDaoK, line 34: So, if user wants to just stop parsing without error, they cannot do it now?

I guess we still have use case like I found the response I want and, now we can just stop parsing and it's not an error?

TingDaoK: Oh, I saw the description.

To me, I feel like the complexity is not too bad, just one more boolean, while it helps to improve performance? I just don't like the idea of having to do the extra parsing.

graebm: Yes. Only errors can stop the parser now.

I found that in all cases but one, we were returning false because an error happened.

My 1st pass did have an extra `bool *stop_parsing` param that the callback could set, but I got sick of ignoring the param (`(void)stop_parsing;`) in every callback. Also, the `stop_parsing` code paths were not well tested. I felt better just removing the feature.

It was only used in the `aws_xml_get_top_level_tag()` functions. I changed it so its callback just immediately returns if its called again after finding the one thing it's looking for.

But in most cases we're gathering data from a few elements (e.g. credentials provider needs access key, secret, expiration, etc) and didn't use it

TingDaoK (approved):

PR #1047, opened 2023-07-26
Pickup latest release from cJSON

TingDaoK:
*Issue #, if available:*

Pick up change from https://github.com/DaveGamble/cJSON/releases/tag/v1.7.16

*Description of changes:*

- Copy/pasted from the cJSON repo, and updated based on our lint.
- To review, you can ignore the white space.


source/external/cJSON.c
```diff
@@ -1028,954 +1038,959 @@ static cJSON_bool print_object(const cJSON * const item, printbuffer * const out
...
-  }
-
-  return item;
+    parse_buffer buffer = { 0, 0, 0, 0, { 0, 0, 0 } };
+    cJSON *item = NULL;
+
+    /* reset error position */
+    global_error.json = NULL;
+    global_error.position = 0;
+
```
waahm7, line 1104: global_error is removed in our copy of cJson due to race conditions https://github.com/awslabs/aws-c-common/pull/972/files

source/external/cJSON.c
```diff
@@ -86,87 +86,93 @@ THE SOFTWARE.
...
-  size_t position;
+    const unsigned char *json;
+    size_t position;
 } error;
+static error global_error = { NULL, 0 };
+
+CJSON_PUBLIC(const char *) cJSON_GetErrorPtr(void)
+{
+    return (const char*) (global_error.json + global_error.position);
+}
```
waahm7: global_error and `cJSON_GetErrorPtr` is removed in our copy of cJson due to race conditions [#972 (files)](https://github.com/awslabs/aws-c-common/pull/972/files)

source/external/cJSON.c
```diff
@@ -177,844 +183,848 @@ static void * CJSON_CDECL internal_realloc(void *pointer, size_t size)
...
+
+    /* This checks for NaN and Infinity */
+    if (isnan(d) || isinf(d))
+    {
+        length = snprintf((char*)number_buffer, sizeof(number_buffer) / sizeof(char), "null");
+    }
+	else if (d == (double)item->valueint) // NOLINT
+	{
+        length = snprintf((char*)number_buffer, sizeof(number_buffer) / sizeof(char), "%d", item->valueint);
+	}
```
waahm7, line 563: bad lint? Not sure why we haven't removed NOLINT in this case

waahm7 (approved):
Fix & Ship. I think we should just disable lint on these files and keep it similar to external cJson as much as possible.

include/aws/common/external/cJSON.h
```diff
@@ -280,15 +280,22 @@ CJSON_PUBLIC(cJSON*) cJSON_AddObjectToObject(cJSON * const object, const char *
 CJSON_PUBLIC(cJSON*) cJSON_AddArrayToObject(cJSON * const object, const char * const name);
 
 /* When assigning an integer value, it needs to be propagated to valuedouble too. */
-#define cJSON_SetIntValue(object, number) ((object) ? (object)->valueint = (object)->valuedouble = (number) : (number)) //NOLINT
+#define cJSON_SetIntValue(object, number) ((object) ? (object)->valueint = (object)->valuedouble = (number) : (number))
 /* helper for the cJSON_SetNumberValue macro */
 CJSON_PUBLIC(double) cJSON_SetNumberHelper(cJSON *object, double number);
-#define cJSON_SetNumberValue(object, number) ((object != NULL) ? cJSON_SetNumberHelper(object, (double)number) : (number)) //NOLINT
+#define cJSON_SetNumberValue(object, number) (((object) != NULL) ? cJSON_SetNumberHelper((object), (double)(number)) : (number))
```
waahm7, line 286: Brackets arount object seems redundant 🤔 Ignore if there is a reason for those
```suggestion
#define cJSON_SetNumberValue(object, number) ((object != NULL) ? cJSON_SetNumberHelper(object, (double)(number)) : (number))
```

source/external/cJSON.c
TingDaoK: aha, it's white space vs tab.

include/aws/common/external/cJSON.h
waahm7: Saw the lint errors, we should probably just disable lint on these files.

include/aws/common/external/cJSON.h
```diff
@@ -28,10 +28,10 @@ THE SOFTWARE.
  *   (3) Remove cJSON_GetErrorPtr and global_error as they are not thread-safe.
  */
 
-/* clang-format off */
+/* NOLINTBEGIN */
```
DmitriyMusatkin, line 32: why are we reenabling clang format? if we disabling tidy, we should disable format as well

TingDaoK: there is a .clang-format config that disabled the format already.

PR #1048, opened 2023-07-27
Clang14 tidy

TingDaoK:
- Upgrade our clang tidy to llvm14
- Disable warnings for:
  - `bugprone-easily-swappable-parameters`
  - `readability-identifier-length` -- variable name needs to be more than 3 chars :)
  - `misc-no-recursion` -- :)
  - `readability-function-cognitive-complexity` -- Nice to have, but waay too late, and way to many
  - `readability-magic-numbers` -- Const numbers should be declared as variables, not too bad, but just waay too many in our codebase


source/posix/condition_variable.c
```diff
@@ -92,13 +92,13 @@ int aws_condition_variable_wait_for(
...
-
-    time_to_wait += current_sys_time;
+    uint64_t time_to_wait_unsigned = (uint64_t)time_to_wait;
+    time_to_wait_unsigned += current_sys_time;
 
     struct timespec ts;
     uint64_t remainder = 0;
     ts.tv_sec =
-        (time_t)aws_timestamp_convert((uint64_t)time_to_wait, AWS_TIMESTAMP_NANOS, AWS_TIMESTAMP_SECS, &remainder);
+        (time_t)aws_timestamp_convert(time_to_wait_unsigned, AWS_TIMESTAMP_NANOS, AWS_TIMESTAMP_SECS, &remainder);
```
DmitriyMusatkin, line 101: nit: why do we need a temp variable? cant we just have time_to_wait + current_sys_time inline?

DmitriyMusatkin (approved):

include/aws/common/logging.h
```diff
@@ -58,8 +58,10 @@ enum aws_log_level {
 typedef uint32_t aws_log_subject_t;
 
 /* Each library gets space for 2^^10 log subject entries */
-#define AWS_LOG_SUBJECT_STRIDE_BITS 10
-#define AWS_LOG_SUBJECT_STRIDE (1U << AWS_LOG_SUBJECT_STRIDE_BITS)
+enum {
```
bretambrose, line 61: Not a fan of this change.  Adjacent integral constant defines is not necessarily something that should be an enum

include/aws/common/logging.h
```diff
@@ -58,8 +58,10 @@ enum aws_log_level {
...
+    AWS_LOG_SUBJECT_STRIDE_BITS = 10,
+    AWS_LOG_SUBJECT_STRIDE = (1U << AWS_LOG_SUBJECT_STRIDE_BITS),
```
graebm, line 63: This might be changing `AWS_LOG_SUBJECT_STRIDE` from unsigned to signed.
all enums are `int` in C
but it looks like the old `#define` was trying to be unsigned

include/aws/common/statistics.h
```diff
@@ -18,8 +18,10 @@ struct aws_array_list;
 typedef uint32_t aws_crt_statistics_category_t;
 
 /* Each library gets space for 2^^8 category entries */
-#define AWS_CRT_STATISTICS_CATEGORY_STRIDE_BITS 8
-#define AWS_CRT_STATISTICS_CATEGORY_STRIDE (1U << AWS_CRT_STATISTICS_CATEGORY_STRIDE_BITS)
+enum {
+    AWS_CRT_STATISTICS_CATEGORY_STRIDE_BITS = 8,
+    AWS_CRT_STATISTICS_CATEGORY_STRIDE = (1U << AWS_CRT_STATISTICS_CATEGORY_STRIDE_BITS)
```
graebm, line 23: same thing with signed/unsigned

include/aws/common/logging.h
DmitriyMusatkin: is your concern mostly with just grouping of unrelated constants under the same enum or enum usage for constants in general?

PR #1058, opened 2023-09-12
Consistently use lowercase for Windows libraries and header files

giordano:
Windows typically uses case-insensitive file systems, so the case of libraries and header files is irrelevant, but MinGW consistently use lowercase names everywhere, and when cross-compiling on Linux systems with case-sensitive file systems the case of these names is very much relevant.


graebm (approved):
LGTM
I'll get the CI to run...

graebm:
It's failing to build with MSVC (see CI links)

giordano:
Ok, I removed the include of `winnt.h` which was orthogonal to the cases problem, I think it was something like #1059 but now I don't remember what it was needed for exactly.

PR #1059, opened 2023-09-12
Include `winsock.h` to define `htonl` on Windows

giordano:
[`htonl`](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-htonl) is defined in `winsock.h`.

### DIFF
--- a/include/aws/common/byte_order.inl
+++ b/include/aws/common/byte_order.inl
@@ -11,6 +11,7 @@
 
 #ifdef _WIN32
 #    include <stdlib.h>
+#    include <winsock.h>
 #else
 #    include <netinet/in.h>
 #endif /* _MSC_VER */


giordano:
Beautiful, including `winsock.h` also includes `wingdi.h` which defines a macro called `ERROR`, which screws up https://github.com/awslabs/aws-c-common/blob/2f8f0825b5c9d3190c1a29f7182958648ae886e1/source/uri.c#L19-L26 I have no clue who thought calling a macro in a system header file `ERROR` was a sensible idea, but you may want to rename that enum.

include/aws/common/byte_order.inl
```diff
@@ -11,6 +11,7 @@
 
 #ifdef _WIN32
 #    include <stdlib.h>
+#    include <winsock.h>
```
graebm, line 14: We'll need to take a different approach.

This `.inl` file is included from our public `.h` files, and it's basically a computer crime to include a Windows header from a public header. Due to the sloppy naming infecting all code that includes them.

We'll need to move this function implemention to a normal .c file and do the include from there, at least for windows

giordano:
Why was this PR closed, and without any explanation?  Was the issue resolved elsewhere?

PR #1063, opened 2023-10-11
Add an environment loader API and implementation for some DMI related fields

JonathanHenson:
Adds an environment loading API as these probes are not cheap and this is my proposal for keeping it from getting to messy over time.

Long term I see this containing metadata about NIC configurations, cpu and gpu information, memory bus information, etc... Then other parts of the system can load the environment once to make optimization and defaulting decisions.

This is currently being used in aws-c-s3 for determining if we're running on ec2 nitro and figuring out the instance type. It's very much intended to contain more data over time as the configurations get more involved.

### DIFF
--- a/include/aws/common/private/system_info_priv.h
+++ b/include/aws/common/private/system_info_priv.h
@@ -16,7 +16,7 @@ struct aws_system_environment {
     enum aws_platform_os os;
     size_t cpu_count;
     size_t cpu_group_count;
-    void *additional_impl_data;
+    void *impl;
 };
 
 /**
--- a/include/aws/common/private/system_info_priv.h
+++ b/include/aws/common/private/system_info_priv.h
@@ -22,8 +22,4 @@ struct aws_system_environment {
 int aws_system_environment_load_platform_impl(struct aws_system_environment *env);
 void aws_system_environment_destroy_platform_impl(struct aws_system_environment *env);
 
-
-void aws_system_environment_load_virtualization_vendor_impl(struct aws_system_environment *env);
-void aws_system_environment_load_virtualization_product_name_impl(struct aws_system_environment *env);
-
 #endif /* AWS_COMMON_SYSTEM_INFO_PRIV_H */
--- a/include/aws/common/private/system_info_priv.h
+++ b/include/aws/common/private/system_info_priv.h
@@ -1,5 +1,5 @@
-#ifndef AWS_COMMON_SYSTEM_INFO_PRIV_H
-#define AWS_COMMON_SYSTEM_INFO_PRIV_H
+#ifndef AWS_COMMON_PRIVATE_SYSTEM_INFO_PRIV_H
+#define AWS_COMMON_PRIVATE_SYSTEM_INFO_PRIV_H
 /**
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
@@ -33,4 +33,4 @@ int aws_system_environment_load_platform_impl(struct aws_system_environment *env
  */
 void aws_system_environment_destroy_platform_impl(struct aws_system_environment *env);
 
-#endif /* AWS_COMMON_SYSTEM_INFO_PRIV_H */
+#endif // AWS_COMMON_PRIVATE_SYSTEM_INFO_PRIV_H
--- a/include/aws/common/private/system_info_priv.h
+++ b/include/aws/common/private/system_info_priv.h
@@ -5,9 +5,9 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 
-#include <aws/common/system_info.h>
 #include <aws/common/byte_buf.h>
 #include <aws/common/string.h>
+#include <aws/common/system_info.h>
 
 struct aws_system_environment {
     struct aws_allocator *allocator;
@@ -19,7 +19,18 @@ struct aws_system_environment {
     void *additional_impl_data;
 };
 
+/**
+ * For internal implementors. Fill in info in env that you're able to grab, such as dmi info, os version strings etc...
+ * in here. The default just returns AWS_OP_SUCCESS. This is currently only implemented for linux.
+ *
+ * Returns AWS_OP_ERR if the implementation wasn't able to fill in required information for the platform.
+ */
 int aws_system_environment_load_platform_impl(struct aws_system_environment *env);
+
+/**
+ * For internal implementors. Cleans up anything allocated in aws_system_environment_load_platform_impl,
+ * but does not release the memory for env.
+ */
 void aws_system_environment_destroy_platform_impl(struct aws_system_environment *env);
 
 #endif /* AWS_COMMON_SYSTEM_INFO_PRIV_H */
--- a/include/aws/common/private/system_info_priv.h
+++ b/include/aws/common/private/system_info_priv.h
@@ -4,13 +4,14 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
  */
-
 #include <aws/common/byte_buf.h>
+#include <aws/common/ref_count.h>
 #include <aws/common/string.h>
 #include <aws/common/system_info.h>
 
 struct aws_system_environment {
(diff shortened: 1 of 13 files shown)


include/aws/common/system_info.h
```diff
@@ -21,8 +22,48 @@ struct aws_cpu_info {
...
+AWS_COMMON_API
+void aws_system_environment_destroy(struct aws_system_environment *env);
+
+/**
+ * Returns the virtualization vendor for the specified compute environment, e.g. "Xen, Amazon EC2, etc..."
+ *
+ * The return value may be empty and in that case no vendor was detected.
+ */
+AWS_COMMON_API
+struct aws_byte_cursor aws_system_environment_get_virtualization_vendor(const struct aws_system_environment *env);
```
graebm, line 45: With this API, there's no way to differentiate "unknown because none of my heuristics matched" from "unexpected error while trying to determine (e.g. IMDS flakiness)"

If you returned `struct aws_string *` you could return empty string for unknown, and NULL-with-error if "unexpected error"

but then, the error would have happened earlier, during `aws_system_environment_load()`, so you'd need to store the reason for failure, and raise it when the appropriate getter is called

include/aws/common/system_info.h
```diff
@@ -21,8 +22,48 @@ struct aws_cpu_info {
...
+struct aws_system_environment;
+
 AWS_EXTERN_C_BEGIN
 
+/**
+ * Allocates and initializes information about the system the current process is executing on.
+ * If successful returns an instance of aws_system_environment. If it fails, it will return NULL.
+ */
+AWS_COMMON_API
+struct aws_system_environment *aws_system_environment_load(struct aws_allocator *allocator);
```
graebm, line 34: Does this design scale? Where you load all the info up front and returned cached results later?

That means, for every new getter you might want to add, the user MUST wait during load() for all N bits of info to be fetched, even if they weren't interested in all of them.

An alternative design is to lazily fetch info when a getter is called, and cache it. This would probably mean a lock around each function implementation.

I guess we could keep this API, load everything upfront for now, and change it later to the lazy fetching approach

source/file.c
```diff
@@ -73,14 +73,25 @@ int aws_byte_buf_init_from_file(struct aws_byte_buf *out_buf, struct aws_allocat
...
+                aws_error_name(aws_last_error()));
+            goto error;
+        } else {
+            AWS_LOGF_WARN(
+                AWS_LS_COMMON_IO,
+                "static: reading file:'%s' reports longer length %d than actually reading from it: %d. This is not "
+                "necessarily an error as devices will usually report a page for length regardless of actual length.",
+                filename,
+                (int)read,
+                (int)out_buf->len);
```
graebm, line 92: debatable: I realize this is only WARN, but If it's going to print every single time someone runs an app and everything is functioning normally... that's annoying. I know we have ERRORs in our codebase that print every time under normal operation and it's annoying.

But also, this might be worth warning about!

How about this as a compromise: Call fread() until EOF and just use the file_size query as a suggestion. fread() doesn't promise to give you everything you ask for in one call anyway (in theory). Also, that would let us get rid of the [alternate implementation of this function in aws-c-iot](https://github.com/awslabs/aws-c-iot/blame/9e1aae59a41fc6574cf5b093a03159e09697cc73/source/linux/network.c#L122).

graebm: I'll do the finicky rewrite to save you time

include/aws/common/private/system_info_priv.h
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+#include <aws/common/byte_buf.h>
+#include <aws/common/string.h>
+#include <aws/common/system_info.h>
+
+struct aws_system_environment {
+    struct aws_allocator *allocator;
+    struct aws_byte_buf virtualization_vendor;
+    struct aws_byte_buf product_name;
```
graebm, line 15: trivial: use the string class
```suggestion
    struct aws_string *virtualization_vendor;
    struct aws_string *product_name;
```

include/aws/common/private/system_info_priv.h
```diff
@@ -0,0 +1,36 @@
...
+    enum aws_platform_os os;
+    size_t cpu_count;
+    size_t cpu_group_count;
+    void *additional_impl_data;
```
graebm, line 19: trivial: impl is fine
```suggestion
    void *impl;
```

include/aws/common/system_info.h
```diff
@@ -21,8 +22,48 @@ struct aws_cpu_info {
...
+
+AWS_COMMON_API
+void aws_system_environment_destroy(struct aws_system_environment *env);
```
graebm, line 37: ref-count O'clock:
```
struct aws_system_environment *aws_system_environment_acquire(struct aws_system_environment *env);
struct aws_system_environment *aws_system_environment_release(struct aws_system_environment *env);
```

include/aws/common/system_info.h
```diff
@@ -21,8 +22,48 @@ struct aws_cpu_info {
...
+ * Returns the number of processors for the specified compute environment.
+ */
+AWS_COMMON_API
+size_t aws_system_environment_get_processor_count(struct aws_system_environment *env);
+
+/**
+ * Returns the number of separate cpu groupings (multi-socket configurations or NUMA).
+ */
+AWS_COMMON_API
+size_t aws_system_environment_get_cpu_group_count(struct aws_system_environment *env);
```
graebm, line 65: the other getters take `const env`

source/system_info.c
```diff
@@ -0,0 +1,69 @@
...
+        (void *)env,
+        AWS_BYTE_CURSOR_PRI(aws_system_environment_get_virtualization_vendor(env)));
+
+    env->os = aws_get_platform_build_os();
+    env->cpu_count = aws_system_info_processor_count();
+    env->cpu_group_count = aws_get_cpu_group_count();
+
+    return env;
+error:
+    aws_mem_release(allocator, env);
```
graebm, line 39: trivial: I like the pattern of just calling the destroy() function, so that if you're halfway through setup when it fails, you still tear down every individual member
```suggestion
    aws_system_environment_destroy(env);
```

source/system_info.c
```diff
@@ -0,0 +1,69 @@
...
+
+    return NULL;
+}
+
+void aws_system_environment_destroy(struct aws_system_environment *env) {
+    if (env) {
+        aws_system_environment_destroy_platform_impl(env);
+        aws_mem_release(env->allocator, env);
```
graebm, line 47: maybe just clean up the strings/buffers here, since they're common fields. Vs doing it in each destroy_platform_impl()

include/aws/common/system_info.h
JonathanHenson: there's no IMDS flaky ness. There's no AWS here, it's just generic, read crap off of the OS as best you can.

There's no handleable error. You either have the info or you don't.

include/aws/common/private/system_info_priv.h
JonathanHenson: why? if i use that I need more copies. I'm reading directly into the buffer and initializing them.

include/aws/common/system_info.h
JonathanHenson: It scales better than needing to load, allocate, and deal with the complexity of mutexes and global state. I don't think those are ever the correct answer unless there's literally no other way to do it.

include/aws/common/system_info.h
JonathanHenson: thanks, perfect.

source/system_info.c
JonathanHenson: the tradeoff is cleanliness. As a general rule, it's better for things to be freed by the code that allocated it, than passed around and god knows who cleans it up.

source/file.c
JonathanHenson: I rewrote it.

graebm: Sorry, I'd already done it before I noticed your comment:
https://github.com/awslabs/aws-c-common/pull/1064

graebm: Mine handles some more edge cases: like the OS reporting 0 file size (this actually happens), or the edge case where the buffer perfectly fills up and there's no room for the null terminator

graebm (approved):

PR #1064, opened 2023-10-12
Support reading "special files" that lie about their size

graebm:
**Issue**:
`aws_byte_buf_init_from_file()` didn't work with "special files", which don't accurately report their size. For example, on my Amazon Linux 2:
- `/proc/cpuinfo`: size is 0, but contents are several KB of data.
- `/sys/devices/virtual/dmi/id/product_name`: size is 4096, but contents are "c5.2xlarge"

**Description of changes**:
- `aws_byte_buf_init_from_file()` no longer 100% trusts the reported size. Size is used as a hint, but it always reads now until EOF, growing the buffer if necessary.
- New function `aws_byte_buf_init_from_file_with_size_hint()` lets users pass a hint, rather than querying the OS for it.

### DIFF
--- a/source/file.c
+++ b/source/file.c
@@ -44,7 +44,8 @@ FILE *aws_fopen(const char *file_path, const char *mode) {
     return file;
 }
 
-static int s_byte_buf_init_from_file(
+/* Helper function used by aws_byte_buf_init_from_file() and aws_byte_buf_init_from_file_with_size_hint() */
+static int s_byte_buf_init_from_file_impl(
     struct aws_byte_buf *out_buf,
     struct aws_allocator *alloc,
     const char *filename,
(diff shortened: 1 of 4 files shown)


bretambrose (commented):
maybe with_size_hint() rather than "special" which isn't very meaningful

graebm:
> maybe with_size_hint() rather than "special" which isn't very meaningful

I also just encountered this term yesterday for the first time, but it came up in 2 different contexts! Apparently it's the official term: https://www.google.com/search?q=special+files

So I assumed it was something everyone but me already knew. But if you and me were both thrown by the term, then probably not.

So yeah, I'll change it

source/file.c
```diff
@@ -34,54 +44,91 @@ FILE *aws_fopen(const char *file_path, const char *mode) {
     return file;
 }
 
-int aws_byte_buf_init_from_file(struct aws_byte_buf *out_buf, struct aws_allocator *alloc, const char *filename) {
+static int s_byte_buf_init_from_file(
+    struct aws_byte_buf *out_buf,
+    struct aws_allocator *alloc,
+    const char *filename,
+    bool use_file_size_as_hint,
+    size_t size_hint) {
```
waahm7, line 52: Debatable / Code Clarity: There are two different ways to use this function, which is a bit confusing. I suggest we change this function to 
```suggestion
    s_byte_buf_init_from_file(
    struct aws_byte_buf *out_buf,
    struct aws_allocator *alloc,
    const char *filename,
    size_t size_hint)
```
and let `aws_byte_buf_init_from_file` calculate size_hint.

waahm7 (approved):

source/file.c
graebm: I tried writing it that way first, but it was more confusing IMHO.

You ended up with 3 functions that needed to deal with error-handling instead of just 1, and scrambled logic around instead of putting it all in 1 place.

PR #1066, opened 2023-10-23
Switch default allocator to basic system one

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Default allocator is currently aligned and in practice has a lot of overhead in peak memory usage in heavy acquire/free scenarios. ex s3 30gb upload can easily hit 14gb peak with aligned allocator, while basic system allocator peaks at around 6gb. No major perf degradations were observed from switching. 

Switch default to basic system allocator as its more memory friendly option. aligned allocator is exposed as alternative and can be manually set. Will revisit making aligned default once we have more bandwidth to test and experiment with allocation patterns.


include/aws/common/allocator.h
```diff
@@ -12,6 +12,25 @@
...
+ * Quick guide to allocators:
+ *  CRT offers several flavours of allocators:
+ * - default: basic allocator, targeted toward most common use case. behavior
+ *   can change over time. Current default aligns small allocations on 8 byte
+ *   boundary and big buffers on 32/64 byte (system dependent) boundary.
+ *   Aligned mem can improve perf on some operations, like memcpy or hashes.
+ *   Depending on a system, can result in higher peak memory count in heavy
+ *   acquire/free scenarios (ex. s3), due to memory fragmentation related to how
+ *   aligned allocators work (overallocate, find aligned offset, release extra memory)
+ * - no_align: similar to default, but does not do any alignment. In practice,
```
waahm7, line 25: trivial: `no_align` name is based on the fact that our `default allocator` does alignment. What if, in the future, we change our `default allocator` to something else? Maybe something like `system_allocator`?  

I am conflicted because `no_alloc` name does signify that the reason we are using system allocator instead of default allocator is because we don't want memory allignment. So `system_allocator` might be confusing to the future reader of the codebase. Up to you.

source/allocator.c
```diff
@@ -117,6 +117,57 @@ struct aws_allocator *aws_default_allocator(void) {
     return &default_allocator;
 }
 
+static void *s_no_align_malloc(struct aws_allocator *allocator, size_t size) {
```
waahm7, line 120: Same here for all these, if we do decide to change the name.

source/allocator.c
```diff
@@ -117,6 +117,57 @@ struct aws_allocator *aws_default_allocator(void) {
...
+        s_no_align_free(allocator, ptr);
+    }
+
+    return new_mem;
+}
+
+static void *s_no_align_calloc(struct aws_allocator *allocator, size_t num, size_t size) {
+    (void)allocator;
+    void *mem = calloc(num, size);
+    AWS_PANIC_OOM(mem, "Unhandled OOM encountered in s_default_malloc");
```
waahm7, line 156:
```suggestion
    AWS_PANIC_OOM(mem, "Unhandled OOM encountered in s_no_align_calloc");
```

waahm7 (approved):

source/allocator.c
```diff
@@ -117,6 +117,57 @@ struct aws_allocator *aws_default_allocator(void) {
     return &default_allocator;
 }
 
+static void *s_no_align_malloc(struct aws_allocator *allocator, size_t size) {
+    (void)allocator;
+    void *result = malloc(size);
```
graebm, line 122:
```suggestion
    /* Protect against https://wiki.sei.cmu.edu/confluence/display/c/MEM04-C.+Beware+of+zero-length+allocations */
    AWS_FATAL_PRECONDITION(size > 0);
    void *result = malloc(size);
```

DmitriyMusatkin: we have that check in the aws_mem_acquire already

include/aws/common/allocator.h
```diff
@@ -12,6 +12,23 @@
 AWS_PUSH_SANE_WARNING_LEVEL
 AWS_EXTERN_C_BEGIN
 
+/*
+ * Quick guide to allocators:
+ *  CRT offers several flavours of allocators:
+ * - default: basic allocator that invokes system one directly.
+ * - aligned: basic allocator, targeted toward most common use case. behavior
+ *   can change over time. Current default aligns small allocations on 8 byte
+ *   boundary and big buffers on 32/64 byte (system dependent) boundary.
```
graebm, line 21: trivial: remove wording about "most common use case" and "Current default"
```suggestion
 * - aligned: basic allocator that aligns small allocations on 8 byte
 *   boundary and big buffers on 32/64 byte (system dependent) boundary.
```

graebm (approved):

PR #1072, opened 2023-11-02
Added cross process Lock.

JonathanHenson:

### DIFF
--- a/include/aws/common/cross_process_lock.h
+++ b/include/aws/common/cross_process_lock.h
@@ -13,7 +13,8 @@ AWS_EXTERN_C_BEGIN
 /**
  * Attempts to acquire a system-wide (not per process or per user) lock scoped by instance_nonce.
  * For any given unique nonce, a lock will be returned by the first caller. Subsequent calls will
- * return NULL until the either the process owning the lock exits or the program owning the lock
+ * return NULL and raise AWS_ERROR_MUTEX_CALLER_NOT_OWNER
+ * until the either the process owning the lock exits or the program owning the lock
  * calls aws_cross_process_lock_release() explicitly.
  *
  * If the process exits before the lock is released, the kernel will unlock it for the next consumer.
--- a/include/aws/common/cross_process_lock.h
+++ b/include/aws/common/cross_process_lock.h
@@ -7,19 +7,19 @@
 #include <aws/common/byte_buf.h>
 #include <aws/common/common.h>
 
-struct aws_ipc_util_instance_lock;
+struct aws_cross_process_lock;
 AWS_EXTERN_C_BEGIN
 
 /**
  * Attempts to acquire a system-wide (not per process or per user) lock scoped by instance_nonce.
  * For any given unique nonce, a lock will be returned by the first caller. Subsequent calls will
  * return NULL until the either the process owning the lock exits or the program owning the lock
- * calls aws_ipc_util_instance_lock_release() explicitly.
+ * calls aws_cross_process_lock_release() explicitly.
  *
  * If the process exits before the lock is released, the kernel will unlock it for the next consumer.
  */
 AWS_COMMON_API
-struct aws_ipc_util_instance_lock *aws_ipc_util_instance_lock_try_acquire(
+struct aws_cross_process_lock *aws_cross_process_lock_try_acquire(
     struct aws_allocator *allocator,
     struct aws_byte_cursor instance_nonce);
 
@@ -27,7 +27,7 @@ struct aws_ipc_util_instance_lock *aws_ipc_util_instance_lock_try_acquire(
  * Releases the lock so the next caller (may be another process) can get an instance of the lock.
  */
 AWS_COMMON_API
-void aws_ipc_util_instance_lock_release(struct aws_ipc_util_instance_lock *instance_lock);
+void aws_cross_process_lock_release(struct aws_cross_process_lock *instance_lock);
 
 AWS_EXTERN_C_END
 
--- a/include/aws/common/cross_process_lock.h
+++ b/include/aws/common/cross_process_lock.h
@@ -1,5 +1,5 @@
-#ifndef AWS_COMMON_IPC_UTIL_H
-#define AWS_COMMON_IPC_UTIL_H
+#ifndef AWS_COMMON_CROSS_PROCESS_LOCK_H
+#define AWS_COMMON_CROSS_PROCESS_LOCK_H
 /**
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
@@ -31,4 +31,4 @@ void aws_cross_process_lock_release(struct aws_cross_process_lock *instance_lock
 
 AWS_EXTERN_C_END
 
-#endif /* AWS_COMMON_IPC_UTIL_H */
+#endif /* AWS_COMMON_CROSS_PROCESS_LOCK_H */
--- a/include/aws/common/ipc_util.h
+++ b/include/aws/common/ipc_util.h
@@ -1,11 +1,11 @@
 #ifndef AWS_COMMON_IPC_UTIL_H
 #define AWS_COMMON_IPC_UTIL_H
 /**
-* Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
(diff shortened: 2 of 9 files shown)


source/posix/ipc_util.c
```diff
@@ -0,0 +1,97 @@
...
+        return NULL;
+    }
+    if (flock(fd, LOCK_EX | LOCK_NB) == -1) {
+        AWS_LOGF_TRACE(
+            AWS_LS_COMMON_GENERAL,
+            "static: Lock file %s already acquired by another instance",
+            (const char *)nonce_buf.buffer);
+        close(fd);
+        aws_byte_buf_clean_up(&nonce_buf);
+        return NULL;
```
graebm, line 72: no aws-error is raised if flock() fails, but we raise AWS_ERROR_NO_PERMISSION if open() fails

I get that failing to get the lock isn't really an error, but it's weird to have an API that returns NULL and may-or-may-not set aws_last_error().

source/windows/ipc_util.c
```diff
@@ -0,0 +1,83 @@
...
+    struct aws_byte_cursor to_find = aws_byte_cursor_from_c_str("\\");
+    struct aws_byte_cursor found;
+    AWS_ZERO_STRUCT(found);
+    if (aws_byte_cursor_find_exact(&instance_nonce, &to_find, &found) != AWS_OP_ERR &&
+        aws_last_error() != AWS_ERROR_STRING_MATCH_NOT_FOUND) {
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        return NULL;
+    }
+
+    struct aws_byte_cursor path_prefix = aws_byte_cursor_from_c_str("Global/");
```
graebm, line 31: trivial: Not sure why we'd bother with a prefix, but make it totally generic.

If we worry the nonce won't be unique enough, then make this more unique, like "AwsCrtGlobal/"

But if we think the nonce will already take care to be unique, then a prefix isn't necessary

source/windows/ipc_util.c
```diff
@@ -0,0 +1,83 @@
...
+    HANDLE mutex = CreateMutexA(NULL, FALSE, (LPCSTR)nonce_buf.buffer);
+
+    if (!mutex) {
+        AWS_LOGF_WARN(
+            AWS_LS_COMMON_GENERAL,
+            "static: Lock %s creation failed with error %" PRIx32,
+            (const char *)nonce_buf.buffer,
+            GetLastError());
+        aws_byte_buf_clean_up(&nonce_buf);
+    }
```
graebm, line 46: `return NULL;`


also, no aws-error is raised in this path

source/windows/ipc_util.c
```diff
@@ -0,0 +1,83 @@
...
+     * to the existing object, and the GetLastError function returns ERROR_ALREADY_EXISTS. */
+    if (GetLastError() == ERROR_ALREADY_EXISTS) {
+        AWS_LOGF_TRACE(
+            AWS_LS_COMMON_GENERAL,
+            "static: Lock %s is already acquired by another  instance",
+            (const char *)nonce_buf.buffer,
+            (void *)mutex);
+        CloseHandle(mutex);
+        aws_byte_buf_clean_up(&nonce_buf);
+        return NULL;
```
graebm, line 59: no aws-error raised in this path

source/windows/ipc_util.c
JonathanHenson: Ah, Global is a thing to inform windows how to scope the lock. I think it probably should be local though, so I'll test that.  I think we want the session namespace. The /tmp path already does that due to the way the /tmp directory is mounted.

from msdn: 
"The name can have a "Global" or "Local" prefix to explicitly create the object in the global or session namespace. The remainder of the name can contain any character except the backslash character (\). "

include/aws/common/ipc_util.h
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Attempts to acquire a system-wide (not per process or per user) lock scoped by instance_nonce.
+ * For any given unique nonce, a lock will be returned by the first caller. Subsequent calls will
+ * return NULL until the either the process owning the lock exits or the program owning the lock
+ * calls aws_ipc_util_instance_lock_release() explicitly.
+ *
+ * If the process exits before the lock is released, the kernel will unlock it for the next consumer.
+ */
+AWS_COMMON_API
+struct aws_ipc_util_instance_lock *aws_ipc_util_instance_lock_try_acquire(
```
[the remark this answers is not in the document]
JonathanHenson: what about aws_cross_process_lock()?

source/windows/ipc_util.c
```diff
@@ -0,0 +1,83 @@
...
+struct aws_ipc_util_instance_lock *aws_ipc_util_instance_lock_try_acquire(
+    struct aws_allocator *allocator,
+    struct aws_byte_cursor instance_nonce) {
+
+    /* validate we don't have a directory slash. */
+    struct aws_byte_cursor to_find = aws_byte_cursor_from_c_str("\\");
+    struct aws_byte_cursor found;
+    AWS_ZERO_STRUCT(found);
+    if (aws_byte_cursor_find_exact(&instance_nonce, &to_find, &found) != AWS_OP_ERR &&
+        aws_last_error() != AWS_ERROR_STRING_MATCH_NOT_FOUND) {
```
bretambrose, line 26: let's log this case

source/windows/ipc_util.c
```diff
@@ -0,0 +1,83 @@
...
+
+    HANDLE mutex = CreateMutexA(NULL, FALSE, (LPCSTR)nonce_buf.buffer);
+
+    if (!mutex) {
+        AWS_LOGF_WARN(
+            AWS_LS_COMMON_GENERAL,
+            "static: Lock %s creation failed with error %" PRIx32,
+            (const char *)nonce_buf.buffer,
+            GetLastError());
+        aws_byte_buf_clean_up(&nonce_buf);
```
bretambrose, line 45: same as posix impl; I prefer goto + a single clean up point

include/aws/common/ipc_util.h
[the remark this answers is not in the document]
graebm: sure

source/windows/cross_process_lock.c
```diff
@@ -0,0 +1,93 @@
...
+            "static: Lock " PRInSTR " creation has illegal character \\",
+            AWS_BYTE_CURSOR_PRI(instance_nonce));
+        aws_raise_error(AWS_ERROR_INVALID_ARGUMENT);
+        return NULL;
+    }
+
+    struct aws_cross_process_lock *instance_lock = NULL;
+
+    /* Local prefix, per the docs, specifies session scope rather than global to the user or system. */
+    struct aws_byte_cursor path_prefix = aws_byte_cursor_from_c_str("Local/aws_crt_cross_process_lock/");
```
graebm, line 39: "Local\\" with backslash. It's not clear in the CreateMutexA() docs, but they link [other docs](https://learn.microsoft.com/en-us/windows/win32/termserv/kernel-object-namespaces) that show it
```suggestion
    /* "Local\" prefix, per the docs, specifies user session scope (rather than "Global\" to the system). */
    struct aws_byte_cursor path_prefix = aws_byte_cursor_from_c_str("Local\\aws_crt_cross_process_lock/");
```

include/aws/common/cross_process_lock.h
```diff
@@ -0,0 +1,34 @@
...
+#include <aws/common/byte_buf.h>
+#include <aws/common/common.h>
+
+struct aws_cross_process_lock;
+AWS_EXTERN_C_BEGIN
+
+/**
+ * Attempts to acquire a system-wide (not per process or per user) lock scoped by instance_nonce.
+ * For any given unique nonce, a lock will be returned by the first caller. Subsequent calls will
+ * return NULL until the either the process owning the lock exits or the program owning the lock
```
graebm, line 16:
```suggestion
 * return NULL and raise AWS_ERROR_MUTEX_CALLER_NOT_OWNER
 * until the either the process owning the lock exits or the program owning the lock
```

graebm (approved):
fix & ship

source/windows/cross_process_lock.c
JonathanHenson: thank you!

PR #1078, opened 2023-11-16
Fixed directory creation to ignore umask, and allow cross process loc…

JonathanHenson:
…k to fall back to read only if it can't open the file in rw mode.

testing on this one is hard to automate. Here's how I verified:

create lock directory as user, create lock as user, make sure it works
create lock dir as user, create lock as root, make sure it works for user
create lock dir as root, create lock as root, make sure it works for user
create lock dir as root, try to create lock as user, make sure it works for user.

### DIFF
--- a/source/posix/file.c
+++ b/source/posix/file.c
@@ -40,6 +40,9 @@ int aws_directory_create(const struct aws_string *dir_path) {
         return aws_translate_and_raise_io_error(errno_value);
     }
 
+    /* bypass umask by setting the perms we actually requested */
+    chmod(aws_string_c_str(dir_path), S_IRWXU | S_IRWXG | S_IRWXO);
+
     return AWS_OP_SUCCESS;
 }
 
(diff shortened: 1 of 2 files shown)


nateprewitt (approved):
LGTM after tests wrap

source/posix/file.c
```diff
@@ -40,6 +40,9 @@ int aws_directory_create(const struct aws_string *dir_path) {
         return aws_translate_and_raise_io_error(errno_value);
     }
 
+    /* bypass umask by setting the perms we actually requested */
+    chmod(aws_string_c_str(dir_path), S_IRWXU | S_IRWXG | S_IRWXO);
```
waahm7, line 44: We should respect the umask by default and only apply this to `/tmp/aws_crt_cross_process_lock/`.

JonathanHenson: fair, will update

waahm7 (approved):

PR #1088, opened 2024-02-08
Use _Exit() on test failure, to avoid LeakSanitizer spam

graebm:
Cut down on unrelated spam when a test fails.

[_Exit()](https://en.cppreference.com/w/c/program/_Exit) is in C99. It avoids running cleanup functions on program termination.

This does prevent us from running multiple tests via the `aws-c-common-tests` executable, but we never ever do that. We always use `ctest`, which launches each test in its own process. I literally only learned today that you can pass `-A` to run all tests.

### DIFF
--- a/include/aws/testing/aws_test_harness.h
+++ b/include/aws/testing/aws_test_harness.h
@@ -42,6 +42,7 @@ the AWS_UNSTABLE_TESTING_API compiler flag
 
 /** Prints a message to AWS_TESTING_REPORT_FD using printf format that appends the function, file and line number.
  * If format is null, returns 0 without printing anything; otherwise returns 1.
+ * If function or file are null, the function, file and line number are not appended.
  */
 static int s_cunit_failure_message0(
     const char *prefix,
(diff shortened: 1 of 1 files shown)


DmitriyMusatkin (approved):

include/aws/testing/aws_test_harness.h
```diff
@@ -478,7 +481,8 @@ static inline int s_aws_run_test_case(struct aws_test_harness *harness) {
         const size_t leaked_bytes = aws_mem_tracer_bytes(allocator);
         if (leaked_bytes) {
             aws_mem_tracer_dump(allocator);
-            PRINT_FAIL_INTERNAL0("Test leaked memory: %zu bytes %zu allocations", leaked_bytes, leaked_allocations);
+            PRINT_FAIL_WITHOUT_LOCATION(
```
graebm, line 484: It always annoyed me how the final "test leaked" and "test failed" messages printed their file location and line no. It made it look like the aws_test_runner.h code is what had bugs

Issue #932, opened 2022-09-18
Preprocessor error in SIZE_MAX: missing binary operator before token "0"


philippbara:
### Description
I run into a preprocessor error in SIZE_MAX while building the V2X framework [Artery](https://github.com/riebl/artery/) against the CPP SDK. The ITS-G5 network stack is provided by [Vanetza](https://github.com/riebl/vanetza). Vanetza defines SIZE_MAX in [asn_system.h](https://github.com/riebl/vanetza/blob/master/vanetza/asn1/support/asn_system.h#L130) as follows: `#define SIZE_MAX   ((~((size_t)0)) >> 1)`. This leads to an error in [math.h:16](https://github.com/awslabs/aws-c-common/blob/bd417541440e2660bbf9ba33565a24abaefe7a9a/include/aws/common/math.h#L16):
![error](https://user-images.githubusercontent.com/37438743/190902258-60a60d59-d1e7-4de8-a305-6bfe74f23d79.jpg)

### Steps to reproduce
The issue can easily be reproduced by defining SIZE_MAX as ((~((size_t)0)) >> 1) instead of a non constant integral like 18446744073709551615UL (taken from stdint.h). 

### Solution
Came across a similar issue in [crypto++](https://github.com/weidai11/cryptopp/issues/864). Redefining SIZE_MAX like [here](https://github.com/weidai11/cryptopp/commit/e8b07b162feb) solved the issue for me.

### System info:
Linux 5.15.0-47-generic 51-Ubuntu SMP Thu Aug 11 07:51:15 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
gcc 11.2.0
AWS_SDK_VERSION_STRING "1.9.348"

DmitriyMusatkin:
Note: at the time when this issue was opened Artery had a bug in their headers where they always overrode SIZE_MAX with a non-preprocessor friendly value when compiling using cpp compiler. That has been fixed since then and should not be a problem anymore. 

I dont think we should be overriding SIZE_MAX based on non-standard defines. 

Let us know if its still an issue.

graebm:
The C99 standard requires SIZE_MAX be a constant, safe to use in an `#if`.

See: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf
Section 7.18.3: Limits of other integer types
> these macros shall be replaced by a constant expression suitable for use in #if preprocessing directive

This is a bug in whatever library is re-defining SIZE_MAX.

PR #1091, opened 2024-02-13
Change logic for how we determine how many bits a ptr has

DmitriyMusatkin:
*Issue #, if available:*
https://github.com/awslabs/aws-c-common/issues/932

*Description of changes:*

### DIFF
--- a/include/aws/common/math.h
+++ b/include/aws/common/math.h
@@ -15,9 +15,9 @@
 AWS_PUSH_SANE_WARNING_LEVEL
 
 /* The number of bits in a size_t variable */
-#if SIZE_MAX == UINT32_MAX
+#if UINTPTR_MAX == UINT32_MAX
 #    define SIZE_BITS 32
-#elif SIZE_MAX == UINT64_MAX
+#elif UINTPTR_MAX == UINT64_MAX
 #    define SIZE_BITS 64
 #else
 #    error "Target not supported"


include/aws/common/math.h
graebm, line 1: Edit the PR description to link the issue, else god help the future person who tries to figure out WHY we did this

include/aws/common/math.h
```diff
@@ -15,9 +15,9 @@
 AWS_PUSH_SANE_WARNING_LEVEL
 
 /* The number of bits in a size_t variable */
-#if SIZE_MAX == UINT32_MAX
+#if UINTPTR_MAX == UINT32_MAX
```
graebm, line 18: Tsk, tsk, that other library isn't defining SIZE_MAX to spec:

https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf
> Each instance of these macros shall be replaced by a constant expression suitable for use
in #if preprocessing directive

graebm: But seriously, if SIZE_MAX isn't defined, and so that other library is defining it, what makes us think UINTPTR_MAX is defined?

What environment is that user in?

DmitriyMusatkin: linked the issue. artery library used to always override SIZE_MAX when compiled using C++ with non-preprocessor friendly value.

since we only support c99+ and its been ages for all compilers to add those defines to their header, i would not be too concerned with UINTPTR_MAX missing - if its missing then there are likely myriad of other things we dont support.

bretambrose: Why can't this be a cmake feature check rather than in-source adhoc #if?

PR #1096, opened 2024-03-08
Best Effort Thread Pinning to CPU_ID

waahm7:
*Description of changes:*
Sometimes, `pthread_create` can fail with an `EINVAL` error, or on FreeBSD, it can fail with an `EDEADLK` error if the `cpu_id` is restricted. Since pinning to a particular `cpu_id` is supposed to be best-effort, try to launch a thread again without pinning to a specific `cpu_id`.

### DIFF
--- a/source/posix/thread.c
+++ b/source/posix/thread.c
@@ -277,7 +277,7 @@ int s_init_pthread_attr(size_t stack_size, int32_t cpu_id, pthread_attr_t *out_a
         CPU_ZERO(&cpuset);
         CPU_SET((uint32_t)cpu_id, &cpuset);
 
-        attr_return = pthread_attr_setaffinity_np(attributes_ptr, sizeof(cpuset), &cpuset);
+        attr_return = pthread_attr_setaffinity_np(out_attributes, sizeof(cpuset), &cpuset);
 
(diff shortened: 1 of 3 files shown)


source/posix/thread.c
```diff
@@ -296,12 +296,11 @@ int aws_thread_launch(
...
 
             if (attr_return) {
-                AWS_LOGF_ERROR(
+                AWS_LOGF_WARN(
                     AWS_LS_COMMON_THREAD,
-                    "id=%p: pthread_attr_setaffinity_np() failed with %d.",
+                    "id=%p: pthread_attr_setaffinity_np() failed with %d. Continuing without cpu affinity",
                     (void *)thread,
                     attr_return);
-                goto cleanup;
```
TingDaoK, line 304: I don't like we just continue here:
1. if anything fails from `pthread_create`, we still check if we have the cpu_id set, and we will retry, which will lead to the same `pthread_create` and probably fail the same way.
2. In my opinion, it's easier to follow if we just `goto cleanup` here.

waahm7: Thanks, I have re-added the goto cleanup line.

PR #1098, opened 2024-03-28
import libcbor

TingDaoK:
*Issue #, if available:*

import libcbor as 3rd party dependency. https://github.com/PJK/libcbor

*Description of changes:*

### DIFF
--- a/scripts/import_libcbor.py
+++ b/scripts/import_libcbor.py
@@ -3,29 +3,19 @@ import tempfile
 import shutil
 import subprocess
 import argparse
+import re
 
 def parse_version(version_string):
-    if version_string.startswith('v'):
-        version_string = version_string[1:]
-    parts = version_string.split('.')
-    if not parts:
+    match = re.fullmatch(r'v(\d+)\.(\d+)\.(\d+)', version_string)
+    if not match:
         raise ValueError("Invalid version string")
-    major = int(parts[0])
-    minor = 0
-    patch = 0
-    if len(parts) > 1:
-        minor = int(parts[1])
-
-    if len(parts) > 2:
-        patch = int(parts[2])
-
-    return major, minor, patch
+    return match.group(1), match.group(2), match.group(3)
 
 argument_parser = argparse.ArgumentParser(
     description="Helper to import libcbor as external dependency.")
 
 argument_parser.add_argument("--version",
-                                default="v0.11.0", help="Version string to import")
+                                required=True, help="Version string to import")
 
 args = argument_parser.parse_args()
 major_version, minor_version, patch_version = parse_version(args.version)
@@ -80,10 +70,8 @@ CONFIGURATION_H = f"""
 #endif
 
 #ifdef __clang__
-#    pragma clang diagnostic push
 #    pragma clang diagnostic ignored "-Wreturn-type"
 #elif defined(__GNUC__)
-#    pragma GCC diagnostic push
 #    pragma GCC diagnostic ignored "-Wreturn-type"
 #    pragma GCC diagnostic ignored "-Wunknown-pragmas"
 #    pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
--- a/source/external/libcbor/cbor/configuration.h
+++ b/source/external/libcbor/cbor/configuration.h
@@ -36,10 +36,8 @@
 #endif
 
 #ifdef __clang__
-#    pragma clang diagnostic push
 #    pragma clang diagnostic ignored "-Wreturn-type"
 #elif defined(__GNUC__)
-#    pragma GCC diagnostic push
 #    pragma GCC diagnostic ignored "-Wreturn-type"
 #    pragma GCC diagnostic ignored "-Wunknown-pragmas"
 #    pragma GCC diagnostic ignored "-Wmaybe-uninitialized"


graebm:
Would it be wise to merge this into some `cbor` feature branch, instead of `main` as we iterate on this?

scripts/import_libcbor.py
```diff
@@ -0,0 +1,64 @@
...
+        for file in files:
+            if file.endswith((".c", ".h")):
+                src_file = os.path.join(root, file)
+                rel_path = os.path.relpath(src_file, src_dir)
+                dest_file = os.path.join(output_dir, rel_path)
+                os.makedirs(os.path.dirname(dest_file), exist_ok=True)
+                if dir_name == "cbor":
+                    with open(src_file, 'r') as f:
+                        contents = f.read()
+                    pattern = r'#include "cbor/'
```
graebm, line 41: You can avoid doing any edits if you add the following to aws-c-common's CMakeLists.txt:
```
# libcbor files do includes assuming their at the root, like: #include "cbor/cbor_export.h"
# To make these paths work, add the location we're storing them as a search path.
target_include_directories(${PROJECT_NAME} PRIVATE
        ${CMAKE_CURRENT_SOURCE_DIR}/source/external/libcbor)
```

scripts/libcbor_customize/configuration.h
```diff
@@ -0,0 +1,38 @@
+#ifndef LIBCBOR_CONFIGURATION_H
```
graebm, line 1: Add a comment at the top of this file like: 
```
/**
 * DO NOT DIRECTLY MODIFY THIS FILE:
 *
 * The code in this file is generated from scripts/.../path_to_actual_source
 *  and any modifications should be in there.
 */
```

Otherwise people are going to tweak the comiler warnings below to fix some random CI failure. Then a few months later it will get blown away when someone updates cbor using this script

scripts/libcbor_customize/cbor_export.h
```diff
@@ -0,0 +1,8 @@
+
```
graebm, line 1: same about DO NOT MODIFY

scripts/import_libcbor.py
```diff
@@ -0,0 +1,64 @@
...
+    # Create a separate folder for the copied files
+    output_dir = os.path.join(
+        os.path.dirname(__file__), "..", "source", "external", "libcbor")
+    os.makedirs(output_dir, exist_ok=True)
+
+    # Copy files ending with .c and .h from the /src directory
+    src_dir = os.path.join(temp_repo_dir, "src")
+
+    # Use our customized configurations
+    src_customize_cbor_dir = os.path.join(
```
graebm, line 25: debatable: I'm not really sure if this is better or not ... but instead of additional external files, you can just have strings at the top of this file like:
```
CBOR_EXPORT_H = """
#ifndef CBOR_EXPORT_H
#define CBOR_EXPORT_H
...
"""
```

Or ... just don't bother copying these files in from anywhere, just have them live in the repo and the update script doesn't touch them.

scripts/import_libcbor.py
```diff
@@ -0,0 +1,64 @@
...
+import subprocess
+import re
+
+# Create a temporary directory for cloning the repository
+temp_repo_dir = tempfile.mkdtemp()
+
+try:
+    # Clone the repository into the temporary directory
+    repo_url = "https://github.com/PJK/libcbor.git"
+    clone_command = f"git clone {repo_url} {temp_repo_dir}"
```
graebm, line 13: this script isn't using the latest release
maybe force someone to pass it in via argparse

CMakeLists.txt
```diff
@@ -213,6 +218,10 @@ endif()
 set_target_properties(${PROJECT_NAME} PROPERTIES VERSION 1.0.0)
 set_target_properties(${PROJECT_NAME} PROPERTIES SOVERSION 1)
 
+# libcbor files do includes assuming their at the root, like: #include "cbor/cbor_export.h"
```
graebm, line 221:
```suggestion
# libcbor files do includes like: #include "cbor/cbor_export.h"
```

THIRD-PARTY-LICENSES.txt
```diff
@@ -55,3 +55,29 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 
+------
+
+** libcbor; version 0.11.0 -- https://github.com/PJK/libcbor
```
graebm, line 60: can we omit the version here?

THIRD-PARTY-LICENSES.txt isn't being updated by import_libcbor.py
but even if it were, we also have that scallion tool that other people will use in the future to add more licenses to this file, and we're never going to remember to update that as well...

or is it scallion that put this in here to begin with?

scripts/import_libcbor.py
```diff
@@ -0,0 +1,116 @@
+import os
+import tempfile
+import shutil
+import subprocess
+import re
+
+# TODO support pass-in version and use the tag
```
graebm, line 7: just do this

scripts/import_libcbor.py
```diff
@@ -0,0 +1,116 @@
...
+
+major_version = 0
+minor_version = 11
+patch_version = 0
+
+GENERATED_NOTES = """/**
+ * DO NOT DIRECTLY MODIFY THIS FILE:
+ *
+ * The code in this file is generated from scripts/import_libcbor.py
+ *  and any modifications should be in there.
```
graebm, line 17: utterly trivial:
```suggestion
 * and any modifications should be in there.
```

scripts/import_libcbor.py
```diff
@@ -0,0 +1,116 @@
...
+    for root, dirs, files in os.walk(src_dir):
+        dir_name = os.path.basename(root)
+        for file in files:
+            if file.endswith((".c", ".h")):
+                # copy the source code to ../source/external/libcbor
+                src_file = os.path.join(root, file)
+                rel_path = os.path.relpath(src_file, src_dir)
+                dest_file = os.path.join(output_dir, rel_path)
+                os.makedirs(os.path.dirname(dest_file), exist_ok=True)
+                shutil.copy(src_file, dest_file)
```
graebm, line 97: trivial:
```suggestion
                shutil.copy(src_file, dest_file, dirs_exist_ok=True)
```

scripts/import_libcbor.py
```diff
@@ -0,0 +1,116 @@
...
+try:
+    # Clone the repository into the temporary directory
+    repo_url = "https://github.com/PJK/libcbor.git"
+    clone_command = f"git clone {repo_url} {temp_repo_dir}"
+    subprocess.run(clone_command, shell=True, check=True)
+
+    # Create a separate folder for the copied files
+    output_dir = os.path.join(
+        os.path.dirname(__file__), "..", "source", "external", "libcbor")
+    os.makedirs(output_dir, exist_ok=True)
```
graebm, line 84: delete everything in output_dir before copying things in again

that way, if libcbor ends up renaming or removing a file, we'll also get rid of the old file

scripts/import_libcbor.py
```diff
@@ -0,0 +1,139 @@
...
+    if len(parts) > 2:
+        patch = int(parts[2])
+
+    return major, minor, patch
+
+argument_parser = argparse.ArgumentParser(
+    description="Helper to import libcbor as external dependency.")
+
+argument_parser.add_argument("--version",
+                                default="v0.11.0", help="Version string to import")
```
graebm, line 28: don't have a default value.
someone is going to run this in the future and accidentally downgrade us
```suggestion
                             required=True,
                             help="Version to import (e.g. v0.11.0)")
```

scripts/import_libcbor.py
```diff
@@ -0,0 +1,139 @@
...
+    major = int(parts[0])
+    minor = 0
+    patch = 0
+    if len(parts) > 1:
+        minor = int(parts[1])
+
+    if len(parts) > 2:
+        patch = int(parts[2])
+
+    return major, minor, patch
```
graebm, line 22: trivial: this can be simpler
```suggestion
    match = re.fullmatch(r'v(\d+)\.(\d+)\.(\d+)', version_string)
    if not match:
        raise ValueError("Invalid version string")
    return match.group(1), match.group(2), match.group(3)
```

scripts/import_libcbor.py
```diff
@@ -0,0 +1,139 @@
...
+#    pragma warning(disable : 4244)
+#    pragma warning(disable : 4701)
+#    pragma warning(disable : 4703)
+#endif
+
+#ifdef __clang__
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wreturn-type"
+#elif defined(__GNUC__)
+#    pragma GCC diagnostic push
```
graebm, line 86: trivial: I don't think there's any point in "pushing" a new level in the warning stack, since we never "pop" it off
```suggestion
#    pragma clang diagnostic ignored "-Wreturn-type"
#elif defined(__GNUC__)
```

graebm (approved):
fix & ship

PR #1105, opened 2024-04-19
Move is ipv4/ipv6 functions from Aws-c-sdkutils to Aws-c-common

waahm7:
*Description of changes:*
Moves host_utils into Aws-c-common so that we can use them in all repos. Fixed the `::1` case by making the following two changes.
1. removed `_starts_with(substr,':')` condition because it can start with :
2. Added the `group_count>2` condition while checking the group count and double colon loop.


source/host_utils.c
```diff
@@ -0,0 +1,127 @@
...
+        aws_byte_cursor_advance(&host, 1);
+        --host.len;
+    }
+
+    struct aws_byte_cursor substr = {0};
+    /* first split is required ipv6 part */
+    bool is_split = aws_byte_cursor_next_split(&host, '%', &substr);
+    AWS_ASSERT(is_split); /* function is guaranteed to return at least one split */
+
+    if (!is_split || substr.len == 0 || s_ends_with(substr, ':') ||
```
waahm7, line 95: To fix the `::1` case:
Changes made:
removed `_starts_with(substr,':')` condition here.

source/host_utils.c
```diff
@@ -0,0 +1,127 @@
...
+
+    uint8_t group_count = 0;
+    bool has_double_colon = false;
+    struct aws_byte_cursor group = {0};
+    while (aws_byte_cursor_next_split(&substr, ':', &group)) {
+        ++group_count;
+
+        if (group_count > 8 ||                                         /* too many groups */
+            group.len > 4 ||                                           /* too many chars in group */
+            (has_double_colon && group.len == 0 && group_count > 2)) { /* only one double colon allowed */
```
waahm7, line 108: To fix the `::1` case:
Changes made:
Added the `group_count>2` condition here.

DmitriyMusatkin (approved):

PR #1112, opened 2024-05-08
Fix host parsing for IPv6 URI

waahm7:
*Issue #, if available:*

*Description of changes:*
If it's IPv6 literal uri like `http://[::1]:80/path`, the host is not `[::1]` and is `::1` only. 
RFC: https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2
urllib3:
image

### DIFF
--- a/source/uri.c
+++ b/source/uri.c
@@ -382,12 +382,12 @@ static void s_parse_authority(struct uri_parser *parser, struct aws_byte_cursor 
          * brackets.
          * Ignore the sqaure brackets.
          */
-        size_t host_name_offset = is_IPv6_literal ? -1 : 0;
+        size_t host_name_offset = is_IPv6_literal ? 1 : 0;
         size_t host_name_length_correction = is_IPv6_literal ? 2 : 0;
         if (!port_delim) {
             parser->uri->port = 0;
             parser->uri->host_name.ptr = authority_parse_csr.ptr + host_name_offset;
-            parser->uri->host_name.len = port_delim - authority_parse_csr.ptr - host_name_length_correction;
+            parser->uri->host_name.len = authority_parse_csr.len - host_name_length_correction;
             return;
         }
 
--- a/source/uri.c
+++ b/source/uri.c
@@ -382,7 +382,7 @@ static void s_parse_authority(struct uri_parser *parser, struct aws_byte_cursor 
          * brackets.
          * Ignore the sqaure brackets.
          */
-        size_t host_name_offset = is_IPv6_literal ? 1 : 0;
+        size_t host_name_offset = is_IPv6_literal ? -1 : 0;
         size_t host_name_length_correction = is_IPv6_literal ? 2 : 0;
         if (!port_delim) {
             parser->uri->port = 0;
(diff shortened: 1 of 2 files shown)


source/uri.c
```diff
@@ -375,17 +377,23 @@ static void s_parse_authority(struct uri_parser *parser, struct aws_byte_cursor
         }
 
         const uint8_t *port_delim = memchr(port_search_start, ':', port_search_len);
-
+        /*
+         * RFC-3986 section 3.2.2: A host identified by an IPv6 literal address is represented inside the square
+         * brackets.
+         * Ignore the sqaure brackets.
```
DmitriyMusatkin, line 383: nit: typo

tests/uri_test.c
```diff
@@ -509,7 +509,7 @@ static int s_test_uri_ipv6_parse(struct aws_allocator *allocator, void *ctx) {
         aws_byte_cursor_from_c_str("[2001:db8:85a3:8d3:1319:8a2e:370:7348%25en0]:443");
     ASSERT_BIN_ARRAYS_EQUALS(expected_authority.ptr, expected_authority.len, uri.authority.ptr, uri.authority.len);
 
-    struct aws_byte_cursor expected_host = aws_byte_cursor_from_c_str("[2001:db8:85a3:8d3:1319:8a2e:370:7348%25en0]");
```
DmitriyMusatkin, line 512: can you also verify, whether we should be stripping zone idx or not? what do other libs do?

source/uri.c
```diff
@@ -375,17 +377,23 @@ static void s_parse_authority(struct uri_parser *parser, struct aws_byte_cursor
...
+         */
+        size_t host_name_offset = is_IPv6_literal ? 1 : 0;
+        size_t host_name_length_correction = is_IPv6_literal ? 2 : 0;
         if (!port_delim) {
             parser->uri->port = 0;
-            parser->uri->host_name = authority_parse_csr;
+            parser->uri->host_name.ptr = authority_parse_csr.ptr + host_name_offset;
+            parser->uri->host_name.len = authority_parse_csr.len - host_name_length_correction;
```
DmitriyMusatkin, line 390: ugh, not a big fan of manual cursor manipulation? 
assign authority_parse_csr to host and then advance and dec len by 1 to remove brackets?

tests/uri_test.c
waahm7: It is included 
image

source/uri.c
waahm7: Thanks, fixed.

source/uri.c
waahm7: Thanks, I have updated the code to advance and decrement the length. Initially, I wrote the following code: https://github.com/awslabs/aws-c-common/pull/1112/commits/1127cad9dfc3a730647b27dd238bb1dbe01236c4, but later refactored it to reduce code duplication. We can also refactor that version to use 'advance' and 'decrement length.' If you like that version better, I am conflicted between them.

DmitriyMusatkin: thanks. i think intent of what you are trying to do is much clearer in the latest version.

one further cleanup we could potentially do is to use aws_byte_cursor_next_split to separate host from port in authority. we can probably even get rid of memchr using that approach. but its a bigger refactoring, so we can do it in future

DmitriyMusatkin (approved):

PR #1122, opened 2024-06-05
Fix memtracer bad assumptions on the size of stack trace

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
Fix memtracer assumptions on the size of stack. With optimizations on, there is no guarantees how many frames we get back. So handle any size of stack gracefully.

### DIFF
--- a/source/memtrace.c
+++ b/source/memtrace.c
@@ -131,51 +131,55 @@ static void s_alloc_tracer_track(struct alloc_tracer *tracer, void *ptr, size_t 
     aws_high_res_clock_get_ticks(&alloc->time);
 
     if (tracer->level == AWS_MEMTRACE_STACKS) {
-        /* capture stack frames, skip 2 for this function and the allocation vtable function */
+        /* capture stack frames, 
+        * skip 2 for this function and the allocation vtable function if we have a full stack trace
+        * and otherwise just capture what ever stack trace we got
+        */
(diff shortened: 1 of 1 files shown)


waahm7 (approved):

source/memtrace.c
```diff
@@ -134,7 +134,7 @@ static void s_alloc_tracer_track(struct alloc_tracer *tracer, void *ptr, size_t
         /* capture stack frames, skip 2 for this function and the allocation vtable function */
         AWS_VARIABLE_LENGTH_ARRAY(void *, stack_frames, (FRAMES_TO_SKIP + tracer->frames_per_stack));
         size_t stack_depth = aws_backtrace(stack_frames, FRAMES_TO_SKIP + tracer->frames_per_stack);
-        if (stack_depth >= FRAMES_TO_SKIP) {
+        if (stack_depth > 0) {
```
graebm, line 137: we talked, and maybe there's some other code that crashes if this is 0?
We should either:
- fix that other code, to allow stack_depth of 0
- or FATAL_ASSERT assert here that stack_depth > 0

DmitriyMusatkin: discussed offline. lets assert for now that we have a stack trace, since the rest of the memtraces assumes that allocations have a stack trace.

PR #1123, opened 2024-06-05
disable optimization was not working

TingDaoK:
*Issue #, if available:*

`__asm__ __volatile__("" ::: "memory")` doesn't prevent GCC 7.5.0 to optimize those functions on arm64, and result in a test failure

*Description of changes:*

Use another way to disable optimization.

### DIFF
--- a/tests/memtrace_test.c
+++ b/tests/memtrace_test.c
@@ -63,31 +63,35 @@ static int s_test_memtrace_count(struct aws_allocator *allocator, void *ctx) {
 }
 AWS_TEST_CASE(test_memtrace_count, s_test_memtrace_count)
 
-#if defined(__GNUC__) || defined(__clang__)
-#    define AWS_PREVENT_OPTIMIZATION __asm__ __volatile__("" ::: "memory")
-#else
-#    define AWS_PREVENT_OPTIMIZATION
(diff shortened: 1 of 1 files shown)


tests/memtrace_test.c
```diff
@@ -63,32 +63,36 @@ static int s_test_memtrace_count(struct aws_allocator *allocator, void *ctx) {
...
+#    pragma GCC optimize("O0")
 #endif
 
 AWS_NO_INLINE void *s_alloc_1(struct aws_allocator *allocator, size_t size) {
-    AWS_PREVENT_OPTIMIZATION;
     return aws_mem_acquire(allocator, size);
 }
 
 AWS_NO_INLINE void *s_alloc_2(struct aws_allocator *allocator, size_t size) {
-    AWS_PREVENT_OPTIMIZATION;
```
DmitriyMusatkin, line 78: do we use this anywhere else? should we just update this macro?

TingDaoK: it's not used anywhere else, and it's working differently than the previous way, we need two macros now one for start disable and another one for end. 

But, yeah, it's only used here, so, just got rid of the macro instead.

DmitriyMusatkin (approved):

PR #1126, opened 2024-06-18
Fix default thread options for windows to not pin to any cpu_id

waahm7:
*Issue #, if available:*

*Description of changes:*
- cpu_id should be -1 if we don't want to pin to any cpu_id


graebm (approved):
😳

PR #1135, opened 2024-07-15
Add IS_CXX option to sanitizer cmake module

sfod:
*Issue #, if available:*

`aws_check_sanitizer` checks for the sanitizer flags using `check_c_compiler_flag`. However, for C++ projects C language could be not enabled, which leads to the following error:
```
CMake Error at /usr/share/cmake3/Modules/CheckCSourceCompiles.cmake:109 (try_compile):
  Unknown extension ".c" for file
    aws-iot-device-sdk-cpp-v2/samples/mqtt5/mqtt5_pubsub/cmake-build-tsan/CMakeFiles/CMakeTmp/src.c
  try_compile() works only for enabled languages.  Currently these are:
    CXX
  See project() command to enable other languages.
``` 

A possible solution could be to add `enable_language(C)` before calling `check_c_compiler_flag`. But I'm not sure that C++ and C compilers have the same set of the sanitizer options. And even if they do currently, this might change.

So, using `check_cxx_compiler_flag` for C++ project looks like a more correct approach.

*Description of changes:*

Add `IS_CXX` option to `aws_check_sanitizer` and `aws_add_sanitizers` CMake functions.
This change does not break or change the existing usages of these functions.

### DIFF
--- a/cmake/AwsSanitizers.cmake
+++ b/cmake/AwsSanitizers.cmake
@@ -7,27 +7,18 @@ include(CheckCXXCompilerFlag)
 option(ENABLE_SANITIZERS "Enable sanitizers in debug builds" OFF)
 set(SANITIZERS "address;undefined" CACHE STRING "List of sanitizers to build with")
 
-# This function checks if a given sanitizer is available.
-# Parameters:
-#  sanitizer: The sanitizer to check.
-#  [out_variable]: The variable to assign the result to. Defaults to HAS_SANITIZER_${sanitizer}.
+# This function checks if a sanitizer is available
 # Options:
-#  IS_CXX: This option switches the function to check the CXX compiler.
+#  sanitizer: The sanitizer to check
+#  out_variable: The variable to assign the result to. Defaults to HAS_SANITIZER_${sanitizer}
 function(aws_check_sanitizer sanitizer)
-    set(options IS_CXX)
-    cmake_parse_arguments(ARG "${options}" "" "" ${ARGN})
-
-    # Determine the number of extra parameters passed to the function.
-    list(LENGTH ARG_UNPARSED_ARGUMENTS extra_count)
-
+    list(LENGTH ARGN extra_count)
     if(${extra_count} EQUAL 0)
-        # Use the default out variable.
         set(out_variable HAS_SANITIZER_${sanitizer})
         # Sanitize the variable name to remove illegal characters
         string(MAKE_C_IDENTIFIER ${out_variable} out_variable)
     elseif(${extra_count} EQUAL 1)
-        # Use the out variable provided by a function caller.
-        set(out_variable ${ARG_UNPARSED_ARGUMENTS})
+        set(out_variable ${ARGN})
     else()
         message(FATAL_ERROR "Error: aws_check_sanitizer() called with multiple out variables")
     endif()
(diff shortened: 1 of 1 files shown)


cmake/AwsSanitizers.cmake
```diff
@@ -32,23 +42,37 @@ function(aws_check_sanitizer sanitizer)
...
         set(${out_variable} 0 PARENT_SCOPE)
     endif()
 endfunction()
 
 # This function enables sanitizers on the given target
-# Options:
+# Parameters:
 #  SANITIZERS: The list of extra sanitizers to enable
+# Options:
+#  IS_CXX: This option switches the function to check the CXX compiler.
```
DmitriyMusatkin, line 59: cant you just check whether the target is getting compiled for c or c++?
something like ```get_target_property(target_lang mytarget LANGUAGE)```

sfod: Hmmm, it seems there is no universal property for a target language (which is kinda unexpected). There is the `LINKER_LANGUAGE` property, but it's optional.

Also, the `aws_check_sanitizer` function doesn't have `target` as a parameter, so we can't use this approach there.

I also can't find a way to get a cmake project language.
So, the only alternative approach I can think of is to check the enabled languages, and use `check_cxx_compiler_flag` if C++ is present. But I'm not sure it won't cause some issues in in the projects where both C and C++ are used.

DmitriyMusatkin: maybe try CMAKE_CXX_COMPILER_LOADED and CMAKE_C_COMPILER_LOADED. seems to work fine on my machine and as far as i can tell should be cross platform

sfod: Yep, `CMAKE_CXX_COMPILER_LOADED` is [available since at least CMake 3.0](https://github.com/Kitware/CMake/blob/v3.0.0/Modules/CMakeCXXCompiler.cmake.in#L13) and I don't see any platform specifics there.

One thing though. It is a global variable, so when a C++ project uses a C project via `add_subdirectory`, `CMAKE_CXX_COMPILER_LOADED` will be defined for the C project. For example, `aws-c-common` used by `aws-iot-device-sdk-cpp-v2` sees C++ enabled:

```
$ cmake -S aws-iot-device-sdk-cpp-v2/ -B cmake-build-tsan -DENABLE_SANITIZERS=ON -DSANITIZERS=thread
...
- aws-c-common CMakeLists.txt: calling aws_add_sanitizers
- aws_check_sanitizer: CMAKE_CXX_COMPILER_LOADED is defined
- aws_check_sanitizer: CMAKE_C_COMPILER_LOADED is defined
...
```

I think, this shouldn't cause any issues in practice, but I just don't feel comfortable with checking C++ compiler flags in a C project.

That said, if you're strongly against adding `IS_CXX` option, I'll use `CMAKE_CXX_COMPILER_LOADED`. Or, there is also [ENABLED_LANGUAGES](https://cmake.org/cmake/help/latest/prop_gbl/ENABLED_LANGUAGES.html) property, which is also suitable here and it's documented.

sfod: Thinking more about this, checking C++ compiler in C subprojects can't cause any issues, because C++ compiler should be checked anyway. So yeah, I'll just add a check for CMAKE_CXX_COMPILER_LOADED.

PR #1138, opened 2024-07-18
Add no copy api variants to json interface

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*
There are several json apis that currently take key by cursor, including some heavily used apis like get_object or has_object. Since underlying lib relies on const char, that means that every one of those apis needs to make a temp copy of the key. In grand scheme of things copies are cheap, but for heavy users of json (ex. endpoint resolution) cost ends up quickly. 

This PR adds const char variants for all functions taking a key to avoid a copy. And as a side bonus you can now just write get_object(json, "field_name"), which is cleaner in my opinion than having to wrap it in cursor.

Can we just make cursor version not do a copy? Not without major surgery of cJson, which internally relies on c strings a lot. 

### DIFF
--- a/include/aws/common/json.h
+++ b/include/aws/common/json.h
@@ -23,12 +23,24 @@ AWS_EXTERN_C_BEGIN
  *
  * Note: You will need to free the memory for the aws_json_value using aws_json_destroy on the aws_json_value or
  * on the object/array containing the aws_json_value.
- * @param string A byte pointer to the string you want to store in the aws_json_value
+ * @param string A byte cursor you want to store in the aws_json_value
  * @param allocator The allocator to use when creating the value
  * @return A new string aws_json_value
  */
 AWS_COMMON_API
 struct aws_json_value *aws_json_value_new_string(struct aws_allocator *allocator, struct aws_byte_cursor string);
+
+/**
+ * Creates a new string aws_json_value with the given string and returns a pointer to it.
+ *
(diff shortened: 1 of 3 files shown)


source/json.c
```diff
@@ -104,6 +110,25 @@ int aws_json_value_add_to_object(
     return result;
 }
 
+int aws_json_value_add_to_object_c_str(struct aws_json_value *object, const char *key, struct aws_json_value *value) {
```
waahm7, line 113: trivial: we can make the old functions call new functions to avoid code duplication.

include/aws/common/json.h
```diff
@@ -23,13 +23,25 @@ AWS_EXTERN_C_BEGIN
  *
  * Note: You will need to free the memory for the aws_json_value using aws_json_destroy on the aws_json_value or
  * on the object/array containing the aws_json_value.
- * @param string A byte pointer to the string you want to store in the aws_json_value
+ * @param string A byte cursor you want to store in the aws_json_value
```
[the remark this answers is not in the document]
DmitriyMusatkin: added a generic comment that it can be slower. we try to hide implementation details, since this header is a kind of interface. but in all likely hood we are not going to change underlying impl any time soon

source/json.c
DmitriyMusatkin: good point. forgot about updating the old functions. done now.

waahm7 (approved):

PR #1141, opened 2024-07-31
New Get_ENV Functions

waahm7:
*Description of changes:*
`aws_get_environment_value` is error-prone because one of the frequent use cases is to check an environment variable, else check other source like config file. In that case, the previous implementation required us to check three cases:
  - NULL -> check the alternative
  - Empty -> destroy and check the alternative
  - Available -> use that
  
<del>Since we don't need to differentiate between empty values and unset values, we treat empty environment values as unset values. I have added two new functions `aws_get_env` and `aws_get_env_nonempty` for these use cases. They directly take char* so that we don't need to needlessly allocate a string in some cases.</del>

Introduce two new functions `aws_get_env` and `aws_get_env_nonempty` to improve these functions for both use-cases where we may or may not care about empty env variable. These functions take `char *` so that we don't need to always allocate a string.


bretambrose:
>       Since we don't need to differentiate between empty values and unset values, treat empty environment values as unset.
> 

This is a pretty big assumption since you're not only asserting it for right now, but for everything that comes in the future too.  I don't really agree with this.

source/posix/environment.c
```diff
@@ -14,7 +14,7 @@ int aws_get_environment_value(
     struct aws_string **value_out) {
 
     const char *value = getenv(aws_string_c_str(variable_name));
-    if (value == NULL) {
+    if (value == NULL || value[0] == '\0') {
```
graebm, line 17: It sucks, but this is how it is with environment variables. Sometimes just the existence of the env-var is meaningful, even if it's blank. Maybe create an alternate function like `aws_get_environment_variable_nonempty()` that has this behavior, and scrub the codebase for places that would benefit from using that instead

graebm (changes requested):

source/posix/environment.c
waahm7: Yeah, I agree. We can discuss the following options in the office tomorrow.

1. Modify the existing function, and we can create a new function in the future if required, like if we have a use case where `env-var` can be blank.

2. Create a new function, `aws_get_environment_variable_nonempty`, `aws_get_environment_variable_cursor`, or `aws_get_environment_variable_c_str`. I prefer the `aws_get_environment_variable_cursor` if we decide to just create a new function.

DmitriyMusatkin (approved):

PR #1143, opened 2024-08-06
Always set the symbol visibility preset to hidden.

### LINKED ISSUE #1142: Consider enabling `-fvisibility=hidden` when building a static library.
### Describe the feature

When building `aws-c-common` (and the other AWS C libraries), we pass `-fvisibility=hidden`, but only when building a shared library (since #516). I am proposing to pass it always.

### Use Case

Compiling with `-fvisibility=hidden` makes sense in static libraries too (see [related CMake policy](https://cmake.org/cmake/help/latest/policy/CMP0063.html)). There has been [an issue](https://github.com/apache/arrow/issues/42154) where the interaction between two Python wheels with their own statically linked copy of the AWS SDK resulted in unexplained failures, because both AWS SDKs exported private symbols, which was fixed by forcing the AWS libraries in one of the wheels to compile with `-fvisiblity=hidden`.

The AWS SDK is already [going to always enable](https://github.com/aws/aws-sdk-cpp/blob/878d0c1ca7a0f068dc03e90fde600e5424bf11d9/CMakeLists.txt#L372-L374) `-fvisiblity=hidden` and so does s2n for non-testing builds. This issue tracks doing the same for the AWS C libraries.

### Proposed Solution

Revert to the behavior before #516, or follow s2n's example and pass `-fvisibility=hidden` always unless we have enabled testing and are building a shared library.

### Other Information

_No response_

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

teo-tsirpanis:
*Issue #, if available:* Fixes #1142

*Description of changes:*

This PR updates `AwsCFlags.cmake` to always pass `-fvisibility=hidden` to the targets, even when we are building static libraries. Instead of directly specifying the compiler option, we use the [`VISIBILITY_PRESET`](https://cmake.org/cmake/help/latest/prop_tgt/LANG_VISIBILITY_PRESET.html) CMake property, for increased portability across compilers.


graebm:
Sorry for the silence. I've been chipping away at this
- I noticed this change caused CMake to spit out a warning about [cmake-policy CMP0063](https://cmake.org/cmake/help/latest/policy/CMP0063.html).
- So I made my own [branch](https://github.com/awslabs/aws-c-common/compare/15a9e48ea674c2ff87161c69af5f435137baace0...visibility-hidden) which did set that policy, but we'd need to update ALL repos (aws-c-http, aws-c-s3, there are more than a dozen others...) to set that policy as well.
- Which made the team wonder why we didn't just update the minimum CMake version, so we wouldn't need to manually set that policy. Updating CMake is something we've wanted to do for years, so now I'm testing that out in the [update-cmake](https://github.com/awslabs/aws-c-common/tree/update-cmake) branch

waahm7:
This was fixed in https://github.com/awslabs/aws-c-common/releases/tag/v0.10.0.

PR #1147, opened 2024-08-22
aws_explicit_aligned_allocator_new

TingDaoK:
*Issue #, if available:*

- Add `aws_explicit_aligned_allocator_new` to support customized aligned instead of the default one.

*Description of changes:*


include/aws/common/allocator.h
```diff
@@ -228,6 +228,21 @@ size_t aws_small_block_allocator_page_size(struct aws_allocator *sba_allocator);
 AWS_COMMON_API
 size_t aws_small_block_allocator_page_size_available(struct aws_allocator *sba_allocator);
 
+/*
+ * Create an aligned allocator with customized alignment.
+ */
+AWS_COMMON_API
+struct aws_allocator *aws_customized_aligned_allocator_new(
```
DmitriyMusatkin, line 235: maybe just aws_custom_alignment_allocator_new?

source/allocator.c
```diff
@@ -34,6 +34,11 @@ bool aws_allocator_is_valid(const struct aws_allocator *alloc) {
     return alloc && AWS_OBJECT_PTR_IS_READABLE(alloc) && alloc->mem_acquire && alloc->mem_release;
 }
 
+struct aws_aligned_allocator_impl {
+    struct aws_allocator *allocator;
+    size_t alignment_size;
+};
+
 static void *s_aligned_malloc(struct aws_allocator *allocator, size_t size) {
     (void)allocator;
```
DmitriyMusatkin, line 43: probably not needed anymore?

source/allocator.c
```diff
@@ -46,9 +51,15 @@ static void *s_aligned_malloc(struct aws_allocator *allocator, size_t size) {
...
      *
      * We use PAGE_SIZE as the boundary because we are not aware of any allocations of
-     * this size or greater that are not data buffers
+     * this size or greater that are not data buffers.
+     *
+     * Unless there is a customized alignment size.
      */
-    const size_t alignment = sizeof(void *) * (size > (size_t)PAGE_SIZE ? 8 : 2);
+    size_t alignment = sizeof(void *) * (size > (size_t)PAGE_SIZE ? 8 : 2);
+    if (allocator->impl) {
```
DmitriyMusatkin, line 59: allocator will not have impl if its aligned by default?

DmitriyMusatkin: dumb idea, but maybe just let impl be the actual alignment (i.e. just cast impl pointer to size_t and back) instead of allocating another chunk of mem for it?

include/aws/common/allocator.h
graebm: I assume it's named this way because the nice simple name `aws_aligned_allocator()` is already taken, but that one automagically determines the alignment, and it differs based on allocation size.

Move these new functions to be right next to the existing `aws_aligned_allocator()` (instead of the end of the file) so that someone searching "align" is more likely to see both options and pick the right one.

I like "explicit" more than "custom". "custom" sounds like "you can do whatever you want like via a callback" while "explicit" sounds to me like "we simply use this 1 number you give us".

I'd suggest `aws_explicit_aligned_allocator()` (sounds like the pre-existing allocator) or `aws_explicit_alignment_allocator()` (nicer sounding gramatically)

source/allocator.c
```diff
@@ -34,6 +34,11 @@ bool aws_allocator_is_valid(const struct aws_allocator *alloc) {
     return alloc && AWS_OBJECT_PTR_IS_READABLE(alloc) && alloc->mem_acquire && alloc->mem_release;
 }
 
+struct aws_aligned_allocator_impl {
+    struct aws_allocator *allocator;
+    size_t alignment_size;
```
graebm, line 39: trivial:
```suggestion
    size_t alignment;
```

source/allocator.c
```diff
@@ -34,6 +34,11 @@ bool aws_allocator_is_valid(const struct aws_allocator *alloc) {
     return alloc && AWS_OBJECT_PTR_IS_READABLE(alloc) && alloc->mem_acquire && alloc->mem_release;
 }
 
+struct aws_aligned_allocator_impl {
```
graebm, line 37: use "explicit" or "custom" in this struct name (match the public function name)

Otherwise, one would assume this is for `aws_aligned_allocator()`

Which is extra confusing because both share the s_aligned_malloc() call, and it's tough to explain "you can tell that it's NOT the `aws_aligned_allocator()` because it uses the `aws_aligned_allocator_impl`" 🤯🤯🤯

TingDaoK: Go with @DmitriyMusatkin 's suggestion to remove the whole impl, and use the point address to store the size_t for the alignment.

DmitriyMusatkin (approved):

Issue #1155, opened 2024-10-11
When building on Windows w/swiftc, can accidently include both msvc and gcc math.inls.


jeffdav:
### Describe the bug

On Windows, I'm building `aws-sdk-swift` using CMake.  I build `aws-c-*` using `ExternalProject_Add()`. 
(Thank you for providing `CMakeLists.txt` for your projects!)

The issue is when building `aws-sdk-swift`, we end up using clang, which causes `math.h` include both `math.msvc.inl` and `math.gcc_builtin.inl` during module import, leading to duplicate symbols.

### Expected Behavior

We should only include one or the other. 

### Current Behavior

We include both.

### Reproduction Steps

Build `aws-sdk-swift` using CMake on Windows.  I'm happy to share my `CMakeLists.txt` for doing this if it is helpful.  

### Possible Solution

Currently I'm patching:

```c
#if defined(__clang__) || defined(__GNUC__)
```

To:
```c
#if (defined(__clang__) || defined(__GNUC__)) && !defined(AWS_HAVE_MSVC_INTRINSICS_X64)
#    include <aws/common/math.gcc_builtin.inl>
#endif
```

I'm not 100% certain this is the most correct, but it works.

### aws-c-common version used

commit 69789467ebb76526bffc8ad0e552858256b6a326 (origin/main, origin/HEAD, main)

### Compiler and version used

swiftc: compnerd.org Swift version 6.0-dev (LLVM a527f49f462b0d7, Swift eef85a7be270421), clang: compnerd.org clang version 17.0.6, cl: Microsoft (R) C/C++ Optimizing Compiler Version 19.41.34120 for x64

### Operating System and version

Windows 11 23H2

TingDaoK:
Yeah, please share the CMakeLists, which will help us to reproduce the issue.

jmklix:
Please note that it is still a feature request to get swiftc working on windows. As mentioned [here](https://github.com/awslabs/aws-crt-swift/issues/289#issuecomment-2391869469) by @waahm7
> Currently, both Aws-crt-swift and Aws-sdk-swift don't support Windows. This is a feature request, and I think we will probably hit more issues than just https://github.com/awslabs/aws-crt-swift/pull/290 to properly support Windows and also add CI for it. Swift SDK might have to do some work as well on their side. We have added this to our backlog and will update here as soon as we have any updates.
>
> Please 👍 the original issue and that will help us prioritize this.

jjerphan:
[...]

PR #1156, opened 2024-10-11
Don't accidently include both math.msvc.inl and math.gcc_builtin.inl.

jeffdav:
*Issue #, if available:* #1155

*Description of changes:*
Add additional check to make sure we're not including both `math.msvc.inl` and `math.gcc_builtin.inl`.

Note: I'm not 100% sure this is the most correct solution, so I'd love feedback and I'm happy to iterate.

### DIFF
--- a/include/aws/common/math.inl
+++ b/include/aws/common/math.inl
@@ -44,7 +44,7 @@ AWS_EXTERN_C_BEGIN
 #    endif /*  AWS_HAVE_GCC_OVERFLOW_MATH_EXTENSIONS */
 #endif     /*  defined(AWS_HAVE_GCC_OVERFLOW_MATH_EXTENSIONS) && (defined(__clang__) || !defined(__cplusplus)) */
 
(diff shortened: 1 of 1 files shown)


jeffdav:
I agree.  I think the problem here is it gets built with MSVC, because I use your CMakeLists via ExternalProject.  But we're not using the C modules directly; we're using [`AwsSdkSwift`](https://github.com/awslabs/aws-sdk-swift).  And when I build that with CMake as part of the rest of the project the headers get imported as a module by clang, and that's where I start getting duplicate symbol errors.

waahm7:
Not sure I fully understand. The problem is that it should be built with either MSVC or Clang. If both MSVC and Clang are set, everything will break, and this won't be the only issue. I am closing this in favor of the Windows support feature request at https://github.com/awslabs/aws-crt-swift/issues/289, since there will likely be many more problems with Swift + Windows. We can tackle all of them as part of official Windows support. Please feel free to reopen or create another issue.

PR #1159, opened 2024-10-17
Update CMake to 3.9

waahm7:
*Issue #, if available:*
- https://github.com/awslabs/aws-c-common/issues/1142

*Description of changes:*
- set visibility always to hidden 
- Update CMake to 3.9


cmake/AwsCFlags.cmake
```diff
@@ -231,7 +231,6 @@ function(aws_set_common_properties target)
 
         # try to check whether compiler supports LTO/IPO
         if (POLICY CMP0069)
```
graebm, line 233: I know we're not trying to rewrite the world, but since this touches a POLICY ... we can replace this ~if (POLICY CMP0069)~ block with simpler code
```
        # check whether compiler supports LTO/IPO
        include(CheckIPOSupported)
        check_ipo_supported(RESULT ipo_supported)
        if (ipo_supported)
            message(STATUS "Enabling IPO/LTO for Release builds")
        else()
            message(STATUS "AWS_ENABLE_LTO is enabled, but cmake/compiler does not support it, disabling")
            set(_ENABLE_LTO_EXPR OFF)
        endif()
```
see: https://cmake.org/cmake/help/v3.9/policy/CMP0069.html

CMakeLists.txt
```diff
@@ -1,17 +1,13 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0.
 
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.9)
```
graebm, line 4: Add comment about WHY we're picking CMake 3.9. Because it's the latest supported version in  ... RHEL5? Amazon Linux 1?

graebm (approved):

PR #1161, opened 2024-10-23
Forward CMake variables to prebuilding dependencies

sfod:
*Issue #, if available:*

When prebuilding dependencies, we need to pass certain CMake variables from parent cmake process to child cmake process (e.g. toolchain file or path to C compiler). CMake doesn't provide this functionality out of the box.

*Description of changes:*

This PR implements a new CMake function that checks CMake cache file and collects variables that may affect build process. `aws_prebuild_dependency` uses this new function, so no changes are required in the downstream projects.


cmake/AwsGetCmdArguments.cmake
```diff
@@ -0,0 +1,49 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0.
+
+# Get list of variables provided via command line arguments (i.e. -DVARIABLE=VALUE). This function iterates over all
+# variables and determines if they're provided via command line or not by checking their help strings. CMake sets help
+# strings for variables provided via command line to the "No help, variable specified on the command line." phrase since
+# at least v3.0.
+# Via https://cmake.org/pipermail/cmake/2018-January/067002.html
```
graebm, line 8: IMO:
- Move this to `AwsPrebuildDependency.cmake`
- name it something like `aws_get_cmd_args_for_prebuild_dependency()`
- don't bother taking arguments, just hardcode the CMAKE_INSTALL_PATH exception

You've done great working getting this to work, and it's noble to try and make it a generic helper, but let's be honest, all this complexity is an abomination, hoisted upon us because we can't use [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) from CMake 3.11, because we're stuck on 3.9 until we can drop support for RHEL5 (very extremely likely next year)

graebm: Well, my rant about how FetchContent would solve all our problems was wrong.
Turns out FetchContent won't "prebuild" AWS-LC before configuring S2n
Maybe you explained this to me recently, and then I forgot?

cmake/AwsGetCmdArguments.cmake
```diff
@@ -0,0 +1,49 @@
...
+# Arguments:
+#  VARS_TO_IGNORE Variables that should be ignored even if they were provided via command line. Multiple variables can
+#                 be specified separated by space.
+#
+# MUST be done before call to 'project'. The reason is that project() resets help strings for some variables
```
graebm, line 13: How would the requirement to call it before `project()` work with aws-sdk-cpp, which builds CRT via [add_subdirectory(crt/aws-crt-cpp)
](https://github.com/aws/aws-sdk-cpp/blob/4b5c0cf73ef3929858bfe7460bc72983a1dc98a3/CMakeLists.txt#L231)

cmake/AwsGetCmdArguments.cmake
graebm, line 1: What about taking a completely different approach for aws-crt-java, python, nodejs, etc where we're invoking cmake from some other script within the project?

Stop trying to have the one CMakeLists.txt try to re-call cmake on aws-lc and s2n.
Instead, just have the pom.xml (for aws-crt-java), or setup.py (for aws-crt-python), call cmake on aws-lc, then on s2n, then on the one CMakeLists.txt that builds the rest of the aws-c-***.

So we don't have to try and figure out what arguments were passed from within a .cmake script. We'd just pass the same arguments each time we called cmake

graebm: aws-crt-cpp would be the odd man out, since that's one where we ask users to invoke CMake, and it would be a breaking change if it didn't just work via `add_subdirectory()` like it used to. But also, that's the one repo that's not going to work well with this new function, that has to be invoked before `project()`

sfod: I really liked the idea. Unfortunately, pom.xml is not used on ARM and Android. So, cross-compiling for ARM still needs gathering command line arguments.

cmake/AwsGetCmdArguments.cmake
sfod: The most important variables are set explicitly in `aws_prebuild_dependency`. This PR adds `CMAKE_TOOLCHAIN_FILE`, which should handle the simplest cross-compiling use cases even without forwarding command-line arguments.

If we had to deal only with variables listed here (https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html), we could probably just check them all inside `aws_prebuild_dependency`. But various platforms can add custom variables.
For example, `gradle` in the Android CI job in this repo passes multiple `ANDROID_*` variables (described [here](https://developer.android.com/ndk/guides/cmake#variables)) via command line (that's where this "gather cmd vars" idea came from).

sfod: I removed parsing cmd vars

cmake/AwsPrebuildDependency.cmake
```diff
@@ -75,3 +92,30 @@ function(aws_prebuild_dependency)
...
+function(aws_get_variables_for_prebuild_dependency AWS_CMAKE_PREBUILD_ARGS)
+    set(variables "")
+    set(variablesToIgnore CMAKE_INSTALL_PREFIX)
+
+    get_cmake_property(vars CACHE_VARIABLES)
+    foreach(var ${vars})
+        message("= Checking ${var}")
+        if (var MATCHES "^(CMAKE_)?ANDROID_"
+                OR var STREQUAL "CMAKE_TOOLCHAIN_FILE"
+                OR var STREQUAL "CMAKE_SYSTEM_NAME"
```
graebm, line 106: in a normal build, is CMAKE_SYSTEM_NAME in the cache? and therefore always passed to the prebuild?

I see that this affects [CMAKE_CROSSCOMPILING](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html)
> This variable will be set to true by CMake if the CMAKE_SYSTEM_NAME variable has been set manually (i.e. in a toolchain file or as a cache entry from the cmake command line). In most cases, manually setting CMAKE_SYSTEM_NAME will only be done when cross compiling since, if not manually set, it will be given the same value as CMAKE_HOST_SYSTEM_NAME, which is correct for the non-cross-compiling case. In the event that CMAKE_SYSTEM_NAME is manually set to the same value as CMAKE_HOST_SYSTEM_NAME, then CMAKE_CROSSCOMPILING will still be set to true.

Is that OK?

Are there other weird side-effects to worry about, due to manually passing stuff in the prebuild, that got into the main build's cache automatically?

sfod: I checked on Linux and Mac, `CMAKE_SYSTEM_NAME` is not set.
However, to play it safe, I'll add a `CMAKE_CROSSCOMPILING` check along with `CMAKE_SYSTEM_NAME`.

sfod: OK, I put all vars that make sense only in cross-compiling mode to a separate `if` block checking additionally for `CMAKE_CROSSCOMPILING` var.

cmake/AwsPrebuildDependency.cmake
```diff
@@ -75,3 +92,52 @@ function(aws_prebuild_dependency)
...
+# Get list of optional or platform-specific variables that may affect build process.
+function(aws_get_variables_for_prebuild_dependency AWS_CMAKE_PREBUILD_ARGS)
+    set(variables "")
+    set(variablesToIgnore CMAKE_INSTALL_PREFIX)
+
+    # The CMake variables below were chosen for Linux, BSD, and Android platforms. If you want to use the prebuild logic
+    # on other platforms, the chances are you have to handle additional variables (like CMAKE_OSX_SYSROOT). Refer to
+    # https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html to update the list of handled variables, and
+    # then you can enable a new platform here.
+    if (NOT UNIX OR APPLE)
```
graebm, line 105: trivial: add parens to make this more clear
```suggestion
    if ((NOT UNIX) OR APPLE)
```

cmake/AwsPrebuildDependency.cmake
```diff
@@ -75,3 +92,52 @@ function(aws_prebuild_dependency)
         DESTINATION ${CMAKE_INSTALL_PREFIX}
     )
 endfunction()
+
+# Get list of optional or platform-specific variables that may affect build process.
+function(aws_get_variables_for_prebuild_dependency AWS_CMAKE_PREBUILD_ARGS)
+    set(variables "")
+    set(variablesToIgnore CMAKE_INSTALL_PREFIX)
```
graebm, line 99: trivial: I don't see `variablesToIgnore` being used anywhere. Should it be deleted, or did we forget to do a check later?

cmake/AwsPrebuildDependency.cmake
```diff
@@ -75,3 +92,52 @@ function(aws_prebuild_dependency)
...
+
+    # The CMake variables below were chosen for Linux, BSD, and Android platforms. If you want to use the prebuild logic
+    # on other platforms, the chances are you have to handle additional variables (like CMAKE_OSX_SYSROOT). Refer to
+    # https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html to update the list of handled variables, and
```
graebm, line 103: 👍 great docs on the thinking behind this, and what someone would need to look into if they're going to modify it in the future

cmake/AwsPrebuildDependency.cmake
```diff
@@ -75,3 +92,52 @@ function(aws_prebuild_dependency)
...
+    # then you can enable a new platform here.
+    if (NOT UNIX OR APPLE)
+        message(FATAL_ERROR "aws_get_variables_for_prebuild_dependency is called for unsupported platform")
+    endif()
+
+    get_cmake_property(vars CACHE_VARIABLES)
+    foreach(var ${vars})
+        # Variables in this block make sense only in cross-compiling mode. The variable list is created from the CMake
+        # documentation on toolchains: https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html
+        # NOTE: Some variables are missed here (e.g. CMAKE_SYSROOT) because they can be set via toolchain file only.
```
graebm, line 113: 👍 again great docs

cmake/AwsPrebuildDependency.cmake
```diff
@@ -75,3 +92,52 @@ function(aws_prebuild_dependency)
...
+            set(escaped_var ${${var}})
+            # To store a list within another list, it needs to be escaped first.
+            string(REPLACE ";" "\\\\;" escapedVar "${${var}}")
+            list(APPEND variables "-D${var}=${escapedVar}")
+        endif()
+
+        # Other optional variables applicable both in cross-compiling and non-cross-compiling modes.
+        if (var STREQUAL "CMAKE_C_COMPILER"
+                OR var STREQUAL "CMAKE_CXX_COMPILER"
+                OR var STREQUAL "CMAKE_C_FLAGS"
```
graebm, line 130: Does AWS-LC use C++ anywhere? If so, we also need CMAKE_CXX_FLAGS. If not, maybe remove CMAKE_CXX_COMPILER?

graebm: Oh man. Should we also worry about linker flags? https://cmake.org/cmake/help/latest/envvar/LDFLAGS.html

graebm: Lots of food for thought in this doc:
https://cmake.org/cmake/help/latest/manual/cmake-env-variables.7.html

graebm (approved):
Looking pretty good.
There might be a few more things we should look into (mentioned in comments), but when you're happy with it, I'm happy with it.

cmake/AwsPrebuildDependency.cmake
sfod: Yep, it's a remaining of the previous attempts to make this more generic, removed it

cmake/AwsPrebuildDependency.cmake
sfod: Good catch with linker flags, I added them.
As for env vars, a cmake process called via `execute_process` sees them all (at least on Linux machines). afaics the usual logic for env vars in cmake is to initialize a property or another variable from them (e.g. LDFLAGS sets `CMAKE_*_LINKER_FLAGS_INIT`). So, we either pass these variables explicitly or cmake will initialize them from env vars.
I hope, I'm not missing anything.

graebm (approved):

Issue #1165, opened 2024-11-07
xml_parser does not support empty tags


rvansa:
### Describe the bug

I am trying to connect using `aws-c-s3` to a localstack instance. On listing objects in a bucket localstack replies with
```
<?xml version='1.0' encoding='utf-8'?>
<ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01\">
  <IsTruncated>false</IsTruncated>
  <Name>foo</Name>
  <Prefix />
  <MaxKeys>1000</MaxKeys>
  <EncodingType>url</EncodingType>
  <KeyCount>1</KeyCount>
  <Contents> (redacted) </Contents>
</ListBucketResult>
```
but xml_parser is unable to parse empty tag `<Prefix />`: it ignores that it is an empty tag and looks for `</Prefix>`, failing to parse the document.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Valid XML is parsed.

### Current Behavior

Parser logs `XML document is invalid.` and returns error.

### Reproduction Steps

Try to parse
```
<?xml version='1.0' encoding='utf-8'?>
<foo><bar /></foo>
```

### aws-c-common version used

v0.9.27

### Compiler and version used

GCC 11.2

### Operating System and version

Linux

jmklix:
Thanks for finding this bug. We have a PR that should fix this: https://github.com/awslabs/aws-c-common/pull/1168

PR #1168, opened 2024-11-12
fix empty xml node handling

DmitriyMusatkin:
*Issue #, if available:*
https://github.com/awslabs/aws-c-common/issues/1165

*Description of changes:*

### DIFF
--- a/tests/xml_parser_test.c
+++ b/tests/xml_parser_test.c
@@ -173,6 +173,8 @@ static int s_xml_parser_child_empty_tag(struct aws_allocator *allocator, void *c
     struct sibling_text_capture capture;
     AWS_ZERO_STRUCT(capture);
 
+    capture.capture1 = aws_byte_cursor_from_c_str("random values");
+
     struct aws_xml_parser_options options = {
         .doc = aws_byte_cursor_from_c_str(siblings_with_empty_tag),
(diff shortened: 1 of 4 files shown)


tests/xml_parser_test.c
```diff
@@ -165,6 +165,40 @@ static int s_xml_parser_siblings_with_text_test(struct aws_allocator *allocator,
 
 AWS_TEST_CASE(xml_parser_siblings_with_text, s_xml_parser_siblings_with_text_test)
 
+const char *siblings_with_empty_tag = "<?xml version=\"1.0\" "
+                                      "encoding=\"UTF-8\"?><rootNode><child1 /><child2>TestBody2</child2></rootNode>";
+
+static int s_xml_parser_child_empty_tag(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+    struct sibling_text_capture capture;
+    AWS_ZERO_STRUCT(capture);
```
TingDaoK, line 174: trivial: instead of zero out the `capture.capture1`, set it to some garbage so that we can test that the parse sets the outcome to empty cursor correctly.

TingDaoK (approved):

tests/xml_parser_test.c
DmitriyMusatkin: ah, good point

PR #1170, opened 2024-11-21
Remove reliance on hardcoded user in ci

DmitriyMusatkin:
*Issue #, if available:*

*Description of changes:*


.github/workflows/ci.yml
```diff
@@ -6,19 +6,18 @@ on:
...
-  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-  AWS_REGION: us-east-1
+  CRT_CI_ROLE: ${{ secrets.CRT_CI_ROLE_ARN }}
+  AWS_DEFAULT_REGION: us-east-1
 
 jobs:
   linux-compat:
-    runs-on: ubuntu-20.04 # latest
+    runs-on: ubuntu-22.04 # latest
```
waahm7, line 20: Why not 24?

waahm7 (approved):

.github/workflows/ci.yml
DmitriyMusatkin: some of the older compilers are not available on 24 (clang 3 and gcc 4.8 if i remember correctly). dropping those from our ci is a bridge we can cross later

waahm7: BTW, Does this version of Ubuntu matter since our Docker base image is Ubuntu 18?

DmitriyMusatkin: nope: should not matter too much (i think our 18 ubuntu image might be actually 20, but i cant find actual docker file)
This is just a prophylactic image bump.

DmitriyMusatkin: actually you are right. it should not affect the compiler cause underlying image is old. let me retest. i saw old compilers failing when i moved to 24, so i assumed its related, but maybe it is not.

.github/workflows/ci.yml
```diff
@@ -6,19 +6,21 @@ on:
       - 'main'
 
 env:
-  BUILDER_VERSION: v0.9.62
```
graebm, line 9: utterly trivial: my vscode user setting have `"files.trimTrailingWhitespace": true`

tests/hash_table_test.c
```diff
@@ -1084,7 +1084,7 @@ static int s_test_hash_churn_fn(struct aws_allocator *allocator, void *ctx) {
             int was_present;
             err_code = aws_hash_table_remove(&hash_table, e->key, NULL, &was_present);
             ASSERT_SUCCESS(err_code, "Unexpected failure removing element");
-            if (i == 0 && entries[i - 1].key == e->key && entries[i - 1].is_removed) {
+            if (i != 0 && entries[i - 1].key == e->key && entries[i - 1].is_removed) {
```
graebm, line 1087: oh man ... this test
I don't know what it's trying to do
but a few lines up it's calling [rand()](https://en.cppreference.com/w/c/numeric/random/rand) which returns a random **integer** and checking whether it's `< 0.05` (floating point) 😬. I think they expect that to pass 5.0% of the time, but in reality it's going to pass 0.003% of the time.

.github/workflows/ci.yml
```diff
@@ -33,15 +35,18 @@ jobs:
           - rhel8-x64
           - al2-x64
     steps:
-        # We can't use the `uses: docker://image` version yet, GitHub lacks authentication for actions -> packages
+    - uses: aws-actions/configure-aws-credentials@v4
+      with:
+        role-to-assume: ${{ env.CRT_CI_ROLE }}
+        aws-region: ${{ env.AWS_DEFAULT_REGION }}
```
graebm, line 41: you'd mentioned in-office having some shared code in aws-crt-builder

did you decide to just copy/paste the same 4 lines instead of doing that?

I'm ok with either, just asking

tests/hash_table_test.c
graebm: OK, I fixed the test in a separate PR: https://github.com/awslabs/aws-c-common/pull/1172

graebm (commented):
shipit 👍

PR #1176, opened 2024-12-10
Fix dependency build failure on old cmake versions

sfod:
*Issue #, if available:*

aws-lc can't be prebuilt on Ubuntu 18: https://github.com/awslabs/aws-crt-cpp/issues/691

*Description of changes:*

Apparently, CMake prior to v3.13 requires that the `-B<BUILD_DIRECTORY>` option should come after the source directory argument:
```
cmake source_dir/ -Bbuild_dir
```

### DIFF
--- a/cmake/AwsPrebuildDependency.cmake
+++ b/cmake/AwsPrebuildDependency.cmake
@@ -36,7 +36,7 @@ function(aws_prebuild_dependency)
     list(APPEND cmakeCommand ${cmakeOptionalVariables})
 
     # The following variables should always be used.
-    list(APPEND cmakeCommand ${AWS_PREBUILD_SOURCE_DIR})
+    list(APPEND cmakeCommand -H${AWS_PREBUILD_SOURCE_DIR})
     list(APPEND cmakeCommand -B${depBinaryDir})
     list(APPEND cmakeCommand -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE})
     list(APPEND cmakeCommand -DCMAKE_PREFIX_PATH=${ESCAPED_PREFIX_PATH})
--- a/cmake/AwsPrebuildDependency.cmake
+++ b/cmake/AwsPrebuildDependency.cmake
@@ -36,8 +36,8 @@ function(aws_prebuild_dependency)
     list(APPEND cmakeCommand ${cmakeOptionalVariables})
 
     # The following variables should always be used.
-    list(APPEND cmakeCommand -B${depBinaryDir})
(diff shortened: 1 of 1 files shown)


cmake/AwsPrebuildDependency.cmake
```diff
@@ -36,8 +36,8 @@ function(aws_prebuild_dependency)
     list(APPEND cmakeCommand ${cmakeOptionalVariables})
 
     # The following variables should always be used.
-    list(APPEND cmakeCommand -B${depBinaryDir})
     list(APPEND cmakeCommand ${AWS_PREBUILD_SOURCE_DIR})
```
graebm, line 39: debatable: 
```suggestion
    list(APPEND cmakeCommand -H${AWS_PREBUILD_SOURCE_DIR})
```

So, got the idea to use the undocumented flag from aws-crt-builder, which sets both `-B<build-dir> -` **and** `-H<source-dir>` ([see here](https://github.com/awslabs/aws-crt-builder/blob/42d15ba4ee3fbe0e812e25094a8cf59eaef0cf1e/builder/actions/cmake.py#L145-L147))

But I only used `-B` because I thought "better to only use 1 undocumented flag than 2". But maybe it's better to just use both and be explicit about what the arg is

This didn't work in 3.9 because the [argument parsing loop](https://github.com/Kitware/CMake/blob/v3.9.0/Source/cmake.cxx#L761-L765) was like "this arg doesn't have a flag, so assume they're setting source dir, and configure all directories the old fashioned way (which oops I blows away any previously configured build dir)" It was fixed by [this change in 3.13](https://github.com/Kitware/CMake/commit/a10d63d5783275f9972aaa86d41071a1ddca7921#diff-9451a015783be6eaba0a5a470186420c922d536fc465b80a4cdba55dec61d749R861-R866), where they make sure not to reset the build dir if it was already set by another flag

Maybe better to explicitly set `-H` which is treated the same as the documented `-S` option [even today](https://github.com/Kitware/CMake/blob/ed0baf1ec5151ab05d41e0dde652abe54a067737/Source/cmake.cxx#L993-L999). So we can avoid that old fashioned [SetDirectoriesFromFile(String)](https://github.com/Kitware/CMake/blob/ed0baf1ec5151ab05d41e0dde652abe54a067737/Source/cmake.cxx#L1774-L1775) function that tries to guess a lot of things.

graebm (approved):

cmake/AwsPrebuildDependency.cmake
sfod: Yep, you're right, here's the line where they reset the build directory: https://github.com/Kitware/CMake/blob/v3.12.4/Source/cmake.cxx#L851

Issue #1117, opened 2024-05-20
Improvement to build, install, and distribution mechanisms


jmklix:
### Describe the feature

CRT tracking issue for all issues/PRs related to cmake changes. These changes will be done in parallel(or after) cmake changes to [aws-sdk-cpp](https://github.com/aws/aws-sdk-cpp/issues/1888)

### Use Case

- https://github.com/awslabs/aws-c-common/issues/844
- https://github.com/awslabs/aws-c-common/pull/792
- https://github.com/awslabs/aws-c-mqtt/pull/203
- https://github.com/awslabs/aws-c-common/pull/850
- https://github.com/awslabs/aws-crt-cpp/pull/325
- https://github.com/awslabs/aws-crt-cpp/issues/606

### Acknowledgements

- [ ] I may be able to implement this feature request
- [X] This feature might incur a breaking change

PR #1178, opened 2024-12-19
A bunch of CMake fixes

graebm:
**Issue #:**
- https://github.com/awslabs/aws-c-common/issues/1117
- https://github.com/awslabs/aws-crt-cpp/issues/606
- Install CMake modules in a more standard way
  - https://github.com/awslabs/aws-c-event-stream/issues/15
  - https://github.com/awslabs/aws-c-common/pull/850
  - https://github.com/awslabs/aws-c-common/issues/844
- Respect CMAKE_INSTALL_LIBDIR, CMAKE_INSTALL_BINDIR, CMAKE_INSTALL_INCLUDEDIR
  - https://github.com/awslabs/aws-c-common/pull/792
  - https://github.com/awslabs/aws-c-mqtt/pull/203
  - https://github.com/awslabs/aws-crt-cpp/pull/325

**Description of changes:**
- Indicate that project is known to work with latest CMake
    - OLD: `cmake_minimum_required(VERSION 3.9)`
    - NEW: `cmake_minimum_required(VERSION 3.9...3.31)`
    - REASONING: This gets the CMake 3.10 deprecation warnings to go away. Months ago I realized this fixed the warnings, but hesitated to add it since the "...&lt;max&gt;" syntax wasn't introduced until CMake 3.12. But in [the docs](https://cmake.org/cmake/help/v3.31/command/cmake_minimum_required.html) they call out that it's safe to use even in older projects
        > If the running version of CMake is older than 3.12, the extra ... dots will be seen as version component separators, resulting in the ...&lt;max&gt; part being ignored and preserving the pre-3.12 behavior of basing policies on &lt;min&gt;.
- Respect `CMAKE_INSTALL_LIBDIR`, `CMAKE_INSTALL_BINDIR`, `CMAKE_INSTALL_INCLUDEDIR`
    - Use these variables directly, instead of defining our own (~LIBRARY_DIRECTORY~, ~RUNTIME_DIRECTORY~)
    - `include(GNUInstallDirs)` on all platforms
    - REASONING: This removes a bunch of boilerplate, and makes it obvious we should defer to CMake for install directory names. I assume the original author thought `GNUInstallDirs` wasn't safe for Windows and Apple, but despite its name, it's fine. [Professional CMake](https://crascit.com/professional-cmake/release-notes/12th-edition/) says:
        > CMake’s GNUInstallDirs module provides a very convenient way to use a standard directory layout. ... it also provides various other standard locations that conform to both GNU coding standards and the FHS. ... the layout can even be used for Windows deployments
- Move installed `aws-c-common-config.cmake`
    - OLD: `lib/aws-c-common/cmake/...`
    - NEW: `lib/cmake/aws-c-common/...`
    - REASONING: while both the old and new locations will work with [find_package()](https://cmake.org/cmake/help/v3.31/command/find_package.html#config-mode-search-procedure), the new location is *more* standard.
      - [Professional CMake](https://crascit.com/professional-cmake/release-notes/12th-edition/) says:
        > ...two very simple and commonly used locations are `<prefix>/cmake` and `<prefix>/lib/cmake/<packageName>`, with variations on the latter being a little friendlier to multi-architecture deployments
      - The [example in the CMakePackageConfigHelpers docs](https://cmake.org/cmake/help/v3.31/module/CMakePackageConfigHelpers.html#example-generating-package-files) use this location
      - openSUSE maintainers have [asked for this change](https://github.com/awslabs/aws-c-event-stream/issues/15#issuecomment-734795493). [Here is the script](https://src.opensuse.org/pool/aws-c-common/src/commit/612d57f07334765e4241f54dfb283d715b6b6ee384e7f7ea3c5a177c9fc1a768/aws-c-common.spec#L74-L78) they're currently using to fix up aws-c-common's installation paths for their platform.
- Move installed CMake modules like `AwsCFlags.cmake` 
    - OLD: `lib/cmake/...`
    - NEW: `lib/cmake/aws-c-common/modules/...`
    - REASONING: it's standard to put ALL of a library's scripts under the same folder. It also makes the location easier to find in the next step...
- Tweak `aws-c-common-config.cmake` so that, when any dependency calls `find_package(aws-c-common)`, the  `CMAKE_MODULE_PATH` is modified to include the installed location of modules like `AwsCFlags.cmake`.
    - REASONING: aws-c-common has a bunch of CMake modules (e.g. AwsCFlags.cmake) that are used in downstream dependencies (e.g. aws-checksums). The way that aws-c-common installs these modules, and the way that dependencies consume them, is non-standard. This leads to a lot of hacky boilerplate in each dependency (example: [aws-checksums/CMakeListst.txt](https://github.com/awslabs/aws-checksums/blob/2c7d355b0777ed14fa1377d8197a65b2c4316bf6/CMakeLists.txt#L7-L25), and has caused headaches for various distributions that try and package these dependencies.
- Stop manually setting `FIND_LIBRARY_USE_LIB64_PATHS`, just let CMake handle it
    - REASONING: Here's what [Professional CMake: A Practical Guide 12th Edition](https://crascit.com/professional-cmake/release-notes/12th-edition/) has to say on this topic:
        > CMake is generally aware of the variations and when setting up the platform defaults, it populates the global properties FIND_LIBRARY_USE_LIB32_PATHS, FIND_LIBRARY_USE_LIB64_PATHS and FIND_LIBRARY_USE_LIBX32_PATHS with appropriate values to control which architecture-specific directories should be searched first, if any. Projects can override these with their own custom prefix using the CMAKE_FIND_LIBRARY_CUSTOM_LIB_SUFFIX variable, but the need for this should be very rare.

      This wasn't causing issues, but I'm trying to cut down on the hacky boilerplate copy/pasted into dozens of other CMake files amongst our repos. In the [PR that introduced this](https://github.com/awslabs/aws-c-common/pull/226), it's commented with "this is the absolute dumbest thing in the world" and a [reviewer mentions that it's not necessary in aws-sdk-cpp](https://github.com/awslabs/aws-c-common/pull/226#issuecomment-452829177). My best guess as to why it seemed necessary is that we were trying to get cross-compile working somewhere and this fixed it. But it seems like the real fix is to address this in the cross-compilation toolchain file. It shouldn't be up to every CMake library on earth to include this hack. Removing now in 2024 doesn't seem to cause any issues, so 🪓

Thanks to @r-burns @madebr @glaubitz and others for helpful comments, and the Pull Requests it took us years to finally look at


waahm7 (approved):

sfod (approved):

glaubitz:
Thanks for fixing this. However, this still needs to be addressed in [aws-crt-cpp](https://github.com/awslabs/aws-crt-cpp).

graebm:
> Thanks for fixing this. However, this still needs to be addressed in [aws-crt-cpp](https://github.com/awslabs/aws-crt-cpp).

In progress: https://github.com/awslabs/aws-crt-cpp/pull/702

PR #1182, opened 2025-01-23
Detect when AVX is disabled via OSXSAVE

graebm:
**Issue #:**
internal V1647663988

[SEV-SNP enabled EC2 instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sev-snp.html) crash when using AWS CLI to upload with checksum algorithm CRC64NVME. [AWS CLI v2](https://github.com/aws/aws-cli/tree/v2) uses aws-c-common under the hood to do CRC64NVME.

**Investigation:**
aws-c-common has checks for AVX support, using [CPUID](https://en.wikipedia.org/wiki/CPUID). But apparently, just because CPUID reports that it can do AVX, doesn't necessarily mean it's enabled by the OS. On SEV-SNP enabled instances, AVX is not allowed.

GCC had the same bug until 7.4: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85100

Stackoverflow question: https://stackoverflow.com/questions/72522885/are-the-xgetbv-and-cpuid-checks-sufficient-to-guarantee-avx2-support

**Description of changes:**
* Do further checks before deciding that AVX can be used.
* Move all Intel feature detection into 1 function.
   * This is simpler, it's how we're doing feature detection on some other architectures.
   * Add checks to find the max value that CPUID accepts. We weren't doing this before.

### DIFF
--- a/source/arch/intel/cpuid.c
+++ b/source/arch/intel/cpuid.c
@@ -13,57 +13,65 @@
 #include <stdlib.h>
 
 extern void aws_run_cpuid(uint32_t eax, uint32_t ecx, uint32_t *abcd);
+extern uint64_t aws_run_xgetbv(uint32_t xcr);
 
-typedef bool(has_feature_fn)(void);
+static bool s_cpu_features[AWS_CPU_FEATURE_COUNT];
+static bool s_cpu_features_cached;
 
-static bool s_has_clmul(void) {
+static void s_cache_cpu_features(void) {
+    /* First, find the max EAX value we can pass to CPUID without undefined behavior */
     uint32_t abcd[4];
-    uint32_t clmul_mask = 0x00000002;
-    aws_run_cpuid(1, 0, abcd);
+    aws_run_cpuid(0x0, 0x0, abcd);
+    const uint32_t max_cpuid_value = abcd[0]; /* EAX */
 
-    if ((abcd[2] & clmul_mask) != clmul_mask)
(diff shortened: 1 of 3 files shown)


source/arch/intel/cpuid.c
```diff
@@ -13,57 +13,65 @@
...
-     *
-     * NOTE: It SHOULD be impossible for a CPU to support AVX2 without supporting AVX.
+    /***************************************************************************
+     * CPUID(EAX=7H, ECX=0H): Extended Features
+     **************************************************************************/
+    RUN_CPUID_OR_RETURN(0x7, 0x0);
+    s_cpu_features[AWS_CPU_FEATURE_BMI2] = abcd[1] & (1 << 8);        /* EBX[bit 8] */
+    s_cpu_features[AWS_CPU_FEATURE_VPCLMULQDQ] = abcd[2] & (1 << 10); /* ECX[bit 10] */
+
+    /* NOTE: It SHOULD be impossible for a CPU to support AVX2 without supporting AVX.
```
DmitriyMusatkin, line 74: is this still accurate? can os disable one, but not the other?

graebm: This was happening on some customer's individual Windows computer. We could never reproduce it. But this did fix it. Seems harmless to keep it in

DmitriyMusatkin (approved):

PR #1185, opened 2025-01-27
Fix heap overflow on uri parsing

TingDaoK:

- After we get the colon position, we try to check the next char in the string, but didn't check against the string length. It's possible to cause a heap overflow

*Description of changes:*

### DIFF
--- a/source/uri.c
+++ b/source/uri.c
@@ -282,8 +282,9 @@ static void s_parse_scheme(struct uri_parser *parser, struct aws_byte_cursor *st
         return;
     }
 
-    /* make sure we didn't just pick up the port by mistake */
-    if ((size_t)(location_of_colon - str->ptr) < str->len && *(location_of_colon + 1) != '/') {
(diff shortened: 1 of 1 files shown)


source/uri.c
```diff
@@ -282,8 +282,9 @@ static void s_parse_scheme(struct uri_parser *parser, struct aws_byte_cursor *st
         return;
     }
 
-    /* make sure we didn't just pick up the port by mistake */
-    if ((size_t)(location_of_colon - str->ptr) < str->len && *(location_of_colon + 1) != '/') {
+    /* Ensure location_of_colon is not the last character before checking *(location_of_colon + 1) */
```
waahm7, line 285:
```suggestion
    /* Ensure location_of_colon is not the last character before checking *(location_of_colon + 1) to avoid buffer overflow */
```

waahm7 (approved):

DmitriyMusatkin (approved):

PR #1186, opened 2025-01-28
Integration test for CPU feature detection

graebm:
**Issue:**
I accidentally committed broken CPU-feature-detection code in https://github.com/awslabs/aws-c-common/pull/1182 (fixed the next day by https://github.com/awslabs/aws-c-common/pull/1184)

This kind of accident slipped through because I was in a rush, and we don't have unit tests to double-check this particular code. Unit tests for feature detection are hard because these features vary wildly machine to machine.

**Description of changes:**
Add an integration test that checks CPU-feature-detection against linux's easy-to-parse `/proc/cpuinfo`.

This test is skipped on platforms like Windows and Apple where there's no simple way to query CPU features

### DIFF
--- a/bin/system_info/test-print-system-info.py
+++ b/bin/system_info/test-print-system-info.py
@@ -4,7 +4,7 @@ import os
 from pathlib import Path
 import platform
 import re
-from subprocess import run
+import subprocess
 from typing import Dict, List, Optional
 
 PARSER = argparse.ArgumentParser(
@@ -15,6 +15,14 @@ PARSER.add_argument("--build-dir",
                     help="Search this dir for print-sys-info app")
 
 
+FEATURE_ALIASES = {
+    'avx512': ['avx512f'],
+    'clmul': ['pclmulqdq'],
+    'crc': ['crc32'],
+    'crypto': ['aes'],
+}
+
+
 def main():
     args = PARSER.parse_args()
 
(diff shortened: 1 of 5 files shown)


bin/system_info/test-print-system-info.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python3
```
waahm7, line 1: trivial/debatable: We are mixing Python and C scripts in the system_info/* directory. Should we move CMakeLists.txt and print_system_info.c into another directory?

waahm7: Another option is moving python script to integration-testing directory like http https://github.com/awslabs/aws-c-http/tree/main/integration-testing.

waahm7 (approved):

bin/system_info/test-print-system-info.py
graebm: 🤷‍♀️ I'm fine mixing file types in a folder.
I personally prefer to group things by "system" vs type.
The theory being:
- if you're deleting the `bin/system_info` directory, you delete everything related to it.
- If you're copying the `bin/system_info` directory, to make another tool like it... hey what's this python script doing in here? an integration test? huh, maybe my new tool should also have an integration test

PR #1188, opened 2025-02-14
aws_base64_compute_encoded_len() is now exact, doesn't add 1 extra for null-terminator

graebm:
*Issue:*

One of the core principles of aws_byte_buf and aws_byte_cursor, is you DO NOT assume there's a null-terminator, because null-terminators, and the resulting confusion between string "length" and "size"  have led to so many bugs in the long history of C.

But `aws_base64_encode()` tried to be "nice" and add a secret null-terminator after `.len`, but before `.capacity`. To achieve this, `aws_base64_compute_encoded_len()` would say it needed 1 more byte than necessary. This caused trouble, catching Dmitriy off guard the other day. Searching for uses, I see aws-crt-cpp also once had a bug due to this ([fixed in this PR](https://github.com/awslabs/aws-crt-cpp/pull/126))

In changing this, I found that `aws_hex_encode()` did similar, but even worse. It always added a null-terminator, and *included* the null-terminator in the `.len`! Bad! Fortunately, this function was never used except in generating random data for tests. When Bret needed a function like this in 2019 for signing, he avoided this function entirely and built his own [alternate version](https://github.com/awslabs/aws-c-common/pull/328) that doesn't add a null-terminator (it has other differences too).

*Description of changes:*

- `aws_hex_encode()` doesn't add null-terminator anymore
  - `aws_hex_compute_encoded_len()` updated to account for this
- `aws_base64_encode()` doesn't add null-terminator anymore
  - `aws_base64_compute_encoded_len()` updated to account for this
- `aws_base64_compute_decoded_len()` math adjusted so that overflow is impossible
- document some math that confused me
- stop using the term "size" in tests, we care about "len"

### DIFF
--- a/source/encoding.c
+++ b/source/encoding.c
@@ -66,7 +66,7 @@ static const uint8_t BASE64_DECODING_TABLE[256] = {
 int aws_hex_compute_encoded_len(size_t to_encode_len, size_t *encoded_length) {
     AWS_ASSERT(encoded_length);
 
-    size_t temp = (to_encode_len << 1) + 1;
+    size_t temp = to_encode_len << 1;
 
     if (AWS_UNLIKELY(temp < to_encode_len)) {
         return aws_raise_error(AWS_ERROR_OVERFLOW_DETECTED);
@@ -98,7 +98,7 @@ int aws_hex_encode(const struct aws_byte_cursor *AWS_RESTRICT to_encode, struct 
         output->buffer[written++] = HEX_CHARS[to_encode->ptr[i] & 0x0f];
     }
 
-    output->buffer[written] = '\0';
+    AWS_ASSERT(written == encoded_len);
     output->len = encoded_len;
 
     return AWS_OP_SUCCESS;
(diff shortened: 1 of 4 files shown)


source/encoding.c
```diff
@@ -66,7 +66,9 @@ static const uint8_t BASE64_DECODING_TABLE[256] = {
 int aws_hex_compute_encoded_len(size_t to_encode_len, size_t *encoded_length) {
     AWS_ASSERT(encoded_length);
 
-    size_t temp = (to_encode_len << 1) + 1;
+    /* For every byte of input, there will be 2 hex chars of encoded output */
+
+    size_t temp = to_encode_len << 1;
 
     if (AWS_UNLIKELY(temp < to_encode_len)) {
```
TingDaoK, line 73: trivial: unrelated, but will it be easier to use `aws_mul_size_checked` directly?

TingDaoK (approved):

source/encoding.c
graebm: I'm trying to leave as much as possible unchanged.

I actually rewrote a bunch of them yesterday, thinking that they were always overestimating, then discovered I didn't totally understand the math, so put it back like it was, just with more comments 🤷‍♀️

DmitriyMusatkin (approved):

PR #1194, opened 2025-03-24
fix: Avoid double definition of inline functions

jjerphan:

Fix https://github.com/awslabs/aws-c-common/issues/1155.

*Description of changes:*

This prevents double definitions of inline functions by avoiding an inclusion of `math.gcc_builtin.inl` after another header (for instance after `math.msvc.inl` when `clang-cl` is used).

Please, let me know how this can be tested.


graebm:
Thanks for this PR.
I opened another (similar) one over here https://github.com/awslabs/aws-c-common/pull/1195 that makes some more severe edits

PR #1197, opened 2025-04-16
Error handling docs

graebm:
Whenever someone contributes code for the first time, I need to walk them through all this tribal knowledge. So finally write it down.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,8 @@ returning control to the caller, if you have an error to raise, use the `aws_rai
 * For APIs returning an allocated instance of an object, return the memory on success, and `NULL` on failure. Before
 returning control to the caller, if you have an error to raise, use the `aws_raise_error()` function.
 
+See [error-handling.md](docs/error-handling.md) for a longer tutorial.
+
 #### Log Subjects & Error Codes
 The logging & error handling infrastructure is designed to support multiple libraries. For this to work, AWS maintained libraries
 have pre-slotted log subjects & error codes for each library. The currently allocated ranges are:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -229,11 +231,17 @@ Example:
     AWS_COMMON_API
  void aws_module_destroy(aws_module_t *module);
 
-* Avoid c-strings, and don't write code that depends on `NULL` terminators. Expose `struct aws_byte_buf` APIs
-and let the user figure it out.
+* Avoid c-strings, and don't write code that depends on `NULL` terminators.
+    * Pass strings via `struct aws_byte_cursor`. This is a non-owning view type. Pass it by value.
+    * Only pass `const char *` when thinly wrapping an OS function that *requires* a `NULL` terminator.
```
DmitriyMusatkin, line 236: should this call out that cursors dont have null terminator after then and you need to call a function to get a char * with one

docs/error-handling.md
```diff
@@ -0,0 +1,437 @@
...
+```
+
+Functions that may fail, but don't return a pointer, generally look like this:
+```c
+/* Return AWS_OP_SUCCESS (0) if thing successfully did something.
+ * Returns AWS_OP_ERR (-1) if it failed, use aws_last_error() to get the error code explaining why. */
+int aws_thing_do_something(...);
+```
+
+`aws_last_error()` returns an error code that's stored in a thread-local variable. Any function that fails will make sure to set that thread-local error code by calling `aws_raise_error(int error_code)`.
```
DmitriyMusatkin, line 19: should we note somewhere here that even though return is int, the only 2 valid values are SUCCESS and ERR?

DmitriyMusatkin: i see it mentioned later, but i dont see explanation for why we do it this way

DmitriyMusatkin (approved):

docs/error-handling.md
graebm: I'll say it again here. You're right, it is very important to understand

waahm7 (approved):

PR #1201, opened 2025-05-09
Fix invalid XML Buffer Overflow Error

waahm7:
*Description of changes:*
Our XML parser assumes that the XML provided is valid and is intended for internal use only. I have fixed buffer overflow for some cases as it's nice to fix them but there might be other issues with invalid XML.

### DIFF
--- a/AWSCRTAndroidTestRunner/gradle/wrapper/gradle-wrapper.properties
+++ b/AWSCRTAndroidTestRunner/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Wed Apr 01 11:18:00 PDT 2020
+#Fri May 09 10:40:15 PDT 2025
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.9-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
--- a/source/xml_parser.c
+++ b/source/xml_parser.c
@@ -127,6 +127,12 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
             goto clean_up;
         }
 
+        if (start > location) {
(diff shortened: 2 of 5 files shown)


AWSCRTAndroidTestRunner/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Apr 01 11:18:00 PDT 2020
+#Fri May 09 10:40:15 PDT 2025
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.9-bin.zip
```
waahm7, line 4: VSCode promoted me to upgrade Gradle.

tests/xml_parser_test.c
```diff
@@ -95,6 +95,25 @@ static int s_xml_parser_child_with_text_test(struct aws_allocator *allocator, vo
 
 AWS_TEST_CASE(xml_parser_child_with_text, s_xml_parser_child_with_text_test)
 
+const char *malformed = "<?xml version=\"1.0\" encoding=\"UTF-8\"?><rootNode>><child1>TestBody</child1></rootNode>";
+
+static int s_xml_parser_malformed_test(struct aws_allocator *allocator, void *ctx) {
```
DmitriyMusatkin, line 100: nit: i would call out exactly what is malformed in this test, in case we want to add more malformed tests

DmitriyMusatkin (approved):

source/xml_parser.c
```diff
@@ -300,8 +307,12 @@ int aws_xml_node_traverse(
             parent_closed = true;
         }
 
-        size_t node_name_len = end_location - next_location;
-
+        size_t node_name_len;
+        if (aws_sub_size_checked((size_t)end_location, (size_t)next_location, &node_name_len)) {
```
TingDaoK, line 311: maybe use an if check against the pointer, which will be safer

TingDaoK (approved):
LGTM

tests/xml_parser_test.c
waahm7: Thanks, updated.

source/xml_parser.c
waahm7: Thanks, updated.

PR #1212, opened 2025-07-07
Add API for a more compact (no dashes) UUID-to-str

bretambrose:
Used by some new downstream MQTT logic to generate client ids. Servers are not required to support '-' as a client id character (although IoT Core does).


source/uuid.c
```diff
@@ -66,14 +70,18 @@ int aws_uuid_init_from_str(struct aws_uuid *uuid, const struct aws_byte_cursor *
...
 
-int aws_uuid_to_str(const struct aws_uuid *uuid, struct aws_byte_buf *output) {
+static int s_uuid_to_str(
+    const struct aws_uuid *uuid,
+    struct aws_byte_buf *output,
+    const char *format_string,
+    size_t padded_formatted_length) {
     size_t space_remaining = output->capacity - output->len;
-    AWS_ERROR_PRECONDITION(space_remaining >= AWS_UUID_STR_LEN, AWS_ERROR_SHORT_BUFFER);
+    AWS_ERROR_PRECONDITION(space_remaining >= padded_formatted_length, AWS_ERROR_SHORT_BUFFER);
```
xiazhvera, line 79: If we don’t need the null terminator in this case, should we be comparing the remaining space against `padded_formatted_length - 1` instead?

Also, though the function is totally for internal use, but it would be helpful to mention in the function description that `padded_formatted_length` would include the space for the null terminator.

xiazhvera (approved):
LGTM. Fix & ship.

source/uuid.c
bretambrose: We don't use the null terminator in the output but we still need space for it because snprintf is going to add it.

PR #1229, opened 2025-11-13
Base64url support

DmitriyMusatkin:

*Description of changes:*
There are 2 flavors of base64 - regular one, which we already support and url one. 
url one replaces non url safe chars + and \ with - and _, and avoids padding. 
this pr adds support for url version

### DIFF
--- a/source/encoding.c
+++ b/source/encoding.c
@@ -279,6 +279,7 @@ int aws_base64_compute_decoded_len(const struct aws_byte_cursor *AWS_RESTRICT to
     AWS_ASSERT(to_decode);
     AWS_ASSERT(decoded_len);
 
+    /* strip padding */
     struct aws_byte_cursor trimmed = aws_byte_cursor_right_trim_pred(to_decode, s_ispadding);
 
     const size_t len = trimmed.len;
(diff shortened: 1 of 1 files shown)


source/encoding.c
```diff
@@ -320,13 +370,17 @@ int aws_base64_encode(const struct aws_byte_cursor *AWS_RESTRICT to_encode, stru
...
 
-        output->buffer[str_index++] = BASE64_ENCODING_TABLE[(block >> 18) & 0x3F];
-        output->buffer[str_index++] = BASE64_ENCODING_TABLE[(block >> 12) & 0x3F];
-        output->buffer[str_index++] = BASE64_ENCODING_TABLE[(block >> 6) & 0x3F];
-        output->buffer[str_index++] = BASE64_ENCODING_TABLE[block & 0x3F];
+        output->buffer[str_index++] = encoding_table[(block >> 18) & 0x3F];
+        output->buffer[str_index++] = encoding_table[(block >> 12) & 0x3F];
+        if (AWS_LIKELY(i + 1 < buffer_length)) {
+            output->buffer[str_index++] = encoding_table[(block >> 6) & 0x3F];
+            if (AWS_LIKELY(i + 2 < to_encode->len)) {
```
bretambrose: The fact that we switch between `to_encode->len` and `buffer_length` both here and in the `block` value construction above is confusing.  Can we just use one or the other?

bretambrose (approved):
Minor comment

source/encoding.c
```diff
@@ -216,73 +227,104 @@ int aws_hex_decode(const struct aws_byte_cursor *AWS_RESTRICT to_decode, struct
...
+    size_t decoded_len_tmp = 0;
 
-    /* But last two ascii chars might be padding. */
-    AWS_ASSERT(len >= 4); /* we checked earlier len != 0, and was divisible by 4 */
-    size_t padding = 0;
-    if (input[len - 1] == '=' && input[len - 2] == '=') { /*last two chars are = */
-        padding = 2;
-    } else if (input[len - 1] == '=') { /*last char is = */
-        padding = 1;
+    if (aws_mul_size_checked(len, 3, &decoded_len_tmp)) {
```
xiazhvera: The above comment mentioned we should divide by 4 first to avoid overflow, but we are doing multiply first here?

PR #1237, opened 2026-03-04
process_common: fixed handle leak

ProjectMutilation:
When going on the on_finish label after check aws_byte_buf_append_dynamic(&result_buffer, &cursor), it is possible that the output_stream will not be closed.

*Description of changes:*
- add implicit init of output_stream
- replace closing in on_finish label

### DIFF
--- a/source/process_common.c
+++ b/source/process_common.c
@@ -48,7 +48,7 @@ int aws_run_command(
     AWS_FATAL_ASSERT(options);
     AWS_FATAL_ASSERT(result);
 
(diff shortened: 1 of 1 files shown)


source/process_common.c
```diff
@@ -89,6 +84,13 @@ int aws_run_command(
     ret = AWS_OP_SUCCESS;
 
 on_finish:
+    if (output_buffer) {
```
TingDaoK:
```suggestion
    if (output_stream) {
```

PR #1244, opened 2026-05-01
Add helpers to unescape xml strings

DmitriyMusatkin:

*Description of changes:*
XML supports escaping special characters in the text bodies and in tag values. 
(technically in cdata as well, but we would not support that for now). 
Special characters set is limited to <, >, &, ", ' and generic decimal/hex code points.
This pr adds helper to take a string and unescape special xml chars out of it. i.e replace escaped chars like &quot; with quote and all code points in form of &#nnn; with corresponding utf8 bytes.
Also adds a helper to retrieve escaped body.


source/xml_parser.c
```diff
@@ -419,3 +419,120 @@ int s_node_next_sibling(struct aws_xml_parser *parser) {
 
     return parser->error;
 }
+
+int s_build_unescaped_buffer(struct aws_byte_cursor data, struct aws_byte_buf *out) {
```
TingDaoK: trivial: static

Testing/Temporary/CTestCostData.txt
TingDaoK, line 1: accident?

source/xml_parser.c
```diff
@@ -419,3 +419,120 @@ int s_node_next_sibling(struct aws_xml_parser *parser) {
...
+    if (s_build_unescaped_buffer(data, &temp)) {
+        aws_byte_buf_clean_up(&temp);
+        return AWS_OP_ERR;
+    }
+
+    /*
+     * reserve room up front for the worst possible case: nothing unescaped
+     */
+    if (aws_byte_buf_reserve_relative(out, temp.len)) {
+        return AWS_OP_ERR;
```
TingDaoK, line 523: missed `aws_byte_buf_clean_up(&temp);`

source/xml_parser.c
TingDaoK: it would be helpful if we can document a bit on what we are escaping from this function.

TingDaoK (approved):
fix & ship

PR #1245, opened 2026-05-15
odirect write support

TingDaoK:

*Description of changes:*


include/aws/common/file.h
```diff
@@ -267,6 +267,47 @@ int aws_file_path_read_from_offset_direct_io_with_chunk_size(
...
+ * If O_DIRECT is not supported, returns AWS_ERROR_UNSUPPORTED_OPERATION.
+ *
+ * The file must already exist; the caller is responsible for creating it.
+ *
+ * For aligned writes (offset and length both aligned to page size), O_DIRECT is used for the
+ * entire write.
+ *
+ * Notes:
+ * - ONLY supports linux for now and raises AWS_ERROR_UNSUPPORTED_OPERATION on all other platforms.
+ * - The offset, data.len, and data.ptr all need to be aligned with the page size. Otherwise,
```
DmitriyMusatkin: nit: what does it mean for offset, len to be aligned on page size?  should they be a multiple of page size?

source/linux/file_direct_io.c
```diff
@@ -139,3 +139,89 @@ int aws_file_path_read_from_offset_direct_io(
     return aws_file_path_read_from_offset_direct_io_with_chunk_size(
         file_path, offset, max_read_length, AWS_FILE_MAX_READ_CHUNK, output_buf, out_actual_read);
 }
+
+int aws_file_path_write_to_offset_direct_io(
+    const struct aws_string *file_path,
+    uint64_t offset,
+    struct aws_byte_cursor data) {
+
+    if (O_DIRECT == 0) {
```
DmitriyMusatkin, line 168: does this detect correctly on bsd?

source/linux/file_direct_io.c
```diff
@@ -139,3 +139,89 @@ int aws_file_path_read_from_offset_direct_io(
...
+        AWS_LOGF_ERROR(
+            AWS_LS_COMMON_GENERAL,
+            "Failed to open file %s for writing with O_DIRECT, errno: %d",
+            aws_string_c_str(file_path),
+            errno_value);
+        aws_translate_and_raise_io_error(errno_value);
+        goto cleanup;
+    }
+
+    if (lseek(fd, (off_t)offset, SEEK_SET) == -1) {
```
DmitriyMusatkin, line 208: would this fail to seek if offset is not aligned?

source/linux/file_direct_io.c
```diff
@@ -139,3 +139,89 @@ int aws_file_path_read_from_offset_direct_io(
...
+            aws_string_c_str(file_path),
+            errno_value);
+        aws_translate_and_raise_io_error(errno_value);
+        goto cleanup;
+    }
+
+    size_t page_size = (size_t)sysconf(_SC_PAGESIZE);
+
+    /* Require length to be page-aligned — caller handles unaligned final writes */
+    if (data.len % page_size != 0) {
```
DmitriyMusatkin: no check that data.ptr is aligned?

tests/file_test.c
```diff
@@ -12,6 +12,9 @@
 #include <aws/testing/aws_test_harness.h>
 
 #include <fcntl.h>
+#ifdef __linux__
```
DmitriyMusatkin: dont we have our own defines for whether its linux?

source/linux/file_direct_io.c
TingDaoK: yes, follow the same as the read path in line 134

source/linux/file_direct_io.c
TingDaoK: well, in the end `write` call will fail if those are not aligned. This is mostly a helper to better debug it. We can probably remove this.

source/linux/file_direct_io.c
TingDaoK: the seek itself will not fail. But the write will afterward.

source/linux/file_direct_io.c
TingDaoK: >  The O_DIRECT flag may impose alignment restrictions on the length
 and address of user-space buffers and the file offset of I/Os.  In
 Linux alignment restrictions vary by filesystem and kernel version
 and might be absent entirely.  The handling of misaligned O_DIRECT
 I/Os also varies; they can either fail with EINVAL or fall back to
 buffered I/O.
       

https://man7.org/linux/man-pages/man2/open.2.html
looks like it **may** fail.

DmitriyMusatkin (approved):

PR #1248, opened 2026-05-27
base64 decode fix

TingDaoK:
- decode will report a wrong decoded length for corner case like `AB=D`, which is an invalid base64. Reject it instead of success


sbSteveK:
Title explains what the PR is but not what is being fixed. Let's add some minimal info to the description of what's broken and being fixed here.

sbSteveK (approved):

azkrishpy (approved):

source/encoding.c
```diff
@@ -480,16 +481,29 @@ int aws_base64_decode(const struct aws_byte_cursor *AWS_RESTRICT to_decode, stru
...
 
         if (value3 != BASE64_SENTINEL_VALUE) {
             output->buffer[buffer_index++] = (uint8_t)(((value2 << 4) & 0xF0) | ((value3 >> 2) & 0x0F));
             if (value4 != BASE64_SENTINEL_VALUE) {
-                output->buffer[buffer_index] = (uint8_t)((value3 & 0x03) << 6 | value4);
+                output->buffer[buffer_index++] = (uint8_t)((value3 & 0x03) << 6 | value4);
             }
         }
     }
     output->len = decoded_length;
```
azkrishpy, line 499: should we maybe assign output->len = buffer_index instead of decoded_length? and compare that below? Because buffer_index tracks what is written into 'output' better than precomputed decode length.

TingDaoK: no, the buffer index can be negative.

azkrishpy: yeah that makes sense.. should we be catching the encoded length = 0 early maybe? but yes, I think I understand it now

PR #1253, opened 2026-06-26
Byte Buf Helper Func Dynamic or Static

sbSteveK:
Adds `aws_byte_buf_append_dynamic_or_static` helper function. This function can be called and will check the append `to` `aws_byte_buf` for an allocator. If one is present it will call `aws_byte_buf_append_dynamic`. If it's not present it will call `aws_byte_buf_append`.

This is to pass down the helper function that is used in `aws-c-s3` here: https://github.com/awslabs/aws-c-s3/blob/cebed1bd6256c572e55fe68e6ce4f6c07e8cc57d/source/s3_meta_request.c#L1584-L1590 

This also outlines the growability contract in the description of `aws_byte_buf` to explain when an `aws_byte_buf` can be expanded dynamically.

Test for `aws_byte_buf_append_dynamic_or_static` is also included.


include/aws/common/byte_buf.h
```diff
@@ -414,6 +427,28 @@ int aws_byte_buf_append_with_lookup(
...
+ * growability is not known at the call site — most commonly a buffer
+ * obtained from an aws_s3_buffer_pool ticket, where pool-backed tickets
+ * return a fixed-size buffer (allocator == NULL) and the default pool may
+ * return a growable one.
+ *
+ * `from` and `to` may be the same buffer, permitting copying a buffer
+ * into itself.
+ */
+AWS_COMMON_API
+int aws_byte_buf_append_dynamic_or_static(struct aws_byte_buf *to, const struct aws_byte_cursor *from);
```
TingDaoK: i don't like the name, ai suggested:
> Something like `aws_byte_buf_append_safe` or `aws_byte_buf_append_best_effort`

And more:
> aws_byte_buf_append_auto — short, implies "figures it out for you"
> aws_byte_buf_append_adaptive — adapts to whatever buffer type it gets 

I'd choice betwen auto or adaptive.

TingDaoK (approved):
fix and ship

include/aws/common/byte_buf.h
sbSteveK: lol, ai suggested `aws_byte_buf_append_growth` and I tried to make it something clearer. I think I like `auto` the most. Will swap to `aws_byte_buf_append_auto`

PR #424, opened 2019-06-20
Make sure clocks are correct

JonathanHenson:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1254, opened 2026-07-03
XML Parser Fixes

sbSteveK:
Fixes following issues:

Prefix match: Open-tag search for <name falsely matched <name_prefix> (e.g. <Owner matched <OwnerId), miscounting depth and rejecting valid XML or returning wrong body cursors. Fixed by requiring a name-terminator byte after the match.

Self-closing depth count: Self-closing tags like <a/> were treated as nesting opens because / is a valid name terminator. Fixed by detecting /> and skipping without incrementing depth.

Empty tag UB: s_load_node_decl read ptr[(size_t)-1] on input <>. Fixed with a len == 0 early-return.

CDATA/comment/PI handling: Both the child-scan loop (aws_xml_node_traverse) and the close-tag search (s_advance_to_closing_tag) treated , <![CDATA[]]>, and <??> as element tags. Fixed by detecting and skipping them in both paths. This prevents false close-tag matches inside CDATA and false open-tag matches inside comments.

O(N²) on nested same-name elements: The inner open-tag search scanned the entire remaining document when no opens existed before the close tag. Fixed by limiting the search region to bytes before the found close tag.

Boundary condition: When an open-tag match filled exactly the remaining search region with no byte after it, the name-terminator check was skipped. Fixed by treating boundary matches as non-matches.

Tests added; all failed pre-fixes and pass post:
xml_parser_prefix_collision_test - <Owner> containing <OwnerId> parses correctly
xml_parser_nested_same_name_with_prefix_children_test - <a> with nested <a> and sibling <ab> parses correctly
xml_parser_comment_skipped_test -  skipped during traversal
xml_parser_cdata_skipped_test - <![CDATA[...]]> skipped during traversal
xml_parser_pi_skipped_test - <?...?> skipped during traversal
xml_parser_comment_with_gt_test - > inside comment doesn't terminate it early
xml_parser_nested_same_name_large_depth_test - N=20000 nested <a> elements parses within 2 seconds
xml_parser_self_closing_same_name_test - <a><a/>inner</a> parses correctly
xml_parser_cdata_false_close_test - </data> inside CDATA doesn't terminate body
xml_parser_comment_false_open_test - <tag> inside comment doesn't increment depth


source/xml_parser.c
```diff
@@ -159,6 +164,113 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
...
+static int s_find_skipping_non_elements(
+    struct aws_byte_cursor input,
+    const struct aws_byte_cursor *to_find,
+    struct aws_byte_cursor *result) {
+
+    while (input.len >= to_find->len) {
+        /* Check for CDATA/comment/PI at current position */
+        if (input.len > 3 && input.ptr[0] == '<') {
+            /* <!-- comment --> */
+            if (input.len > 3 && input.ptr[1] == '!' && input.ptr[2] == '-' && input.ptr[3] == '-') {
```
TingDaoK: can we use `aws_byte_cursor_starts_with` instead?

source/xml_parser.c
```diff
@@ -159,6 +164,113 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
...
+                struct aws_byte_cursor remainder = aws_byte_cursor_from_array(input.ptr + 4, input.len - 4);
+                struct aws_byte_cursor end_marker = aws_byte_cursor_from_c_str("-->");
+                struct aws_byte_cursor found;
```
TingDaoK: why we need get the remainder? 
trivial: the end mark can be a static cursor.

source/xml_parser.c
```diff
@@ -40,6 +40,11 @@ static int s_load_node_decl(
     AWS_PRECONDITION(decl_body);
     AWS_PRECONDITION(node);
 
+    if (decl_body->len == 0) {
```
DmitriyMusatkin, line 43: is <> considered invalid node in xml spec?

source/xml_parser.c
```diff
@@ -159,6 +164,113 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
     return parser.error;
 }
 
+/* Returns true if the byte terminates an XML element name */
+static bool s_is_name_terminator(uint8_t c) {
```
DmitriyMusatkin: is this called out in xml spec? can we doc relevant section in rfc?

source/xml_parser.c
```diff
@@ -159,6 +164,113 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
...
+ * Find `to_find` in `input`, skipping over CDATA sections, comments, and processing
+ * instructions. Returns AWS_OP_SUCCESS and sets *result on match, or raises
+ * AWS_ERROR_STRING_MATCH_NOT_FOUND / AWS_ERROR_INVALID_XML.
+ */
+static int s_find_skipping_non_elements(
+    struct aws_byte_cursor input,
+    const struct aws_byte_cursor *to_find,
+    struct aws_byte_cursor *result) {
+
+    while (input.len >= to_find->len) {
```
TingDaoK: honestly, it may not matter in production, but this algo is not very efficient.

looks like what you are doing here:
1. if the string starts with `CDATA/comment/PI` open, skip to the end.
2. otherwise, search the full string for the to_find
3. check the between of input and to_find for any `CDATA/comment/PI`
4. if found, ignore the current found and move the cursor to beginning of the `CDATA/comment/PI`
5. loop to 1

the time complexity is o(n^2), each search is o(n), and you could end up with n searches.

Instead:
i think it will be better if you do a state machine like scan of the input:
scan the input from beginning, keep moving the cursor. while it can be couple states:
1. initial state -- nothing found -> search for `CDATA/comment/PI` start and `to_find`
2. `CDATA/comment/PI` started state -- found the opening -> only search for the corresponding ending, ignore anything else.
3. `to_found` ending state.

it will be o(n) instead. And also easier to follow/understand.

source/xml_parser.c
DmitriyMusatkin: aws_byte_cursor_starts_with?

source/xml_parser.c
```diff
@@ -159,6 +164,113 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
...
+        /* Check for CDATA/comment/PI at current position */
+        if (input.len > 3 && input.ptr[0] == '<') {
+            /* <!-- comment --> */
+            if (input.len > 3 && input.ptr[1] == '!' && input.ptr[2] == '-' && input.ptr[3] == '-') {
+                struct aws_byte_cursor remainder = aws_byte_cursor_from_array(input.ptr + 4, input.len - 4);
```
DmitriyMusatkin: maybe use advance here? adding ptr math to solve pointer math related issue seems a bit scary? :)

source/xml_parser.c
```diff
@@ -159,6 +164,113 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
...
+                }
+                size_t skip = (found.ptr + 2) - input.ptr;
+                aws_byte_cursor_advance(&input, skip);
+                continue;
+            }
+        }
+
+        /* Try to find the pattern starting from current position */
+        struct aws_byte_cursor find_result;
+        if (!aws_byte_cursor_find_exact(&input, to_find, &find_result)) {
```
DmitriyMusatkin: starts_with is really what you need here

DmitriyMusatkin (approved):
nits

source/xml_parser.c
```diff
@@ -214,26 +326,65 @@ int s_advance_to_closing_tag(
...
+            size_t search_len = close_find_result.ptr - parser->doc.ptr;
+            if (search_len == 0) {
+                /* No room for opens before the close tag — consume it directly. */
+                aws_byte_cursor_advance(&parser->doc, closing_cmp_buf.len);
+                depth_count--;
+                break;
+            }
+            struct aws_byte_cursor open_search_region = aws_byte_cursor_from_array(parser->doc.ptr, search_len);
+
+            if (!s_find_skipping_non_elements(open_search_region, &to_find_open, &open_find_result)) {
```
TingDaoK: trivial: 
```suggestion
            if (s_find_skipping_non_elements(open_search_region, &to_find_open, &open_find_result) == AWS_OP_SUCCESS) {
```
my brain cannot work with this. Sometimes we return true/false for helpers, sometimes we return error code.

source/xml_parser.c
```diff
@@ -214,26 +326,65 @@ int s_advance_to_closing_tag(
...
                     continue;
                 }
+
+                /* If terminator is '/' followed by '>', this is a self-closing tag
+                 * (e.g. <a/>) — skip past it without incrementing depth. */
+                if (*after_match == '/') {
+                    size_t offset_in_region = (after_match + 1) - parser->doc.ptr;
+                    if (offset_in_region < search_len && *(after_match + 1) == '>') {
+                        /* Self-closing tag — skip past it without incrementing depth */
+                        size_t skip_len = open_find_result.ptr - parser->doc.ptr;
```
TingDaoK: all those pointer, length, index is killing me. 🤯 

use the byte cursor please. for example: here you can use cursor_advance, and then start_with.

source/xml_parser.c
```diff
@@ -293,6 +444,62 @@ int aws_xml_node_traverse(
             goto error;
         }
 
+        /* Skip CDATA, comments, and processing instructions — they are not elements. */
```
TingDaoK, line 398: this seems to be duplicated, can we move it as a helper?

source/xml_parser.c
sbSteveK: The spec says any document containing <> must be "well-formed" and that contains one or more elements: https://www.w3.org/TR/xml/#sec-well-formed

Pretty sure this means this is following XML spec by erroring.

sbSteveK: Follow-up; tags for empty elements still seem to require a name tag: https://www.w3.org/TR/xml/#sec-starttags

source/xml_parser.c
sbSteveK: There is no specific callout in the spec for these characters. They're derived from what we expect to be treated as terminators for the purpose of checking whether we are matching a complete name. Specifically characters that we treat as a known "this is the end" or not in the  `NameChar` list of spec compliant name characters: https://www.w3.org/TR/xml/#NT-NameChar

Since we only use this thing to determine a false positive in matching, I'll rename it to `s_is_tag_name_boundary` instead of `s_is_name_terminator` to clarify what's happening.

source/xml_parser.c
sbSteveK: Using `aws_byte_cursor_starts_with()` and establishing checked prefixes and ends for repeated use.

source/xml_parser.c
sbSteveK: I think we need to use find_exact. My comment was bad. We are starting from the current position but searching the entirety for the match. Will re-word the comment here

source/xml_parser.c
sbSteveK: Swapped it out

TingDaoK (changes requested):

source/xml_parser.c
TingDaoK: Since the PR is trying to address one O(n^2) issue, so, I'd argue we must not introduce another O(n^2) here.

source/xml_parser.c
sbSteveK: It was in case for some terrible reason the comment starts and there's an immediate `>` i.e. `<!-->`. Using the remainder make sure we started past the full comment start.

Swapped out the search cursors to aws_byte_cursor but making them all static probably makes more sense.

source/xml_parser.c
sbSteveK: I think we are only ever returning either `AWS_OP_SUCCESS` or `AWS_OP_ERR` from this function. When something goes wrong with the parser, it does seem weird that we are setting `parser->error` to `aws_raise_error(ACTUAL_ERROR)` which translates to `AWS_OP_ERROR` in `parser->error`. But also, `parser->error` is what's eventually returned at the end of this func so....

source/xml_parser.c
sbSteveK: sorted it out

source/xml_parser.c
sbSteveK: swapped to byte cursors everywhere

source/xml_parser.c
sbSteveK: Oh wait, this may be about just general Error and helper funcs? I don't understand why we return an int that is only ever supposed to be either `AWS_OP_SUCCESS` or `AWS_OP_ERROR` and we `aws_raise_error` separately. I think the pattern I've adopted is to always raise the error and just return a success or failure to make sure we're at least logging the error and it doesn't get lost. Maybe it's something we can do a wipe of our repos to unify sometime.

source/xml_parser.c
sbSteveK: Found what I think is the duplicated code you're referring to in `s_find_skipping_non_elements` and they're doing similar searches but doing slightly different things and failure is being handled a little bit differently as well. This search is also looking specifically for any next element to process while the use of the other area is looking for a pattern match. I think unifying them into a single helper would make things a little dirtier but I don't feel strongly enough against it to NOT do it if you want it.

source/xml_parser.c
```diff
@@ -159,6 +164,86 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
...
+        if (aws_byte_cursor_starts_with(&input, &s_pi_prefix)) {
+            struct aws_byte_cursor remainder = input;
+            aws_byte_cursor_advance(&remainder, s_pi_prefix.len);
+            struct aws_byte_cursor found;
+            if (aws_byte_cursor_find_exact(&remainder, &s_pi_end, &found)) {
+                return aws_raise_error(AWS_ERROR_INVALID_XML);
+            }
+            aws_byte_cursor_advance(&input, (found.ptr + s_pi_end.len) - input.ptr);
+            continue;
+        }
```
TingDaoK: this three blocks are basically the identical, maybe move it to a helper?
And the helper can be reused afterward.

source/xml_parser.c
```diff
@@ -159,6 +164,86 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
...
+ * AWS_ERROR_STRING_MATCH_NOT_FOUND / AWS_ERROR_INVALID_XML.
+ */
+static int s_find_skipping_non_elements(
+    struct aws_byte_cursor input,
+    const struct aws_byte_cursor *to_find,
+    struct aws_byte_cursor *result) {
+
+    while (input.len > 0) {
+        /* Find the next '<' as both to_find patterns and non-element markers start with '<'. */
+        const uint8_t *open = memchr(input.ptr, '<', input.len);
```
TingDaoK: add a note/warning, this helper only works if your `to_find` starts with `<`. 
or a quick assert/check at beginning to reject anything that is not starting with `<`.
I know currently, only cases we use this helper, the `to_find` always starts with `<`. But, if we ever reuse this function somewhere else, it will be a bit painful to debug.

source/xml_parser.c
TingDaoK: hmmm, looks like you are not here?

source/xml_parser.c
TingDaoK: you can copy the `doc` cursor as an input, and move the input cursor around after found `<`. then it's basically the same as what you are doing above. 
For error handling, the helper don't need to do any error handling, the caller can handle the error differently.

tests/xml_parser_test.c
```diff
@@ -617,3 +618,373 @@ static int s_xml_parser_child_with_text_escaped_test(struct aws_allocator *alloc
...
+
+    return AWS_OP_SUCCESS;
+}
+
+AWS_TEST_CASE(xml_parser_comment_with_gt_test, s_xml_parser_comment_with_gt_test)
+
+/*
+ * Verify that parsing deeply nested same-name elements (N=20000) completes in O(N) time.
+ * The test asserts the parse finishes within 2 seconds.
+ */
```
TingDaoK, line 856: does it fail before head?

source/xml_parser.c
```diff
@@ -214,26 +299,65 @@ int s_advance_to_closing_tag(
     struct aws_byte_cursor close_find_result;
     AWS_ZERO_STRUCT(close_find_result);
     do {
```
TingDaoK: ### `s_find_skipping_non_elements`

The rewritten version (using `memchr` to jump between `<` characters) is O(n) and clean — nice improvement over the original.

### `s_advance_to_closing_tag`

The two-loop structure (outer do/while searching for the close tag, inner while searching for opens before it) is more complex than it needs to be. The whole function can be a single forward scan:

```c
size_t depth = 1;
while (parser->doc.len > 0) {
    const uint8_t *open = memchr(parser->doc.ptr, '<', parser->doc.len);
    if (!open) break;
    aws_byte_cursor_advance(&parser->doc, open - parser->doc.ptr);

    // Skip non-element markers
    if (starts_with("<!--"))      { skip to "-->"; continue; }
    if (starts_with("<![CDATA[")) { skip to "]]>"; continue; }
    if (starts_with("<?"))        { skip to "?>"; continue; }

    // Closing tag — decrease depth
    if (starts_with(to_find_close)) {
        advance past it;
        depth--;
        if (depth == 0) return SUCCESS;
        continue;
    }

    // Opening tag (with name boundary check) — increase depth
    if (starts_with(to_find_open) && is_tag_name_boundary(next_byte)) {
        // check for self-closing /> — don't increment if so
        advance past it;
        depth++;
        continue;
    }

    // Some other '<' — skip past it
    aws_byte_cursor_advance(&parser->doc, 1);
}
return AWS_ERROR_INVALID_XML;
```

This is O(n) with a single pass — each byte is visited at most once. The logic is easier to follow because the invariant is simple: `depth` tracks how many unmatched opens we've seen so far.

The current two-loop approach (find the close first, then go back to count opens before it) solves the same problem but is harder to reason about correctness, and `s_find_skipping_non_elements` ends up being called multiple times over overlapping regions.

---

source/xml_parser.c
```diff
@@ -293,6 +417,59 @@ int aws_xml_node_traverse(
             goto error;
         }
 
+        /* Skip CDATA, comments, and processing instructions — they are not elements. */
+        size_t offset_to_open = next_location - parser->doc.ptr;
+        size_t remaining_from_open = parser->doc.len - offset_to_open;
+        if (remaining_from_open > 1 && *(next_location + 1) == '!') {
+            /* <!-- comment --> */
+            if (remaining_from_open > 3 && *(next_location + 2) == '-' && *(next_location + 3) == '-') {
```
DmitriyMusatkin: personal pref, but next_location[2] reads slightly easier

source/xml_parser.c
sbSteveK: Setup the `s_try_skip_non_element` to return AWS_OP_ERROR if it doesn't start with a `<` and added an assert to `s_find_skipping_non_elements`. It's not really possible (currently) for anything to reach `s_try_skip_non_element` with anything that's not an `<` but at least it'll be something we may catch if we use it for something beyond what we're using it for now.

The definition of the function explicitly states you need to start with `<`

tests/xml_parser_test.c
sbSteveK: Failed with 
`424/492 Test #424: xml_parser_nested_same_name_large_depth_test .........................***Failed    6.95 sec` 

This is a CI run pre-any fixes where all added tests are failing as expected:
https://github.com/awslabs/aws-c-common/actions/runs/28675682977/job/85048480140

source/xml_parser.c
```diff
@@ -208,45 +282,73 @@ int s_advance_to_closing_tag(
...
+        /* Skip non-element constructs (comments, CDATA, PI). */
+        const uint8_t *non_element_end = NULL;
+        if (!s_try_skip_non_element(parser->doc, &non_element_end)) {
+            aws_byte_cursor_advance(&parser->doc, non_element_end - parser->doc.ptr);
+            continue;
+        }
+
+        /* s_try_skip_non_element returns AWS_OP_ERROR for both "not a non-element" and "unterminated".
+         * Distinguish by checking whether an error was raised. */
+        if (aws_last_error() == AWS_ERROR_INVALID_XML) {
```
sbSteveK: We do this in two places, it's possible (but insanely unlikely) that another thread might flip this to something else after we have raised an `AWS_ERROR_INVALID_XML` during the `s_try_skip_non_element` helper function. Do we want to change the return on that to an enum or is that overkill for something we will probably never run into?

source/xml_parser.c
sbSteveK: I think I got them all now....

source/xml_parser.c
sbSteveK: helper func added

source/xml_parser.c
sbSteveK: swapped with single pass and helper function use.

source/xml_parser.c
sbSteveK: No `+ 2` remaining

source/xml_parser.c
TingDaoK: this change has different behavior vs the previous code. And possible bugs
for input:
1. `><?--?><valid tag>`, it should fail, but with the update, it will start the next iteration ignores the first `>`.
2. `<?--?>>`, if I understand it correctly, this will read `?--?>` as the next node? since the `next_location` will still be the first `<` and then skip the `<?--?>`, you will found next `>` at the end of string and continues with invalid result?

TingDaoK: I was wrong, the second case will not be a bug here. Since you continues after skipping.

source/xml_parser.c
TingDaoK: well, luckily our error code is per thread.

source/xml_parser.c
```diff
@@ -159,6 +164,75 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
...
+ * If the construct is unterminated (no closing delimiter found), returns AWS_OP_ERR and
+ * raises AWS_ERROR_INVALID_XML.
+ *
+ * WARNING: This function assumes `input` starts with '<'. It will not produce correct
+ * results if called at an arbitrary position within a document.
+ */
+static int s_try_skip_non_element(struct aws_byte_cursor input, const uint8_t **out_end) {
+    if (input.len == 0 || input.ptr[0] != '<') {
+        return AWS_OP_ERR;
+    }
```
TingDaoK, line 202: couple suggestions  
1. why not pass a pointer of the input, just move the cursor in the helper?
2. I'd just assert on `input.ptr[0] == '<'`, or just the same error out. We know current use case will never start with anything else for the helper, just in case future people using this function without move the cursor to the start of `<`, they can see the failure directly, instead of dig around. 
3. Return a boolean, let the caller to raise the error.

lastly, I meant 
```
    if (aws_byte_cursor_starts_with(&input, &s_comment_prefix)) {
        struct aws_byte_cursor remainder = input;
        aws_byte_cursor_advance(&remainder, s_comment_prefix.len);
        struct aws_byte_cursor found;
        if (aws_byte_cursor_find_exact(&remainder, &s_comment_end, &found)) {
            aws_raise_error(AWS_ERROR_INVALID_XML);
            return AWS_OP_ERR;
        }
        *out_end = found.ptr + s_comment_end.len;
        return AWS_OP_SUCCESS;
    }
```
this block is repeated can be a helper instead of the full section, but I won't die on the hill for this. It is fine this way as well.

source/xml_parser.c
```diff
@@ -208,45 +282,72 @@ int s_advance_to_closing_tag(
...
-
-        while (parser->doc.len) {
-            if (!aws_byte_cursor_find_exact(&parser->doc, &to_find_open, &open_find_result)) {
-                if (open_find_result.ptr < close_find_result.ptr) {
-                    size_t skip_len = open_find_result.ptr - parser->doc.ptr;
-                    aws_byte_cursor_advance(&parser->doc, skip_len + 1);
-                    depth_count++;
-                    continue;
+        /* Check for closing tag. */
+        if (parser->doc.len >= closing_cmp_buf.len && aws_byte_cursor_starts_with(&parser->doc, &to_find_close)) {
```
TingDaoK: couple suggestions:
1. avoid mixing `closing_cmp_buf` and `to_find_close`, it's just hard to read, while they are referring to the same thing.
2. `parser->doc.len >= closing_cmp_buf.len` is unnecessary,  `aws_byte_cursor_starts_with` will do the same check

source/xml_parser.c
```diff
@@ -208,45 +282,72 @@ int s_advance_to_closing_tag(
...
-            depth_count--;
-            break;
+            aws_byte_cursor_advance(&parser->doc, closing_cmp_buf.len);
+            continue;
         }
-    } while (depth_count > 0);
 
-    size_t len = close_find_result.ptr - node->doc_at_body.ptr;
+        /* Check for opening tag with same name. */
+        if (parser->doc.len >= to_find_open.len && aws_byte_cursor_starts_with(&parser->doc, &to_find_open)) {
```
TingDaoK: ditto, `parser->doc.len >= to_find_open.len` is included in `aws_byte_cursor_starts_with`.

source/xml_parser.c
TingDaoK: actually, dig into this a bit more, a stray `>` is valid. in the spec only `&` and `<` is invalid. 
So, the old behavior was wrong. 

so, my suggestion here is just keep moving the `parser->doc`, we ignores the part before `<` anyway. There is no need to keep the original cursor. 

In short, after found the first `<`, move the cursor there, and then skip any non-element, to keep moving the cursor, and continue if we actually skipped any. 

For the next iteration, we starts from the end of where we skipped to search for the next `<`.

sbSteveK: Updated. We now treat it as valid and made the change you suggested. In doing so and tracing through everything, found a case where we do not detect a malformed XML but unsure if it's worth addressing because it'd require some substantial changes. At this point I'm beginning to think instead of all this we should have just re-written the entire parser but that's well beyond the scope of this ticket.

Issue I think may be a problem (that we obviously dont currently encounter) is that when we are using `aws_xml_node_traverse()` it treats any `</` encountered as `parent_closed`.

e.g.
```
<root>
    <child>text</wrong>
</root>
```
the `</wrong>` would close for `<child>`.

something like:
```
<root>
    </wrong>
    <child>text</child>
</root>
```

would also return when it hit `</wrong>`


Anyway, this fix also made the `s_xml_parser_malformed_end_node_character_before_start_test` test useless. Refactored the test to `xml_parser_empty_element_name_test`

source/xml_parser.c
sbSteveK: Changed to pass the pointer to the helper func and changed it to a bool. Added the check to see if something was unterminated during the check.

source/xml_parser.c
```diff
@@ -159,6 +164,75 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
...
+        if (aws_byte_cursor_find_exact(&remainder, &s_cdata_end, &found)) {
+            aws_raise_error(AWS_ERROR_INVALID_XML);
+            return false;
+        }
+        aws_byte_cursor_advance(input, (found.ptr + s_cdata_end.len) - input->ptr);
+        return true;
+    }
+    if (aws_byte_cursor_starts_with(input, &s_pi_prefix)) {
+        struct aws_byte_cursor remainder = *input;
+        aws_byte_cursor_advance(&remainder, s_pi_prefix.len);
```
TingDaoK: trivial: we don't need the temp cursor. we can just move the input directly.

source/xml_parser.c
```diff
@@ -159,6 +164,75 @@ int aws_xml_parse(struct aws_allocator *allocator, const struct aws_xml_parser_o
...
+ * If `*input` does not start with a non-element construct, returns false without
+ * raising an error (caller should proceed with normal element handling).
+ *
+ * If the construct is unterminated (no closing delimiter found), returns false and
+ * raises AWS_ERROR_INVALID_XML.
+ *
+ * WARNING: This function assumes `*input` starts with '<'. It will not produce correct
+ * results if called at an arbitrary position within a document.
+ */
+static bool s_try_skip_non_element(struct aws_byte_cursor *input) {
```
TingDaoK: aha, sorry I didn't full understand the case where when we skip something we need to handle it differently. 

1. return int and raise error when failed.
2. take another params that is a pointer of bool, and set it to be true when skip actaully happens. Otherwise, false. So, the caller can handle it differently based on the bool.
```suggestion
static int s_try_skip_non_element(struct aws_byte_cursor *input, bool *out_skipped) {
```

source/xml_parser.c
```diff
@@ -285,37 +383,66 @@ int aws_xml_node_traverse(
...
+        if (!end_location) {
             AWS_LOGF_ERROR(AWS_LS_COMMON_XML_PARSER, "XML document is invalid.");
             aws_raise_error(AWS_ERROR_INVALID_XML);
             goto error;
         }
 
         bool parent_closed = false;
 
-        if (*(next_location + 1) == '/') {
+        if (*(parser->doc.ptr + 1) == '/') {
```
TingDaoK: trivial:
```suggestion
        struct aws_byte_cursor closing_prefix = aws_byte_cursor_from_c_str("</");
        if (aws_byte_cursor_starts_with(&parser->doc, &closing_prefix)) {
```

And the prefix can also moved to static. same to `decl_prefix` above.

source/xml_parser.c
```diff
@@ -285,37 +383,66 @@ int aws_xml_node_traverse(
...
+                AWS_LOGF_ERROR(AWS_LS_COMMON_XML_PARSER, "XML document is invalid.");
+                aws_raise_error(AWS_ERROR_INVALID_XML);
+                goto error;
+            }
+            aws_byte_cursor_advance(&parser->doc, (found.ptr + 1) - parser->doc.ptr);
+            continue;
+        }
+
+        /* parser->doc.ptr is now at '<'. Find the closing '>'. */
+        const uint8_t *end_location = memchr(parser->doc.ptr + 1, '>', parser->doc.len - 1);
```
TingDaoK, line 426: check the parser->doc has at least 1 length after the `<`. Otherwise, we are access memory out of bound. 

That's why I really don't like use ptr directly, everytime you do a ptr math, it's a "security risk"

Add a test case where the string ends with `<`, eg: `<a><`, and make sure it won't crash

sbSteveK: Added the check with error and a test to make sure we error and don't crash.

TingDaoK (approved):

PR #1257, opened 2026-07-15
Fix non-ASCII environment variable handling on Windows

kai-ion:
https://github.com/aws/aws-sdk-cpp/issues/3865

*Description of changes:*
On Windows, aws_get_env, aws_get_env_nonempty, and aws_get_environment_value use narrow getenv() which returns bytes in the active ANSI code page (e.g. CP-1252). Downstream,
  aws_fopen_safe converts paths to wide using MultiByteToWideChar(CP_UTF8, ...), assuming the input is UTF-8. For non-ASCII characters (e.g. é = 0xE9 in CP-1252), the encodings
  don't match, the wide path is mangled, and file opens fail silently.

  This causes the profile credentials provider to silently fail when the user's home path contains non-ASCII characters (common with non-English Windows usernames).

  Fix: Replace getenv() with GetEnvironmentVariableW + WideCharToMultiByte(CP_UTF8, ...) so the returned string is proper UTF-8 that round-trips correctly through aws_fopen_safe.

  Added a Windows-only unit test that sets a non-ASCII env var via SetEnvironmentVariableW and verifies aws_get_env returns valid UTF-8.

### DIFF
--- a/source/windows/environment.c
+++ b/source/windows/environment.c
@@ -10,36 +10,50 @@
 
 struct aws_string *aws_get_env(struct aws_allocator *allocator, const char *name) {
 #ifdef _MSC_VER
-#    pragma warning(push)
-#    pragma warning(disable : 4996)
-#endif
+    /* Convert variable name to wide (env var names are ASCII, which is valid UTF-8) */
+    struct aws_byte_cursor name_cursor = aws_byte_cursor_from_c_str(name);
+    struct aws_wstring *w_name = aws_string_convert_to_wchar_from_byte_cursor(allocator, &name_cursor);
+    if (!w_name) {
+        return NULL;
+    }
+
+    /* Read value as wide characters using _wdupenv_s */
+    wchar_t *w_value = NULL;
+    size_t w_value_len = 0;
+    _wdupenv_s(&w_value, &w_value_len, aws_wstring_c_str(w_name));
+    aws_wstring_destroy(w_name);
+
+    if (w_value == NULL || w_value_len == 0) {
+        free(w_value);
+        return NULL;
+    }
+
+    /* Convert wide value to UTF-8 */
+    struct aws_string *result = aws_string_convert_from_wchar_c_str(allocator, w_value);
+    free(w_value);
+
+    return result;
+#else
     const char *value = getenv(name);
-#ifdef _MSC_VER
-#    pragma warning(pop)
-#endif
 
     if (value == NULL) {
         return NULL;
     }
 
     return aws_string_new_from_c_str(allocator, value);
+#endif
 }
 
 struct aws_string *aws_get_env_nonempty(struct aws_allocator *allocator, const char *name) {
-#ifdef _MSC_VER
-#    pragma warning(push)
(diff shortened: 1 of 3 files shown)


source/windows/environment.c
```diff
@@ -10,64 +10,58 @@
...
-#endif
-
-    if (value == NULL) {
-        *value_out = NULL;
-        return AWS_OP_SUCCESS;
-    }
-
-    *value_out = aws_string_new_from_c_str(allocator, value);
-    if (*value_out == NULL) {
-        return aws_raise_error(AWS_ERROR_ENVIRONMENT_GET);
```
sbiscigl, line 68: this error is removed, but not re-introduced anywhere, shouldnt you still be checking value_out for NULL?

tests/environment_test.c
```diff
@@ -82,3 +82,28 @@ static int s_test_env_functions_fn(struct aws_allocator *allocator, void *ctx) {
...
 
 AWS_TEST_CASE(test_env_functions, s_test_env_functions_fn)
+
+static int s_test_env_non_ascii_path_fn(struct aws_allocator *allocator, void *ctx) {
+    (void)ctx;
+
+    /* Set a non-ASCII environment variable using _wputenv_s. The value contains
+     * U+00E9 which is 'é'. This simulates a Windows user whose USERPROFILE
+     * contains accented characters. */
+    _wputenv_s(L"AWS_TEST_NONASCII_VAR", L"C:\\Users\\hom\x00E9");
```
sbiscigl: you should _not_ invoke this api, you should invoke the CRT api, `_wputenv_s` is windows specific, and this test is not

source/windows/environment.c
```diff
@@ -10,60 +10,58 @@
...
+    struct aws_wstring *w_name = aws_string_convert_to_wchar_from_byte_cursor(allocator, &name_cursor);
+    if (!w_name) {
+        return NULL;
+    }
+
+    /* Read value as wide characters using _wdupenv_s */
+    wchar_t *w_value = NULL;
+    size_t w_value_len = 0;
+    _wdupenv_s(&w_value, &w_value_len, aws_wstring_c_str(w_name));
+    aws_wstring_destroy(w_name);
```
DmitriyMusatkin, line 24: check for error?

source/windows/environment.c
```diff
@@ -10,60 +10,58 @@
 
 struct aws_string *aws_get_env(struct aws_allocator *allocator, const char *name) {
 #ifdef _MSC_VER
-#    pragma warning(push)
-#    pragma warning(disable : 4996)
-#endif
+    /* Convert variable name to wide (env var names are ASCII, which is valid UTF-8) */
```
DmitriyMusatkin, line 13: maybe add more doc around here? it seems like in windows env variables always use wide coding, and ascii is just compat layer on top of that. so if we are switching off getenv we need to make everything wide

DmitriyMusatkin (approved):

sbiscigl (approved):

kai-ion:
Made a new PR to run CI

PR #1258, opened 2026-07-15
aws_file_get_last_modified_epoch

TingDaoK:

*Description of changes:*
- add a cross platform `aws_file_get_last_modified_epoch` to fetch the nano second epoch time for the file last modified.


source/posix/file.c
```diff
@@ -2,6 +2,9 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0.
  */
+
+#define _GNU_SOURCE /* NOLINT(bugprone-reserved-identifier) */
```
azkrishpy, line 6: How is this define used?

azkrishpy (approved):

source/posix/file.c
```diff
@@ -308,3 +311,30 @@ int aws_file_get_length(FILE *file, int64_t *length) {
...
+    if (fd == -1) {
+        return aws_raise_error(AWS_ERROR_INVALID_FILE_HANDLE);
+    }
+
+    if (fstat(fd, &file_stats)) {
+        int errno_value = errno; /* Always cache errno before potential side-effect */
+        return aws_translate_and_raise_io_error(errno_value);
+    }
+
+#if defined(AWS_OS_APPLE)
```
DmitriyMusatkin, line 329: do we need special casing for other bsd based systems?

tests/file_test.c
```diff
@@ -4,10 +4,12 @@
  */
 
 #include <aws/common/allocator.h>
+#include <aws/common/clock.h>
 #include <aws/common/device_random.h>
 #include <aws/common/file.h>
 #include <aws/common/string.h>
 #include <aws/common/system_info.h>
+#include <aws/common/thread.h>
```
DmitriyMusatkin, line 12: why does it need thread?

TingDaoK: aws_thread_current_sleep

source/posix/file.c
TingDaoK: nope, both openbsd and freebsd uses `st_mtim`, we also have ci here that confirms it.

source/posix/file.c
TingDaoK: manylinux1 is too old, `st_mtim` was not part of the standard back then, but it's actually existed. 

AI told me that:
Old glibc can expose `st_mtim` with `_GNU_SOURCE`, since it existed as a BSD/GNU extension (gated on __USE_MISC in glibc's bits/stat.h) long before POSIX standardized it in 2008. 

So, `_GNU_SOURCE` will expose the structure on manylinux1. 

Alternative, we can kill manylinux1

azkrishpy: oh yes looks like the commit history shows that it fails without `_GNU_SOURCE`. I guess we can have it.. killing manylinux1 is a different discussion.

DmitriyMusatkin (approved):

PR #1266, opened 2026-08-24
Make UBSan errors fatal

sfod:
By default, UBSan only reports UB without failing the process. This leads to all UBSan errors silently being ignored in CI since logs are not printed for successful tests. 

*Description of changes:*

UBSan provides a special option, `-fno-sanitize-recover`, which makes encountered errors fatal. The downside is that UBSan now fails on the first error, so when there are multiple errors, multiple CI runs might be required to catch them all.

Fix two UB:

1. Signed integer overflow.
2. Pointer arithmetic with NULL.

### DIFF
--- a/cmake/AwsSanitizers.cmake
+++ b/cmake/AwsSanitizers.cmake
@@ -67,8 +67,17 @@ function(aws_add_sanitizers target)
     endforeach()
 
     if(PRESENT_SANITIZERS)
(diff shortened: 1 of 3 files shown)


bretambrose (approved):

sfod:
We need to fix UB findings in all repos before merging this PR because CI uses `main` branch. Marking this PR as draft so it won't be merged accidentally.

bretambrose (approved):

PR #1268, opened 2026-08-25
Optimize the size of encoded double values

sfod:
Narrow the integer conversion in `aws_cbor_encoder_write_float` to 32 bits, so integral values larger than UINT32_MAX could be encoded as a float when lossless instead of a larger 9-byte integer.

The encoding order is the following:
- a 32-bit integer (<= 5 bytes);
- a float if lossless (5 bytes);
- fallback to double (9 bytes).

This also fixes UB for value == 2^63: it no longer reaches the (int64_t) cast, which overflowed int64.

### DIFF
--- a/source/cbor.c
+++ b/source/cbor.c
@@ -141,8 +141,10 @@ void aws_cbor_encoder_write_float(struct aws_cbor_encoder *encoder, double value
         aws_cbor_encoder_write_single_float(encoder, (float)value);
         return;
     }
+    /* Use INT64_MAX+1 as a threshold so it can be represented as double. */
+    double int64_overflow_threshold = 0x1p63;
     /* Conversation from int to floating-type is implementation defined if loss of precision */
-    if (value <= (double)INT64_MAX && value >= (double)INT64_MIN) {
+    if (value < int64_overflow_threshold && value >= (double)INT64_MIN) {
         /**
          * A prvalue of a floating point type can be converted to a prvalue of an integer type. The conversion
          * truncates; that is, the fractional part is discarded. The behavior is undefined if the truncated value cannot
--- a/source/cbor.c
+++ b/source/cbor.c
@@ -141,9 +141,14 @@ void aws_cbor_encoder_write_float(struct aws_cbor_encoder *encoder, double value
         aws_cbor_encoder_write_single_float(encoder, (float)value);
         return;
     }
-    /* Use INT64_MAX+1 as a threshold so it can be represented as double. */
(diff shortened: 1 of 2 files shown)


source/cbor.c
```diff
@@ -141,8 +141,10 @@ void aws_cbor_encoder_write_float(struct aws_cbor_encoder *encoder, double value
         aws_cbor_encoder_write_single_float(encoder, (float)value);
         return;
     }
+    /* Use INT64_MAX+1 as a threshold so it can be represented as double. */
+    double int64_overflow_threshold = 0x1p63;
```
DmitriyMusatkin, line 151: can you leave a more detailed comment on why its done this way? 
int64_max is 1e63-1, but doing (double)(INT64_MAX) will round up on most platforms and which will result in not parsable int64 being stored

sfod: Yeah, I explained the issue in the CI description but neglected the comment.

source/cbor.c
```diff
@@ -141,8 +141,15 @@ void aws_cbor_encoder_write_float(struct aws_cbor_encoder *encoder, double value
...
-    if (value <= (double)INT64_MAX && value >= (double)INT64_MIN) {
+    /* INT64_MAX is 2^63-1, which can't be represented in a double's 53-bit significand. So under the default
+     * round-to-nearest mode (double)INT64_MAX rounds up to 2^63. That makes "value <= (double)INT64_MAX" behave as
+     * "value <= 2^63", letting 2^63 sneak into the if block below and reach the (int64_t) cast, where truncating
+     * 2^63 overflows int64, which is UB and causes the wrong value to be encoded on some platforms.
+     * Using strict "< 2^63" excludes it. The lower bound needs no such handling: INT64_MIN (-2^63) is a power of two
+     * and exactly representable, so (double)INT64_MIN is exactly -2^63.
+     */
+    double int64_overflow_threshold = 0x1p63;
+    if (value < int64_overflow_threshold && value >= (double)INT64_MIN) {
```
TingDaoK, line 152: not really related to this change, but the previous logic I had was wrong. 
The goal of this function is to `write the float value as small as possible`. So, for values from uint32_max to FLT_MAX, we should be able to write it as a float if there is no loss of precision. 

But if we write them as 64bit int, it will take more bytes than float. 

Integers larger than uint32_t max will result in 9 bytes... So, we should probably comparing against 32bit max and min instead.

sfod: What you're suggesting makes sense and makes my fix obsolete, so kinda makes sense to try to implement it in this PR.

TingDaoK (approved):

PR #1269, opened 2026-08-26
External shutdown error

bretambrose:
Adds a new common error code for "externally-requested-shutdown"


xiazhvera (approved):
Change looks good. While "external" isn't very clear to me. If it means user-initiated, could we rename it accordingly?

bretambrose:
> Change looks good. While "external" isn't very clear to me. If it means user-initiated, could we rename it accordingly?

Actually, I regret saying "user" which I think is a very ambiguous term.  I would prefer external.

PR #1272, opened 2026-09-10
Helpers for file descriptor opertaions. 

TingDaoK:

*Description of changes:*
- helpers to open/write/close file descriptors. 
- With file descriptors, we can use `pwrite` to make the seek and write in an atomic operation, which `FILE *` doesn't provide the similar API.
- aws_tribool to help

### DIFF
--- a/include/aws/common/file.h
+++ b/include/aws/common/file.h
@@ -295,12 +295,68 @@ int aws_file_path_write_to_offset_direct_io(
     uint64_t offset,
     struct aws_byte_cursor data);
 
+/*
+ * Value representing "no descriptor", for use with the DIRECT I/O descriptor functions below.
+ */
+#define AWS_FILE_INVALID_FD (-1)
+
+/*
+ * Open a file for writing with DIRECT I/O, producing a descriptor for
+ * aws_file_write_to_offset_direct_io().
+ *
+ * A caller issuing many writes to the same file can hold one descriptor across all of them
+ * instead of paying an open/close pair per write.
+ *
+ * The file must already exist; the caller is responsible for creating it.
+ *
+ * Notes:
+ * - ONLY supports linux for now and raises AWS_ERROR_UNSUPPORTED_OPERATION on all other platforms.
+ * - Release the descriptor with aws_file_close_direct_io().
+ *
+ * @param file_path         The file path to open.
+ * @param out_fd            Set to the open descriptor on success; left untouched on failure.
+ *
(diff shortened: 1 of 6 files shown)


include/aws/common/file.h
```diff
@@ -295,12 +295,120 @@ int aws_file_path_write_to_offset_direct_io(
...
+ * - Release the descriptor with aws_file_close_fd().
+ * - Opened in binary mode, so the bytes written are the bytes given.
+ *
+ * @param file_path         The file path to open.
+ * @param out_fd            Set to the open descriptor on success; left untouched on failure.
+ *
+ * Returns AWS_OP_SUCCESS, or AWS_OP_ERR (after an error has been raised).
+ */
+AWS_COMMON_API
+int aws_file_open_for_write(const struct aws_string *file_path, int *out_fd);
```
azkrishpy, line 325: I believe we have helpers for opening and closing files. Do we need to add a new public api? `aws_fopen` returns a `FILE` pointer.

TingDaoK: exactly. aws_fopen returns a FILE pointer. But this returns a `int *out_fd`.

include/aws/common/file.h
```diff
@@ -295,12 +295,120 @@ int aws_file_path_write_to_offset_direct_io(
     uint64_t offset,
     struct aws_byte_cursor data);
 
+/*
+ * Value representing "no descriptor", for use with the DIRECT I/O descriptor functions below.
+ */
+#define AWS_FILE_INVALID_FD (-1)
```
DmitriyMusatkin: since fd is an int is it better to just enum it? 
i.e. enum { AWS_FILE_INVALID_FD = -1; } so it gets some type checking?

include/aws/common/file.h
DmitriyMusatkin: can you add in pr description on why we need int *out_fd variant?

TingDaoK: Sure.
`With file descriptors, we can use pwrite to make the seek and write in an atomic operation, which FILE * doesn't provide the similar API.`

DmitriyMusatkin (approved):

include/aws/common/file.h
azkrishpy: But the file descriptor can be gotten using [fileno](https://pubs.opengroup.org/onlinepubs/009604599/functions/fileno.html). Maybe we can rewrite the implementation to reuse aws_fopen and then wrap the os specific details for converting a FILE pointer to a file descriptor?

PR #1273, opened 2026-09-11
Debug API to override default thread options

bretambrose:
* Relevant to https://github.com/awslabs/aws-c-io/issues/841


include/aws/common/thread.h
```diff
@@ -117,11 +117,20 @@ struct aws_thread {
...
 const struct aws_thread_options *aws_default_thread_options(void);
 
+/**
+ * Overrides the default thread options used when creating threads (of particular interest, event loop threads)
+ *
+ * Test/Debug-only.  Not thread safe.
+ *
+ * @param options new default thread options
+ */
+AWS_COMMON_API void aws_set_default_thread_options(struct aws_thread_options *options);
```
sfod: The argument should probably be `const struct aws_thread_options *`.

source/posix/thread.c
```diff
@@ -213,6 +213,10 @@ const struct aws_thread_options *aws_default_thread_options(void) {
     return &s_default_options;
 }
 
+void aws_set_default_thread_options(struct aws_thread_options *options) {
+    s_default_options = *options;
```
sfod, line 217: Currently, this does a shallow copy of `aws_thread_options::name`. Not a big deal considering this is a debug API, but let's at least add a comment that if `name` is not empty, it should outlive all thread creations.

include/aws/common/thread.h
sfod: This changes the global state. So, if one test sets thread options to something invalid, I believe all subsequent tests in the process will have the same invalid thread options.
No idea how to reset options automatically, so maybe add a note that a test is supposed to reset options back to default? And who this snippet:
```c
const struct aws_thread_options prev_options = *aws_default_thread_options();
aws_set_default_thread_options(bad_options);
// ...
aws_set_default_thread_options(&prev_options);
``` 

An alternative is to implement a reset function.

bretambrose: Originally I did that but I don't think multiple tests are run per process.  Otherwise a crashing test would wreck a ctest run.  Instead it's "subprocess aborted" or similar.

sfod: Hm, indeed - run tests in one process, got a failure in some random buffer test.

sfod (approved):

